PR #1, opened 2016-11-25
fix build for targeting arm (android)

jarod:
* JNI_XXX consts change to type c_int because c_char is unsigned targeting arm.
* jint, jlong should use fix length type. Now, use i32, i64 to void add dependency on libc. maybe better to use libc::int32_t, libc::int64_t


src/sys.rs
```diff
@@ -28,22 +28,22 @@ impl ::std::default::Default for __va_list_tag {
...
 pub const JNI_VERSION_1_1: ::std::os::raw::c_uint = 65537;
 pub const JNI_VERSION_1_2: ::std::os::raw::c_uint = 65538;
 pub const JNI_VERSION_1_4: ::std::os::raw::c_uint = 65540;
 pub const JNI_VERSION_1_6: ::std::os::raw::c_uint = 65542;
 pub const JNI_VERSION_1_8: ::std::os::raw::c_uint = 65544;
-pub type jint = ::std::os::raw::c_int;
-pub type jlong = ::std::os::raw::c_long;
+pub type jint = i32;
+pub type jlong = i64;
 pub type jbyte = ::std::os::raw::c_char;
```
Pursuit92: While we're at it, the other `j<type>` types should probably get the same treatment.

jarod:
changed all numeral j{type} to fixed length types

src/sys.rs
```diff
@@ -28,26 +28,26 @@ impl ::std::default::Default for __va_list_tag {
...
-pub type jlong = ::std::os::raw::c_long;
-pub type jbyte = ::std::os::raw::c_char;
-pub type jboolean = ::std::os::raw::c_uchar;
-pub type jchar = ::std::os::raw::c_ushort;
-pub type jshort = ::std::os::raw::c_short;
+pub type jint = i32;
+pub type jlong = i64;
+pub type jbyte = i8;
+pub type jboolean = u8;
+pub type jchar = u8;
```
jarod: fixed

Pursuit92:
Sorry to pull this out from under you - I switched to using the upstream `jni-sys` crate which already had your proposed changes in it. No need to keep duplicating effort.

PR #3, opened 2017-01-06
Add JStaticMethodID and fix static method lookup.

vvanders:
Hey there,

I was poking around in Android and using this library when I noticed that static methods weren't working quite right.

I took the liberty of adding JStaticMethodID which uses GetStaticMethodID since static methods aren't interchangeable and it seemed to work well with the Desc trait.

Let me know if this looks good or if you'd like any modifications, I also tried to clean up a couple warnings and some doctext that wasn't accurate.

-Val


Pursuit92:
Ah, must have missed that! Everything looks good to me!

PR #7, opened 2017-03-26
Add support for ByteBuffer related JNI APIs

robmv:
jni-sys already had the function definitions, this only add the JNIEnv wrapper functions.

### DIFF
--- a/src/wrapper/jnienv.rs
+++ b/src/wrapper/jnienv.rs
@@ -4,6 +4,8 @@ use std::marker::PhantomData;
 
 use std::iter::IntoIterator;
 
+use std::slice;
+
 use std::sync::Mutex;
 use std::sync::MutexGuard;
 
@@ -19,6 +21,7 @@ use objects::JMap;
 use objects::JValue;
 use objects::JClass;
 use objects::JObject;
+use objects::JByteBuffer;
 use objects::JString;
 use objects::JThrowable;
 use objects::JMethodID;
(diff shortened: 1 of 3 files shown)


Pursuit92:
ByteBuffers are probably a good candidate for an API similar to the string types where it's easy (and safe) to go from something like `jByteBuffer` -> `&[u8]` -> `Vec<u8>`.

I'm not very familiar with ByteBuffers from the JNI side of things. Can you take a look at the strings module and see how hard it would be to make a ByteBuffer analogue?

src/wrapper/jnienv.rs
```diff
@@ -207,6 +207,41 @@ impl<'a> JNIEnv<'a> {
...
+    }
+
+    /// Returns the starting address of the memory of the direct
+    /// java.nio.ByteBuffer.
+    pub fn get_direct_buffer_address(&self,
+                                     buf: JObject)
+                                     -> Result<*mut c_void> {
+        let ptr = unsafe {
+            jni_call!(self.internal, GetDirectBufferAddress, buf.into_inner())
+        };
```
Pursuit92: This doesn't need to be unsafe

robmv:
The new commit (will rebase later as a single commit) implement `JByteBuffer` and `JNIEnv` related methods use it and slices for type safety. 

- **new_direct_byte_buffer**: use a `mut [u8]` instead of a pointer and a length.
- **get_direct_buffer_address**: return a `mut [u8]`.
- **get_direct_buffer_capacity**: requires a JByteBuffer instead of generic JObject.

PR #8, opened 2017-03-30
feat: add `new_byte_array` to JNIEnv

lifei:


Pursuit92:
Looks good to me. This is the direction I'd like to continue in for #7 as well.

PR #9, opened 2017-04-10
feat: add javavm interfaces

lifei:
I need use the JNI interfaces for JavaVM, so I add a new file. But it's not completed. I need know some rules before continuing.


jechas01:
closing for lack of activity

PR #10, opened 2017-04-26
add class cache to support android multithreading

king6cong:
cache class for jni multithreading support of android ([see
android doc](https://developer.android.com/training/articles/perf-jni.html#faq_FindClass))

Example
cache your jvm and needed classes in the thread before new thread is created:
```rust,ignore
let jvm = env.cache_java_vm().unwrap();
let _ = env.find_class(CLASS_NAME);
```

use `JniCache::get_jvm()` to get cached jvm afterwards
`find_class` calls afterwards will use the cached class

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,4 +1,5 @@
 #![warn(missing_docs)]
+#![feature(drop_types_in_const)]
 
 //! # Safe JNI Bindings in Rust
 //!
@@ -189,6 +190,9 @@ mod wrapper {
     /// Actual communication with the JVM
     mod jnienv;
     pub use self::jnienv::*;
+
+    mod javavm;
+    pub use self::javavm::*;
 }
 
 pub use wrapper::*;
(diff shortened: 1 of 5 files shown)


Pursuit92:
I'm pretty sure that some of this is not safe to do. For example, caching the JVM pointer in such a way that it can be accessed from different threads could lead to unexpected behavior since the pointer returned by GetJavaVM is specific to the thread in which it's called. The doc you linked even touches on this:

> This usually does what you want. You can get into trouble if you create a thread yourself (perhaps by calling pthread_create and then attaching it with AttachCurrentThread). Now there are no stack frames from your application. If you call FindClass from this thread, the JavaVM will start in the "system" class loader instead of the one associated with your application, so attempts to find app-specific classes will fail.

Maybe only keep a hidden jvm pointer around to delete the global refs on Drop? Even then, I'm not sure how safe it is to use another thread's jvm pointer. 

Also, rather than enforce caching for every `find_class` call, add a `find_class_with_cache` method that accepts a cache to use for the lookup. That way, it's up to library consumers whether or not to cache and where to store the cache.

Side question: What happens if a cached class gets reloaded? Does it reload to the same location pointed to by the cached pointer? Or does that pointer get invalidated?

jechas01:
closing due to lack of activity

PR #12, opened 2017-05-07
Fixing jvalue

vvanders:
This fixes jvalue to use the built-in union generated from jni-sys instead of transmute. Not that it looked like floats were the only thing not working(since JNI expects f32 but was being cast to f64 instead).


Pursuit92:
Looks good!

PR #13, opened 2017-06-08
Add function for 'jbyteArray' to 'Vec<u8>' conversion

DarkEld3r:
Hi,

I need a way to convert `jbyteArray` to `Vec<u8>`, so I have made a pretty straightforward implementation. Probably it would be better to export all array related JNI functions, but I don't see a clear way to improve their ergonomic. Other and even more flexible (but less convenient) solution is to make access for the underlying `sys::JNIEnv` interface, so custom workaround can be made if `jni-rs` public api lacks something.

### DIFF
--- a/src/wrapper/jnienv.rs
+++ b/src/wrapper/jnienv.rs
@@ -758,6 +758,17 @@ impl<'a> JNIEnv<'a> {
         Ok(bytes)
     }
 
+    /// Converts a java byte array to a rust vector of bytes.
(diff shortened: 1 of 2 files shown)


Pursuit92:
Looks fine to me! Just waiting on a response to #14.

DarkEld3r:
Sorry for bothering you, but how about this? Should I fix something?

Pursuit92:
Whoops! Could have sworn I merged this before the version bump.

PR #14, opened 2017-06-08
Add 'get_native_interface' function

DarkEld3r:
Simple alternative to #13.

### DIFF
--- a/src/wrapper/jnienv.rs
+++ b/src/wrapper/jnienv.rs
@@ -1008,8 +1008,8 @@ impl<'a> JNIEnv<'a> {
     }
 
     /// Returns underlying `sys::JNIEnv` interface.
-    pub fn get_native_interface(&self) -> sys::JNIEnv {
-        unsafe { *self.internal }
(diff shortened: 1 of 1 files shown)


Pursuit92:
Wait, how is this an alternative to #13? #13 looks fine to me, but if there's an alternative that was supposed to be here, I'd like to see it first, just to cover all the bases.

DarkEld3r:
Well... this is not a straightforward alternative. :smile: But with direct access to `sys::JNIEnv` it would be possible to access any function of this interface. It would be "more unsafe" and verbose than having a convenient wrapper in `jni-rs` itself, but at least it can be used as a quick workaround (till the correct pull request is done).

Perhaps, I'm missing something, aren't I?

Pursuit92:
Ah, I see your point. I'm cool with the intent here - it would be a good stopgap for a lot of missing functionality and would open the door to some low-level optimizations that can't have safe APIs.

Let's make it return a `*mut sys::JNIEnv` instead though. Dereferencing it will almost certainly cause problems since it's only ever intended to be passed around as a pointer into the C-based JNI.

PR #16, opened 2017-06-23
Add 'Clone' and 'Copy' imlementations for the 'JNIEnv'

DarkEld3r:
`JNIEnv` is passed by value to the extern functions, so Clippy warns about "needless pass by value".
I suppose there is no harm in adding `Copy`?

### DIFF
--- a/src/wrapper/jnienv.rs
+++ b/src/wrapper/jnienv.rs
@@ -52,6 +52,7 @@ use signature::Primitive;
 /// applicable, the null error is changed to a more applicable error type, such
 /// as `MethodNotFound`.
 #[repr(C)]
+#[derive(Clone, Copy)]
 pub struct JNIEnv<'a> {
     internal: *mut sys::JNIEnv,
     lifetime: PhantomData<&'a ()>,


DarkEld3r:
Build is failing now because of this: https://github.com/sfackler/rust-jni-sys/pull/6. I'm going to send different pull request for that. :smile:

DarkEld3r:
On second thought this can be unsafe, because with copy/clone there can be several mutable JNI-interfaces.

sfackler:
Why is this unsafe?

jechas01:
Right, I believe that adding Copy/Clone would break some guarantees that the JNIEnv wrapper tries to give you. It's not unsafe *per se*, but it throws the borrow/mutability checker out the window.

DarkEld3r:
@jechas01 
Sorry for reopening, but I have another thought about this.

@sfackler 
At first I thought that if `Copy` and/or `Clone` is implemented, then someone can make a copy, pass it to another thread and call methods simultaneously. Then I remembered that `Copy` and `Send` are completely different things, so everything [should be OK](https://is.gd/aGBcbH).

So again, am I missing something or such implementations will not break anything? :sweat_smile:

jechas01:
This still breaks mutability assumptions. Having Copy/Clone implemented means that you can get two mutable references to the JNI pointer simultaneously, which wouldn't be allowed with Rust's single-mutability principle. See https://play.rust-lang.org/?gist=8c5fa89bdbd14b13d1789cc44772f12d&version=stable&backtrace=1 as an example of what I'm talking about.

If this is *really* something that the library consumer needs, it can be accomplished via the `get_native_interface` method and the `From<*mut sys::JNIEnv>` implementations. This at least makes it obvious that something outside the norm is going on, rather than an implicit breaking of assumptions.

DarkEld3r:
Now everything is clear, thank you.

PR #21, opened 2017-06-26
Expose more JNIEnv array functions

robmv:
- `get_array_length()`
- `new_<type>_array_length(...)`
- `get_<type>_array_region(...)`
- `set_<type>_array_region(...)`

Ideally `new_<type>_array_length(...)` should be called `new_<type>_array(...)` but there is already a function with the signature `new_byte_array(&self, buf: &[i8]) -> Result<jbyteArray>` so the `_length` suffix is needed for the `byte` case.

The API is different to `new_byte_array(&self, buf: &[i8]) -> Result<jbyteArray>` because this function is not good when processing large arrays, it can use double the memory to pass data from and to a byte slice. `get_<type>_array_region(...)` and `set_<type>_array_region(...)` allow setting and retrieving data from big arrays by region.

The new `jni_non_null_call_void` was needed because the used JNI functions return `void`. Any better solution is welcome.

### DIFF
--- a/src/wrapper/jnienv.rs
+++ b/src/wrapper/jnienv.rs
@@ -11,7 +11,9 @@ use std::sync::MutexGuard;
 
 use errors::*;
 
-use sys::{self, jvalue, jint, jlong, jsize, jbyte, jboolean, jbyteArray};
+use sys::{self, jarray, jboolean, jbooleanArray, jbyte, jbyteArray, jchar,
+          jcharArray,  jdouble, jdoubleArray, jfloat, jfloatArray, jint,
+          jintArray, jlong, jlongArray, jshort, jshortArray, jsize, jvalue};
 use std::os::raw::{c_char, c_void};
 
 use strings::JNIString;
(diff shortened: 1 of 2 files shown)


jechas01:
Two things:

1. It might be prudent to add `non_null!` assertions at the start of functions that take java arrays as input.
2. Can any of these throw exceptions? Would probably be good to check those if so.

robmv:
1. Added `non_null!` for those arguments and `new_<type>_array_length(...)` results. It is null if the array can not be created.
2. `jni_non_null_call_void` already call `check_exception!($jnienv);`

jechas01:
I believe that `jni_call` already calls `non_null` for its result - so the one in the new array methods is unneeded.

Also, minor nit: Can we make the `jni_non_null_call_void` into `jni_void_call`? I'd say that a void return is already understood to not be null, so we can shorten the name and reduce redundancy.

robmv:
Both done!

DarkEld3r:
One thought about naming: `_length` suffix is not very descriptive in my opinion. Maybe something like `new_..._uninitialized_array` will be better? Although such names will be somewhat longer. :thinking:

jechas01:
Actually, now that I think about it, the `new_byte_array` method is changing signature (from `&[i8]` to `&[u8]`), which will require an incompatible version bump. We could take the opportunity now to rename it and take the `_length` suffix off of the other methods.

robmv:
> We could take the opportunity now to rename it and take the _length suffix off of the other methods.

Yes please, without the `_length` suffix the function has relatively the same name that the JNI call.

jechas01:
Something like `byte_array_from_slice` perhaps?

Issue #25, opened 2017-07-29
Calling methods on GlobalRef


angel-manuel:
It is possible to call methods on a GlobalRef?

JNIEnv::call_method takes an JObject and i don't see any way to convert between the types.

DarkEld3r:
I had the same question: #24. I suppose it isn't possible now, so I'm going to try to fix it in the nearest future if you don't make it first. 😄

jechas01:
Fixed in #26

PR #26, opened 2017-07-29
Allowing the use of GlobalRef as an JObject and new callback examples

angel-manuel:
These modifications solve #25.

It allows to turn a `GlobalRef` into an `JObject` by implementing both `AsRef` and `From`
```rust
let obj = *self.global_ref.as_ref();
```

The internal `jobject` of `GlobalRef` is replaced by an `JObject<'static>`. For other approach without that change(uses unsafe `mem::transmute`) see https://github.com/angel-manuel/jni-rs/commit/fb6a3de633709829f22fc133632a5f188614cf58

While testing, I added callbacks, both using JObject and GlobalRef, to the example. Maybe is too much for a HelloWorld example, drop it or do what you see fit.

### DIFF
--- a/example/HelloWorld.h
+++ b/example/HelloWorld.h
@@ -15,6 +15,38 @@ extern "C" {
 JNIEXPORT jstring JNICALL Java_HelloWorld_hello
   (JNIEnv *, jclass, jstring);
 
+/*
+ * Class:     HelloWorld
+ * Method:    factAndCallMeBack
+ * Signature: (ILHelloWorld;)V
+ */
+JNIEXPORT void JNICALL Java_HelloWorld_factAndCallMeBack
+  (JNIEnv *, jclass, jint, jobject);
+
+/*
+ * Class:     HelloWorld
+ * Method:    counterNew
+ * Signature: (LHelloWorld;)J
+ */
+JNIEXPORT jlong JNICALL Java_HelloWorld_counterNew
+  (JNIEnv *, jclass, jobject);
+
+/*
+ * Class:     HelloWorld
+ * Method:    counterIncrement
+ * Signature: (J)V
+ */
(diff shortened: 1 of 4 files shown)


angel-manuel:
Hold on. I just realized this can lead to use after free. :/

angel-manuel:
It can be fixed using

```rust
impl<'a> From<&'a GlobalRef> for JObject<'a> {
    fn from(other: &'a GlobalRef) -> JObject<'a> {
        other.obj
    }
}
```

and then the way to do it would be

```rust
let obj: JObject = (&global_ref).into();
```

angel-manuel:
Now the following fails to compile

```rust
let global_ref = env.new_global_ref(obj).unwrap();
let obj2: JObject = (&global_ref).into();
drop(global_ref);
```

jechas01:
Cool, looks pretty good! and thanks for the examples.

Can you also add this method to `GlobalRef`? The `From` trait hackery seems a little unintuitive to me, and I can't figure out a good way to accomplish the same safety through a trait impl, so a regular method seems to be the way to go.

```rust
    /// Get the object from the global ref
    ///
    /// This borrows the ref and prevents it from being dropped as long as the
    /// JObject sticks around.
    pub fn as_obj<'a>(&'a self) -> JObject<'a> {
        self.obj
    }
```

jechas01:
Eh, on second thought, I'll just merge this and add the `as_obj` method afterward. I'm in issue cleanup mode 😄

angel-manuel:
I agree that `as_obj` is worth it. `From` is just useful if any method accepting `JObject` would ask instead for `Into<JObject>`, and it would be necessary to take a reference anyways so it would continue to be pretty ugly. A normal method is easier to find on the docs and much cleaner.

PR #27, opened 2017-07-31
added JNIEnv::new_exception and JNIEnv::new_exception_with_message

debris:
throwing java exceptions should not be difficult

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 target/
 Cargo.lock
 *.class
+*.swp
--- a/src/wrapper/jnienv.rs
+++ b/src/wrapper/jnienv.rs
@@ -916,6 +916,22 @@ impl<'a> JNIEnv<'a> {
         Ok(jni_call!(self.internal, NewStringUTF, ffi_str.as_ptr()))
     }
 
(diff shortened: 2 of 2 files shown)


jechas01:
What's difficult about it? Currently, you can throw an exception with something like

```rust
env.throw(("java/lang/Exception", "something bad happened"));
```

or

```rust
env.throw_new("java/lang/Exception", "something bad happened");
```

where the first can take anything that can be a `JThrowable` descriptor (or a `JThrowable` itself), and the later always creates a new object.

I don't think that it should necessarily be easy to throw a plain exception with no message whatsoever - that's just sloppy error handling. No need to encourage it by making it overly simplel. And if anyone truly needs it, they already have the mechanism.

If it's just the default `"java/lang/Exception"` class you'd like, I can add `Desc` impls for `&str`, `String`, `JNIString`, etc.

debris:
Thank you for clarification. I totally missed it

jechas01:
No problem. I admit that the descriptor system is a little convoluted and poorly documented. One day I'll get around to writing better docs for it (hopefully).

PR #30, opened 2017-08-03
Add ObjectArray related functions

### LINKED ISSUE #23: Add support for working with object arrays
It would be useful to support working with `Object` arrays. I think this just requires wrappers for

- NewObjectArray,
- GetObjectArrayElement
- SetObjectArrayElement

DarkEld3r:
Closes #23.

### DIFF
--- a/src/wrapper/jnienv.rs
+++ b/src/wrapper/jnienv.rs
@@ -929,18 +929,25 @@ impl<'a> JNIEnv<'a> {
 
     /// Construct a new array holding objects in class `element_class`.
     /// All elements are initially set to `initial_element`.
-    pub fn new_object_array(
+    ///
+    /// This function returns a local reference, that must not be allocated excessively.
+    /// See [Java documentation][1] for details.
+    /// [1]: https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/design.html#global_and_local_references
+    pub fn new_object_array<T>(
         &self,
         length: jsize,
-        element_class: JClass,
+        element_class: T,
         initial_element: JObject,
-    ) -> Result<jobjectArray> {
-        non_null!(element_class, "new_object_array element_class argument");
+    ) -> Result<jobjectArray>
+    where
+        T: Desc<'a, JClass<'a>>
+    {
+        let class = element_class.lookup(self)?;
         Ok(jni_call!(
             self.internal,
             NewObjectArray,
             length,
-            element_class.into_inner(),
+            class.into_inner(),
             initial_element.into_inner()
         ))
     }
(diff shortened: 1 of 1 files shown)


src/wrapper/jnienv.rs
```diff
@@ -926,6 +927,56 @@ impl<'a> JNIEnv<'a> {
         Ok(len)
     }
 
+    /// Construct a new array holding objects in class `element_class`.
+    /// All elements are initially set to `initial_element`.
+    pub fn new_object_array(
+        &self,
+        length: jsize,
+        element_class: JClass,
```
jechas01: Can we use `T` where `T: Desc<'a, JClass<'a>>` here?

src/wrapper/jnienv.rs
```diff
@@ -926,6 +927,56 @@ impl<'a> JNIEnv<'a> {
...
+        Ok(jni_call!(
+            self.internal,
+            NewObjectArray,
+            length,
+            element_class.into_inner(),
+            initial_element.into_inner()
+        ))
+    }
+
+    /// Returns an element of the `jobjectArray` array.
```
dmitry-timofeev: I'd add to the documentation that

> Programmers are required to "not excessively allocate" local references. In practical terms this means that if you're creating large numbers of local references, perhaps while running through an array of objects, you should free them manually with DeleteLocalRef instead of letting JNI do it for you. The implementation is only required to reserve slots for 16 local references, so if you need more than that you should either delete as you go or use EnsureLocalCapacity/PushLocalFrame to reserve more.

(taken from https://developer.android.com/training/articles/perf-jni.html#local_and_global_references)

Here is the official documentation:

> In most cases, the programmer should rely on the VM to free all local references after the native method returns. However, there are times when the programmer should explicitly free a local reference. Consider, for example, the following situations:
> * A native method accesses a large Java object, thereby creating a local reference to the Java object. The native method then performs additional computation before returning to the caller. The local reference to the large Java object will prevent the object from being garbage collected, even if the object is no longer used in the remainder of the computation.
> * A native method creates a large number of local references, although not all of them are used at the same time. Since the VM needs a certain amount of space to keep track of a local reference, creating too many local references may cause the system to run out of memory. For example, a native method loops through a large array of objects, retrieves the elements as local references, and operates on one element at each iteration. After each iteration, the programmer no longer needs the local reference to the array element.

https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/design.html#global_and_local_references

@jechas01, I see that `delete_local_ref` is not public. Is there a public API for this purpose?

jechas01: No, not currently. We can probably go ahead and make the local ref-related functions public, provided we document them. A PR would be welcome :)

src/wrapper/jnienv.rs
DarkEld3r: Fixed. 😄

PR #31, opened 2017-08-05
Add `Desc<JClass>` for the `GlobalRef`

DarkEld3r:
This allows to use `GlobalRef` with `JNIEnv`'s `new_object` method.

The use-case is the following: object class is obtained by the `find_class` function, then global reference (obtained by `new_global_ref` call) of that class is stored for the later use. If this object's constructor should be called, then it is required to use the `new_object` function, which cannot take `GlobalRef` as parameter.

### DIFF
--- a/src/wrapper/descriptors/class_desc.rs
+++ b/src/wrapper/descriptors/class_desc.rs
@@ -23,6 +23,7 @@ impl<'a, 'b> Desc<'a, JClass<'a>> for JObject<'b> {
     }
 }
 
+/// This conversion assumes that the `GlobalRef` is a pointer to a class object.
 impl<'a> Desc<'a, JClass<'a>> for &'a GlobalRef {
     fn lookup(self, _: &JNIEnv<'a>) -> Result<JClass<'a>> {
         Ok(self.as_obj().into())
--- a/src/wrapper/descriptors/class_desc.rs
+++ b/src/wrapper/descriptors/class_desc.rs
@@ -1,6 +1,6 @@
 use strings::JNIString;
 
-use objects::{JObject, JClass};
+use objects::{JObject, JClass, GlobalRef};
 
 use descriptors::Desc;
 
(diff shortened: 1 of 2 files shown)


jechas01:
Looks fine. Should probably document that this conversion assumes that the `GlobalRef` is a pointer to a class object and that there's no confusion as to whether the class is the class of the object or whether the object IS a class. Should probably add the same docs to the `From<JObject> for JClass` impl as well

DarkEld3r:
I have added a little bit documentation. Is the wording OK?

src/wrapper/objects/jclass.rs
```diff
@@ -28,6 +28,7 @@ impl<'a> From<JClass<'a>> for JObject<'a> {
     }
 }
 
+/// This conversion assumes that the `JClass` is a pointer to a class object.
```
dmitry-timofeev: > This conversion assumes that the ~~`JClass`~~ `JObject` is a pointer to a class object.

src/wrapper/descriptors/class_desc.rs
```diff
@@ -23,6 +23,7 @@ impl<'a, 'b> Desc<'a, JClass<'a>> for JObject<'b> {
     }
 }
 
+/// This conversion assumes that the `GlobalRef` is a pointer to a class object.
```
dmitry-timofeev: I'd replace «a pointer to» with «a reference to» since Java doesn't have pointers.

DarkEld3r: I see logic here, but there is a raw pointer (in terms of Rust) inside the `GlobalRef`. :thinking: 

@jechas01 What do you think?

DarkEld3r:
@jechas01 Sorry for bothering, but could you please give some update for this? Should I change something?

jechas01:
No problem, the squeaky wheel gets the grease!

Looks fine to me

PR #37, opened 2017-09-01
Remove 'check_exception' call from 'delete_local_ref' function

### LINKED ISSUE #36: Exception-safe JNI function checks for exceptions
`delete_local_ref` (DeleteLocalRef) is [specified](https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/design.html#exception_handling) as one of a few _safe_ functions to call when there is a pending exception (= the one that do not throw). But the implementation of `JNIEnv#delete_local_ref` checks for exceptions _after_ the call:
```#rust
 pub fn delete_local_ref(&self, obj: JObject) -> Result<()> {
        non_null!(obj, "delete_local_ref obj argument");
        Ok(unsafe {
            jni_unchecked!(self.internal, DeleteLocalRef, obj.into_inner());
            check_exception!(self.internal);
        })
    }
```

It might result in an `Err` when a user invokes `delete_local_ref` when they've already got an exception and try to delete local references.

DarkEld3r:
Closes #36.

### DIFF
--- a/src/wrapper/jnienv.rs
+++ b/src/wrapper/jnienv.rs
@@ -317,7 +317,6 @@ impl<'a> JNIEnv<'a> {
         non_null!(obj, "delete_local_ref obj argument");
         Ok(unsafe {
             jni_unchecked!(self.internal, DeleteLocalRef, obj.into_inner());
-            check_exception!(self.internal);
         })
     }
 


DarkEld3r:
@dmitry-timofeev Result is needed because of non_null! macro - it can return error. Although I'm not sure that argument must be checked at all.

jechas01:
Looks good to me. My philosophy is better safe than sorry when it comes to passing raw pointers around, especially when they're getting tossed over the FFI fence.

PR #39, opened 2017-09-01
Add exception checking for primitives in the 'call_method_unsafe'

### LINKED ISSUE #35: Call of a Java method returning a primitive type does not check for exceptions
The implementation of `call_method_unsafe` (which is called from `call_method`) uses`jni_unchecked!` macro if the method return type is `void` or a primitive type. But if the method throws an exception, it won't be detected by the library unless either:
  * the user explicitly calls `JNIEnv#exception_check/#exception_occurred`
  * the user calls any unrelated JNI function, which is not safe to call when there is an exception pending.

DarkEld3r:
Closes #35.

### DIFF
--- a/src/wrapper/jnienv.rs
+++ b/src/wrapper/jnienv.rs
@@ -650,7 +650,7 @@ impl<'a> JNIEnv<'a> {
             JavaType::Primitive(p) => {
                 let v: JValue = match p {
                     Primitive::Boolean => {
-                        (jni_unchecked!(
+                        (jni_non_null_call!(
                             self.internal,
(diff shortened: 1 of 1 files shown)


DarkEld3r:
@dmitry-timofeev `Result` is needed because of `non_null!` macro - it can return error. Although I'm not sure that argument must be checked at all.

jechas01:
It doesn't care - all it does is skip the null check after the exception check. It can be used for references that we **know** are never going to be null, but methods returning primitives were the original use case for it IIRC.

Issue #34, opened 2017-09-01
Wrapper for local reference automatical deletion


DarkEld3r:
Sometimes it is important to manually delete the local references. I have made a simple wrapper for that:
```rust
pub struct AutoLocalRef<'a> {
    object: jobject,
    env: &'a JNIEnv<'a>,
}

impl<'a> AutoLocalRef<'a> {
    pub fn new(env: &'a JNIEnv<'a>, object: jobject) -> Self {
        Self { object, env }
    }

    pub fn as_obj(&self) -> JObject {
        self.object.into()
    }
}

impl<'a> Drop for AutoLocalRef<'a> {
    fn drop(&mut self) {
        if let Err(e) = self.env.delete_local_ref(self.object.into()) {
            error!("Unable to delete local reference: {}", e.description());
        }
    }
}
```
Typical use case: 
```rust
fn make_java_object<'a>(env: &JNIEnv<'a>, value: &Value) -> Result<JObject<'a>> {
    let value = AutoLocalRef::new(env, env.byte_array_from_slice(value)?);
    env.new_object(
        "com/project/foo/bar/baz/ClassName",
        "([B)V",
        &[value.as_obj().into()],
    )
}
```
This function is used for building tree-like structure, so local references must be deleted because there can be many nodes.

Questions:
1. Will it be useful to have such wrapper in the `jni-rs` library?
2. Wrapper stores `jobject` inside and has method for conversion it to the `JObject` because this allows to clone internal object: many functions take `JObject` by value and reference must be deleted after that. Perhaps this can be handled better somehow.

jechas01:
I like the idea - maybe make it match the GlobalRef wrapper?

jechas01:
OK, got an implementation of this in the develop branch. If you would, take a look and make sure I haven't missed anything obvious and that it works for your use case.

DarkEld3r:
Nice, I see no problem in your code. I will try to replace our wrapper with `AutoLocal` and let you know the result.

jechas01:
Okie doke, merged the wrapper into master and bumped the version to 0.6.0, so you should be able to get it from crates.io.

PR #40, opened 2017-09-14
Call delete_local_ref for classes objects

supercurio:
Fixes a "local reference table overflow (max=512)" error experienced with Android <= 7 and performance slow-down with Android 8.

Such error can occur by using `env.get_field()` in loops when going through elements of a jobjectArray.

I also made an Android app to reproduce the bug and test the fix easily, [available on this repo](https://github.com/supercurio/android-rust-jni-tests)

It's possible that this accumulation of local references still occur with the [`env.get_object_class` call in class_desc.rs](https://github.com/prevoty/jni-rs/blob/v0.5.2/src/wrapper/descriptors/class_desc.rs#L22), but I don't know how to fix this one.

Would it be possible to implement automatic calls to delete_local_ref on the Drop trait for everything concerned? I suppose it would mean to keep a JNIEnv around for each.
Since Android limits the local reference count strictly, it could ease the development with this crate a lot and make its behavior in apps in the wild more predictable.

### DIFF
--- a/src/wrapper/jnienv.rs
+++ b/src/wrapper/jnienv.rs
@@ -770,9 +770,12 @@ impl<'a> JNIEnv<'a> {
 
         let class = self.get_object_class(obj)?;
 
-        unsafe {
+        let res = unsafe {
             self.call_method_unsafe(obj, (class, name, sig), parsed.ret, args)
-        }
+        };
+        self.delete_local_ref(JObject::from(class))?;
+
+        res
     }
 
     /// Calls a static method safely. This comes with a number of
@@ -1574,6 +1577,8 @@ impl<'a> JNIEnv<'a> {
 
         let field_id: JFieldID = (class, name, ty).lookup(self)?;
 
+        self.delete_local_ref(JObject::from(class))?;
+
         unsafe { self.get_field_unsafe(obj, field_id, parsed) }
     }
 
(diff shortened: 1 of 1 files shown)


DarkEld3r:
> Would it be possible to implement automatic calls to delete_local_ref on the Drop trait for everything concerned?

I like the idea. We deal with this problem by additional wrapper around `JObject` (https://github.com/prevoty/jni-rs/issues/34), but such functionality in `Drop` seems more appropriate and ergonomic.

Perhaps, it would be nice to have both `JObject` that deletes its local reference and unsafe/raw variant that does not (`jobject` from `jni-sys` can be used instead, but it is less convenient).

supercurio:
@DarkEld3r yes since it's Rust I expected that resources were released automatically when going out of scope, then saw thanks to Android's crash that they were not for JObjects which I fixed in my code after reading your recent patches and proposal in #34.
But the crash remained due to JClass which are currently not possible to release manually (hence this PR).
Since I'm not very experienced yet with this crate or JNI I still don't know which other references could be accumulated, and after which other calls. Are there others you can think of?

I agree that the ergonomics would improve with automatic reference release.

jechas01:
Looks good to me. Seems like a decent enough stopgap until I (or someone else) can get around to implementing #34

supercurio:
Thanks, looking forward to the next improvements!

PR #47, opened 2017-11-09
Add Invocation API and multi-threading support

madadam:
This PR adds initial support for the Invocation API (https://docs.oracle.com/javase/7/docs/technotes/guides/jni/spec/invocation.html) with focus on enabling support for multi-threaded applications.

Changes:
- add `JavaVM` type to wrap (currently) a subset of the invocation API.
- add `get_java_vm` method to `JNIEnv` which returns the `JavaVM` interface.
- add `DetachedGlobalRef` and corresponding `GlobalRef::detach` to allow sending global refs across threads.
- extend the example to show how to write multi-threaded functions.

Limitations:
- Only `AttachCurrentThread`, `AttachCurrentThreadAsDaemon` and `DetachCurrentThread` are implemented of the Invocation API
- The `JavaVMAttachArgs` (last argument of `AttachCurrentThread` and `AttachCurrentThreadAsDaemon`) is not supported yet
- Only threads attached via `AttachCurrentThread` can be detached currently (this is implemented implicitly via RAII pattern. There is no explicit `detach_current_thread` method).
- `DetachedGlobalRef` leaks the global reference if dropped (must be converted to `GlobalRef` via `attach` to prevent this).
- Error handling could be improved (possibly involves adding new `Error` variants).
- Documentation could be more detailed.

### DIFF
--- a/src/wrapper/java_vm.rs
+++ b/src/wrapper/java_vm.rs
@@ -49,6 +49,20 @@ impl JavaVM {
             JNIEnv::from_raw(ptr as *mut sys::JNIEnv)
         }
     }
+
+    /// Get the `JNIEnv` associated with the current thread, or `ErrorKind::Detached`
+    /// if the current thread is not attached to the java VM.
+    pub fn get_env(&self) -> Result<JNIEnv> {
+        let mut ptr = ptr::null_mut();
+        unsafe {
+            let res = jni_unchecked!(self.0, GetEnv, &mut ptr, sys::JNI_VERSION_1_1);
+            match res {
+                sys::JNI_OK => JNIEnv::from_raw(ptr as *mut sys::JNIEnv),
+                sys::JNI_EDETACHED => Err(Error::from(ErrorKind::Detached)),
+                _ => unreachable!(),
+            }
+        }
+    }
 }
 
 /// A RAII implementation of scoped guard which detaches the current thread
--- a/src/wrapper/java_vm.rs
+++ b/src/wrapper/java_vm.rs
@@ -14,7 +14,7 @@ unsafe impl Sync for JavaVM {}
 
 impl JavaVM {
     /// Create a JavaVM from a raw pointer.
-    pub fn from_raw(ptr: *mut sys::JavaVM) -> Result<Self> {
+    pub unsafe fn from_raw(ptr: *mut sys::JavaVM) -> Result<Self> {
         non_null!(ptr, "from_raw ptr argument");
         Ok(JavaVM(ptr))
     }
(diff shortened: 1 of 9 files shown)


DarkEld3r:
@madadam I'm happy to see this pull request, because we need this functionality too. :smile: 

Still, I have several concerns: 
1. Ideally `AttachGuard` shouldn't be `Send` (I'm not sure if it is) because  `DetachCurrentThread` must be called from the same thread that called `AttachCurrentThread`.
2. I'm not sure to what lifetime `AttachGuard` is bounded. It seems that it is `JavaVM` lifetime which isn't correct (and convenient).

madadam:
Hi @DarkEld3r , thanks for the feedback.

1. `AttachGuard` isn't `Send`, because it contains `JNIEnv` which itself isn't `Send`.
2. `AttachGuard`'s lifetime is indeed bounded to `JavaVM`. Why do you thing it isn't correct and/or convenient? What would you want it to bind to?

DarkEld3r:
You are probably right about `AttachGuard`: everything is OK with its lifetime. :smile: But I have one more thought: `JavaVM` interface is thread-safe itself, so perhaps it should be `Sync` in Rust too?

madadam:
You are right. I added the `Sync` impl.

jechas01:
Hey, thanks for the PR! Can you update it to exclude the whitespace changes? It's easy for things to get overlooked when they're buried in whitespace.

madadam:
Hey @jechas01 , just wanted to let you know that I reverted the formatting changes. I also rebased the PR against master while I was at it.

src/wrapper/java_vm.rs
```diff
@@ -0,0 +1,87 @@
...
+
+/// The invocation API.
+pub struct JavaVM(*mut sys::JavaVM);
+
+unsafe impl Send for JavaVM {}
+unsafe impl Sync for JavaVM {}
+
+impl JavaVM {
+    /// Create a JavaVM from a raw pointer.
+    pub fn from_raw(ptr: *mut sys::JavaVM) -> Result<Self> {
```
jechas01: This should probably be marked unsafe since, even if the pointer is non-null, it could still be otherwise invalid.

src/wrapper/objects/global_ref.rs
```diff
@@ -57,3 +64,36 @@ impl Drop for GlobalRef {
         }
     }
 }
+
+/// A detached global JVM reference that can be sent across threads. To do
+/// anything useful with it, it must be `attach`ed first.
+///
+/// Warning: detached global ref will leak memory if dropped. Attach to a
+/// `JNIEnv` to prevent this.
```
jechas01: Hmm, this isn't great. Could we perhaps get a JavaVM handle in the detatch method and then use that to get a JNIEnv during Drop so that we can prevent leaks?

madadam:
Hey @jechase , sorry for the delay. I finally got to address your review comments.

jechase:
Cool, looks good! I like the error-code-to-result pattern. Might be worth using it in some of the other jni methods where we're ignoring return codes in favor of checking for a java exception. Could make the actual error a bit clearer in those cases.

PR #48, opened 2017-11-15
Add 'push_local_frame'/'pop_local_frame' functions

DarkEld3r:


src/wrapper/jnienv.rs
```diff
@@ -298,6 +298,23 @@ impl<'a> JNIEnv<'a> {
         })
     }
 
+    /// Creates a new local reference frame, in which at least a given number of local
+    /// references can be created.
+    ///
+    /// Contrary to the `PushLocalFrame` from the `JNI` this function returns `Err` instead
+    /// of a negative value on failure.
+    pub fn push_local_frame(&self, capacity: i32) -> Result<()> {
```
dmitry-timofeev: > from the `JNI`

Do you mean the standard JNI C interface?

src/wrapper/jnienv.rs
```diff
@@ -298,6 +298,23 @@ impl<'a> JNIEnv<'a> {
         })
     }
 
+    /// Creates a new local reference frame, in which at least a given number of local
+    /// references can be created.
+    ///
+    /// Contrary to the `PushLocalFrame` from the `JNI` this function returns `Err` instead
+    /// of a negative value on failure.
```
dmitry-timofeev: I'd add cross-references between `push_local_frame`, `pop_local_frame`, `AutoLocal` & `delete_local_ref` and update their API documentation, because all of them try to solve the same problem, but in a different way. 

@DarkEld3r , @jechas01 , Is that correct to give such recommendations:
 - Prefer `push_local_frame`/`pop_local_frame` if you create a _bounded_ number of local references in a method but can't rely on automatic de-allocation (e.g., in case of recursion or just deep call stacks).
 - Prefer `AutoLocal` in loops.
 - Avoid using `delete_local_reference` directly in favour of the first two approaches?

src/wrapper/jnienv.rs
```diff
@@ -298,6 +298,23 @@ impl<'a> JNIEnv<'a> {
...
+    pub fn push_local_frame(&self, capacity: i32) -> Result<()> {
+        // `PushLocalFrame` returns `jint`, but we don't need it.
+        Ok(jni_void_call!(self.internal, PushLocalFrame, capacity))
+    }
+
+    /// Pops off the current local reference frame, frees all the local references.
```
dmitry-timofeev: add «… all the local references allocated on the current stack frame»?

src/wrapper/jnienv.rs
DarkEld3r: Yes, should I clarify it somehow?

dmitry-timofeev: I checked the original JNI documentation, it says
> Returns 0 on success, a negative number **and** a pending OutOfMemoryError on failure.

I think that jni-rs users are interested in what behaviour to expect from this library rather than the difference between it and C-JNI (the original one).

Maybe write something like: «Returns `Err` on failure, with a pending `OutOfMemoryError`». Native devs _have_ to handle the former and _might_ handle the latter (though it might not be a good idea to hide OoME).

jechas01:
Looks fine to me.

For even more convenience, what about a function like

```rust
fn with_local_frame<F>(&self, capacity: i32, f: F) -> Result<JObject> 
where
    F: Fn() -> Result<JObject>,
{
    self.push_local_frame(capacity)?;
    let res = f();
    match res {
        Ok(obj) => self.pop_local_frame(obj),
        Err(e) => {
            self.pop_local_frame(JObject::null())?;
            Err(e)
        }
    }
}
```

Which could be used like

```rust
env.with_local_frame(32, || {
    // do some stuff
    env.with_local_frame(32, || {
        // do some more stuff
    }
});
```

to make it easier to not forget to pop your frame when finished with it. Plus it even reads more like you're creating a new scope in rust, since you are for your closure.

jechas01:
Also a question: What happens if you leave extra frame pushed once you leave Rust and control returns to the jvm? Does it clean up for you and pop everything like it does when cleaning up local refs?

dmitry-timofeev:
> What happens if you leave extra frame pushed once you leave Rust and control returns to the jvm? 

In case a Java code calls a native method, JVM must delete all created local references when it returns from the native method:


https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/functions.html#local_references

> They [local refs] are freed automatically after the native method returns. 

https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/design.html#referencing_java_objects

> The registry [of local refs] is deleted after the native method returns, allowing all of its entries to be garbage collected.

They won't be deleted automatically only when there are no Java call frames in the stack (native code starts JVM and allocates some Java objects, so there is no such a thing as a return from a native method).

dmitry-timofeev:
> For even more convenience, what about a function like …

Looks great, as I understand, it makes impossible to "forget" to pop the frame. I wonder if it's possible to make an overloaded version for void methods in Rust (like Java methods accepting a [`Supplier<T>`](https://docs.oracle.com/javase/9/docs/api/java/util/function/Supplier.html) & `Runnable`)?

jechas01:
Yeah, I can't think of a good way to accomplish that at the moment... I'd say this is good enough for now!

PR #56, opened 2017-11-29
Fix pop_local_frame return value checking

### LINKED ISSUE #55: How to use pop_local_frame(JObject::null())
I would like to call `env.pop_local_frame(JObject::null())`.

The documentation says that "resulting `JObject` can be `NULL` if `result` is `NULL`".
However, in my case `pop_local_frame` returns an `Err(NullPtr("PopLocalFrame result"))`.

Shouldn't the result be `Ok(something)`, since it's an expected behaviour?

DarkEld3r:
Closes #55.

### DIFF
--- a/src/wrapper/jnienv.rs
+++ b/src/wrapper/jnienv.rs
@@ -325,7 +325,9 @@ impl<'a> JNIEnv<'a> {
     ///
     /// Note that resulting `JObject` can be `NULL` if `result` is `NULL`.
     pub fn pop_local_frame(&self, result: JObject) -> Result<JObject> {
-        Ok(jni_call!(self.internal, PopLocalFrame, result.into_inner()))
+        Ok(unsafe {
+            jni_unchecked!(self.internal, PopLocalFrame, result.into_inner()).into()
+        })
     }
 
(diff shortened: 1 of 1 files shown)


src/wrapper/jnienv.rs
```diff
@@ -325,7 +325,7 @@ impl<'a> JNIEnv<'a> {
     ///
     /// Note that resulting `JObject` can be `NULL` if `result` is `NULL`.
     pub fn pop_local_frame(&self, result: JObject) -> Result<JObject> {
-        Ok(jni_call!(self.internal, PopLocalFrame, result.into_inner()))
+        Ok(jni_non_null_call!(self.internal, PopLocalFrame, result.into_inner()).into())
```
dmitry-timofeev: Shall we use `jni_non_null_call` which checks if any exceptions has occurred for a _safe_ [`PopLocalFrame`](https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/functions.html#PopLocalFrame) (= a method that does not throw exceptions, according to the spec)?

DarkEld3r:
Finally fixed. :sweat_smile:

PR #59, opened 2017-12-04
Implement get_static_field and get_static_field_unsafe

fpoli:


src/wrapper/jnienv.rs
```diff
@@ -1497,6 +1497,88 @@ impl<'a> JNIEnv<'a> {
...
+    ) -> Result<JValue>
+    where
+        T: Desc<'a, JClass<'a>>,
+        U: Desc<'a, JFieldID<'a>>,
+    {
+        let class = class.lookup(self)?.into_inner();
+
+        let field_id = field.lookup(self)?.into_inner();
+
+        // TODO clean this up
```
dmitry-timofeev: Why TODO?

fpoli: Uhm... I copied it from `get_field_unsafe`, with most of the code.

Maybe it refers to the differences between the `match ty { ... }` of `get_field_unsafe` and of `set_field_unsafe`.

src/wrapper/jnienv.rs
```diff
@@ -1497,6 +1497,88 @@ impl<'a> JNIEnv<'a> {
...
+        Ok(match ty {
+            JavaType::Object(_) | JavaType::Array(_) => {
+                let obj: JObject = jni_call!(self.internal, GetStaticObjectField, class, field_id);
+                obj.into()
+            }
+            // JavaType::Object
+            JavaType::Method(_) => unimplemented!(),
```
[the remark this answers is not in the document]
fpoli: I agree but I'm not an expert, I copied it from `get_field_unsafe`. What about replacing it with `return Err(...)`?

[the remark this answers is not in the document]
DarkEld3r: As for me, panic (or even `unreachable!`) is OK here, especially because this function is marked as `unsafe`.

Can such `Err` be handled meaningfully?..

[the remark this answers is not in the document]
dmitry-timofeev: I think `panic` would be better for that's not a kind of error you ever need to handle: if it occurs, you definitely use an API incorrectly (or there is a serious bug in a library itself).

src/wrapper/jnienv.rs
dmitry-timofeev: I see, thank you. I'd ask the maintainer then if it shall remain for any prospective refactoring.

fpoli:
I implemented the `JavaType::Method(_) => panic!(...)`.
The alternative is a `return Err(ErrorKind::WrongJValueType("Method", "see java field").into());`, like the `Primitive::Void` case.

src/wrapper/jnienv.rs
jechase: Yeah, I couldn't think of a good solution to all of the repetition at the time the original was written, hence the TODO. This is fine for now, TODO included. I'll take a look at it when I get around to doing the cleanup.

jechase:
Probably better to return an error for the Method case rather than panic, since we try to avoid panicking on error cases. Otherwise, looks good!

fpoli:
I substituted the `panic!(...)` with a `return Err(...)`

jechase:
OK, rebased for real this time. Not sure if github's "resolve conflicts" feature doesn't actually work or if I just don't understand how it works. It never seems to do what I want it to 😛

Issue #60, opened 2017-12-05
Extend Invocation API support


dmitry-timofeev:
Our team needs a means to launch a JVM from Rust (in C JNI interface that functionality is provided by [the Invocation API](https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/invocation.html), that is already partly supported by jni-rs). I'd like to ask if anybody is currently working on it, and if not, if we could contribute the remaining parts? Are there any nuances we must consider when designing and implementing this functionality?

fpoli:
This is exactly our use case, I'm really interested in this development.

At the moment we are using an adapted version of the [`Jvm::new` function from Rucaja](https://github.com/kud1ing/rucaja/blob/master/src/jvm.rs#L132) to start the JVM, then we switch to `jni-rs` with `JNIEnv::from_raw`. It works fine for us. The only missing part in `jni-rs` is the link to the dynamic library, so we had to:
- add `links = "jvm"` and `build = "build.rs"` to our `Cargo.toml`
- add a `#[link(name = "jvm")] extern "C" {}` somewhere (e.g. `lib.rs`)
- write a `build.rs` that configures the dynamic library location (an example is [here](https://github.com/kud1ing/rucaja/blob/master/build.rs), with a related [issue](https://github.com/kud1ing/rucaja/issues/23))

In general, I think it can be very useful to have a look at [Rucaja](https://github.com/kud1ing/rucaja), because it already implements the "call JVM from Rust" part. A merge of the two would be really great, because most of the functionalities (call a method, call a static method, ...) are duplicated.

dmitry-timofeev:
@fpoli , thank you a lot for the references, we'll definitely look at Rucaja closer to see how to make things work together!

jechase:
OK, let me see if I have this straight, since I haven't looked at the invocation api at all since it's not part of our use case.

1. You want to start a JVM from rust and
2. That requires linking to a the JVM dynamic library

If implemented, it's probably something that we'll want to put behind a cargo feature, since unconditionally requiring linkage to the jvm lib might be undesirable. Otherwise, sounds like a great thing to have, and would make #61 actually doable.

jechase:
I've started work on some additions to the invocation api in the [invocation_api](https://github.com/prevoty/jni-rs/tree/invocation_api) branch. Feel free to comment, test, PR, etc. from there and when we're all satisfied, I'll merge it to master.

dmitry-timofeev:
[...]

PR #63, opened 2017-12-07
Add an integration test that calls java.util.Arrays.binarySearch

fpoli:
Here is an example of an integration test, related to #60. Any feedback is welcome!

### DIFF
--- a/tests/java_integers.rs
+++ b/tests/java_integers.rs
@@ -1,0 +1,67 @@
+extern crate error_chain;
+extern crate jni;
+
+use error_chain::ChainedError;
+use jni::{InitArgsBuilder, JavaVM, JNIEnv};
+use jni::objects::JObject;
+use jni::objects::JValue;
+use jni::sys::JNI_VERSION_1_8;
+
+fn print_exception(env: &JNIEnv) {
+    let exception_occurred = env.exception_check().unwrap_or_else(
+        |e| panic!(format!("{:?}", e)),
+    );
+    if exception_occurred {
+        env.exception_describe().unwrap_or_else(
+            |e| panic!(format!("{:?}", e)),
+        );
+    }
+}
+
+#[test]
+fn test_java_integers() {
+    unsafe {
+        let jvm_args = InitArgsBuilder::new()
+            .version(JNI_VERSION_1_8)
+            .option("-Xcheck:jni")
+            .option("-Xdebug")
+            .build()
+            .unwrap_or_else(|e| {
+                panic!(format!("{}", e.display_chain().to_string()));
+            });
+
+        let (jvm, env) = JavaVM::new(jvm_args).unwrap_or_else(|e| {
+            panic!(format!("{}", e.display_chain().to_string()));
+        });
+
+        let array_length = 50;
+
+        for value in -10..10 {
+            env.with_local_frame(16, || {
+
+                let integer_value = JObject::from(
+                        env.new_object("java/lang/Integer", "(I)V", &[JValue::Int(value)])?
+                );
+
+                let values_array = JObject::from(
+                    env.new_object_array(array_length, "java/lang/Integer", integer_value)?
+                );
+
+                let result = env.call_static_method(
+                    "java/util/Arrays",
+                    "binarySearch",
+                    "([Ljava/lang/Object;Ljava/lang/Object;)I",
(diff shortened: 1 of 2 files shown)


tests/java_integers.rs
```diff
@@ -0,0 +1,81 @@
...
+            .unwrap_or_else(|e| {
+                panic!(format!("{}", e.display_chain().to_string()));
+            });
+
+        let (jvm, env) = JavaVM::new(jvm_args).unwrap_or_else(|e| {
+            panic!(format!("{}", e.display_chain().to_string()));
+        });
+
+        for int_value in -10..10 {
+            for array_length in 1..50 {
```
dmitry-timofeev: I wouldn't put everything in a single test, because two nested loops are a little bit confusing, and hide your intent.

tests/java_integers.rs
```diff
@@ -0,0 +1,81 @@
...
+                env.push_local_frame(0).unwrap_or_else(|e| {
```
dmitry-timofeev: Why a local frame of zero capacity? The docs don't seem to say anything regarding a default capacity: https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/functions.html#PushLocalFrame

tests/java_integers.rs
```diff
@@ -0,0 +1,81 @@
...
+                    print_exception(&env);
+                    panic!(format!("{}", e.display_chain().to_string()));
+                });
+
+                let integer_value = env.new_object("java/lang/Integer", "(I)V", &[JValue::Int(int_value)])
+                    .map(|x| JObject::from(x))
+                    .and_then(|x| env.new_global_ref(x))
```
dmitry-timofeev: Why create global references if you don't need to store them or pass between threads? BTW, in this case (when you create a JVM and keep things in a single main thread) *you* are responsible to delete even the local references (because there is no such thing as a return from a native method: you start and work in a native method, therefore you must tell the GC when you are done with a reference to an object). `env.with_local_frame`, `Push/PopLocalFrame` or `AutoLocal`s help to do that.

tests/java_integers.rs
```diff
@@ -0,0 +1,81 @@
...
+        });
+
+        for int_value in -10..10 {
+            for array_length in 1..50 {
+                env.push_local_frame(0).unwrap_or_else(|e| {
+                    print_exception(&env);
+                    panic!(format!("{}", e.display_chain().to_string()));
+                });
+
+                let integer_value = env.new_object("java/lang/Integer", "(I)V", &[JValue::Int(int_value)])
```
dmitry-timofeev: `initial_value`?

tests/java_integers.rs
```diff
@@ -0,0 +1,81 @@
...
+                    .and_then(|x| env.new_global_ref(x))
+                    .unwrap_or_else(|e| {
+                        print_exception(&env);
+                        panic!(format!("{}", e.display_chain().to_string()));
+                    });
+
+                let int_array =
+                    env.new_object_array(array_length, "java/lang/Integer", integer_value.as_obj())
+                        .map(|x| JObject::from(x))
+                        .and_then(|x| env.new_global_ref(x))
```
dmitry-timofeev: Same as above: this array reference doesn't have to become a global one.

tests/java_integers.rs
```diff
@@ -0,0 +1,81 @@
...
+                    &[JValue::Object(int_array.as_obj()), JValue::Object(integer_value.as_obj())],
+                ).and_then(|x| x.i())
+                    .unwrap_or_else(|e| {
+                        print_exception(&env);
+                        panic!(format!("{}", e.display_chain().to_string()));
+                    });
+
+                assert!(0 <= result && result < array_length);
+
+                env.pop_local_frame(JObject::null()).unwrap_or_else(|e| {
```
dmitry-timofeev: You might consider using [`env.with_local_frame`](https://docs.rs/jni/0.8.1/jni/struct.JNIEnv.html#method.with_local_frame) so that the lib calls `pop_…` for you.

tests/java_integers.rs
fpoli: Yeah it's a bad name, I was referring to the type. I'll rename it

tests/java_integers.rs
fpoli: Thanks! I was not sure about that

fpoli:
It is much more readable now

PR #64, opened 2017-12-08
Invocation API Feature Branch

jechase:
Discuss invocation api related things here.

### DIFF
--- a/src/wrapper/java_vm/init_args.rs
+++ b/src/wrapper/java_vm/init_args.rs
@@ -1,0 +1,103 @@
+use std::os::raw::c_void;
+
+use std::ffi::CString;
+
+use sys::{jint, JNI_VERSION_1_1, JavaVMInitArgs, JavaVMOption};
+
+error_chain! {
+    errors {
+        NullOptString(opt: String) {
+            display("internal null in option: {}", opt)
+            description("internal null in option string")
+        }
+    }
+}
+
+pub struct InitArgsBuilder {
+    opts: Vec<String>,
+    ignore_unrecognized: bool,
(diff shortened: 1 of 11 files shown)


src/wrapper/jnienv.rs
```diff
@@ -101,8 +102,8 @@ impl<'a> JNIEnv<'a> {
     /// will need to be checked against constants from the sys module.
     ///
     /// TODO: convert this to something more usable.
```
dmitry-timofeev: This PR probably makes «This is encoded and …» and a TODO obsolete.

src/wrapper/java_vm/init_args.rs
```diff
@@ -0,0 +1,126 @@
...
+    ///
+    /// Default: V1
+    pub fn version(self, version: JNIVersion) -> Self {
+        let mut s = self;
+        s.version = version.into();
+        s
+    }
+
+    /// Set the `ignoreUnrecognized` init arg flag
+    ///
```
dmitry-timofeev: May add a clarification from the docs:

«If `ignoreUnrecognized` is true, `JavaVM::new` ignores all unrecognized option strings that begin with "-X" or "_". If `ignoreUnrecognized` is false, `JavaVM::new` returns `Err` as soon as it encounters any unrecognized option strings.»

dmitry-timofeev:
FYI, Oracle published the docs for the 9th version: https://docs.oracle.com/javase/9/docs/specs/jni/functions.html

At least, they added a new constant, not sure if there are any other notable changes:
```
#define JNI_VERSION_9   0x00090000
```

Issue #66, opened 2017-12-28
Redesign and reimplement `GlobalRef`


alexander-irbis:
The current implementation of global references in the form of `GlobalRef` /` DetachedGlobalRef` mistakenly assumes that the reference should be attached to the thread. In fact, this is not the case, the same reference can be concurrently used in different threads at the same time.

In addition, the need to destroy an instance of `GlobalRef` or `DetachedGlobalRef` when attaching / detaching is not ergonomic, since it complicates the storage of the reference as part of the other structures.

On the other hand, the task to protect from a race condition by the type system of Rust goes beyond the scope of the functionality of `GlobalRef`, and ultimately is not solved, due to the possibility of creating an additional global reference to the same data (and, it seems, not can be guaranteed to be solved).

Therefore, the functionality of the references and the attachments to threads should be divided into two independent structures that might roughly look like this:

```
#[derive(Clone)]
struct GlobalRef {
    internal: Arc<GlobalRefGuard>,
}

struct GlobalRefGuard {
    vm: JavaVM,
    obj: JObject<'static>,
}


#[derive(Clone)]
struct JNIEnv {
    env: Rc<JNIEnvGuard>,
}
```

`GlobalRef` in such an implementation will allow to re-use the global reference, including between threads, and also automatically release it, as soon as the need for it gone. In addition, reusing an existing global reference, instead of creating a new one each time (by calling `NewGlobalRef`), allows speeding up in some cases, but this speedup can be neglected if dozens of other calls to JNI or heavy computing operations occurs on each copy of the reference.

The structure `JNIEnv` with the use of a similar pattern would be able to transparently hold the attached JVM thread, while we need it somewhere in the native thread, and detach as soon as the need for it gone (or, knowing that we will need it for a long time, we could keep it without detaching, just saving an additional copy of `JNIEnv`). But this is beyond the scope of this issue and requires more detailed discussion and changes in `JavaVM`.

A bit later I will do PR with a new implementation of `GlobalRef`. But, since this is a breaking change, there may be open questions that should be discussed.

For example, should the current implementation of `GlobalRef` + `DetachedGlobalRef` be deprecated and then discarded or `GlobalRef` should remain as it is, but renamed to `AttachedGlobalRef`, because someone heavily depends on it?

dmitry-timofeev:
I'd also add @madadam who contributed the initial Invocation API support, including `DetachedGlobalRef`, to this thread.

PR #67, opened 2018-01-08
[WIP] Reimplement `GlobalRef`

alexander-irbis:
Description of the problem: #66 

This PR depends on `invocation_api` branch.

The current implementation assumes the need to maintain the previous implementation, but introduces breaking changes: old `GlobalRef` is renamed to `AttachedGlobalRef`, old module `global_ref` is renamed to `global_ref_ad`, and old `JNIEnv::new_global_ref()` is renamed to `JNIEnv::new_global_ref_attached()`. I have no answer so far whether it is necessary. It seems to me that it is better to completely remove `AttachedGlobalRef`.

The current `JNIEnv` implementation assumes only a hierarchical execution order, and this also imposes some restrictions on the new implementation of `GlobalRef`. But this issue can be solved separately in a subsequent PR and after discussion.

----

What other tests to add? Ideas are welcome.

----

```
export LD_LIBRARY_PATH=`java -XshowSettings:properties -version 2>&1 > /dev/null | grep 'java.home' | sed 's/^.*= //'`/lib/server/:$LD_LIBRARY_PATH
```
And then
```
cargo test --features="invocation"
```
or
```
cargo test --all-features
```


src/wrapper/jnienv.rs
```diff
@@ -281,13 +282,24 @@ impl<'a> JNIEnv<'a> {
...
+
+
     /// Turns an object into a global ref. This has the benefit of removing the
     /// lifetime bounds since it's guaranteed to not get GC'd by java. It
     /// releases the GC pin upon being dropped.
     pub fn new_global_ref(&self, obj: JObject) -> Result<GlobalRef> {
         non_null!(obj, "new_global_ref obj argument");
         let new_ref: JObject = jni_call!(self.internal, NewGlobalRef, obj.into_inner());
-        let global = unsafe { GlobalRef::new(self.internal, new_ref.into_inner()) };
+        let global = unsafe { GlobalRef::new(self.get_java_vm()?, new_ref.into_inner()) };
```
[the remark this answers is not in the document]
dmitry-timofeev: I would assume that an underlying `JNIEnv` has a pointer to `JavaVM`, therefore, the cost is likely to be negligible.

src/wrapper/objects/global_ref.rs
```diff
@@ -1,143 +1,96 @@
...
 }
 
+
 impl GlobalRef {
-    /// Create a new global reference object. This assumes that
-    /// `CreateGlobalRef` has already been called.
-    pub unsafe fn new(env: *mut sys::JNIEnv, obj: jobject) -> Self {
+    /// Creates a new global reference. This assumes that `NewGlobalRef`
+    /// has already been called.
+    pub unsafe fn new(vm: JavaVM, obj: sys::jobject) -> Self {
```
dmitry-timofeev: Am I right to assume this method is not supposed to be used by the library clients? If so, is it possible in Rust to make it available only inside the library (= hide from the users)?

tests/jni_global_refs.rs
```diff
@@ -0,0 +1,129 @@
...
+
+    let jh1 = {
+        let barrier = barrier.clone();
+        let mut atomic_integer = atomic_integer.clone();
+        spawn(move || {
+            let env = jvm().attach_current_thread().unwrap();
+            barrier.wait();
+            for _ in 0..10000 {
+                unwrap(&env, unwrap(&env, env.call_method(
+                    atomic_integer.as_obj(), "addAndGet", "(I)I", &[JValue::from(-1)])).i());
```
dmitry-timofeev: `decrementAndGet` here, and `incrementAndGet` below?

tests/jni_global_refs.rs
```diff
@@ -0,0 +1,129 @@
...
+
+    let local_ref = AutoLocal::new(&env, unwrap(&env, env.new_object(
+        "java/util/concurrent/atomic/AtomicInteger",
+        "(I)V",
+        &[JValue::from(0)]
+    )));
+
+    let global_ref_1 = unwrap(&env, env.new_global_ref_attached(local_ref.as_obj()));
+
+    {
```
dmitry-timofeev: Since there are multiple `assert`s and conversions between various reference types, I'd add comments clarifying what you are trying to test, e.g.
```
// Test several global refs to the same object work
…
// Test detached & re-attached global ref works
…
// Test the first global ref unaffected by another gr to the same object detached
```

tests/jni_global_refs.rs
```diff
@@ -0,0 +1,129 @@
...
+
+fn unwrap<T>(env: &JNIEnv, res: Result<T>) -> T {
+    res.unwrap_or_else(|e| {
+        print_exception(&env);
+        panic!(format!("{}", e.display_chain().to_string()));
+    })
+}
+
+#[test]
+pub fn global_ref_works_in_other_threads() {
```
dmitry-timofeev: I'd simplify the test to make it easier to understand by reviewers & future maintainers:

1. Use `incrementAndGet` in both threads.
2. Extract a magic `10000` in a constant (e.g., `incrementsInEachThread`, or `numIncrements`).
3. Use the same constant in both loops (no `10000 + 1`, because that is likely to confuse readers).
4. assert that the resulting value is equal to the number of threads (2) multiplied by the number of increments.

dmitry-timofeev:
I wonder what is the deprecation policy:
 1. Shall the old `GlobalRef` simply go away?
 1. Shall it be deprecated first with a `#[deprecated(since = "x.y.z", note="Use NewGlobalRef")]` **and** retain its original name so that the existing users are not affected, and then be turned into `NewGlobalRef` (old `GlobalRef` removed + `NewGlobalRef` renamed into `GlobalRef`)?
 1. Shall it be deprecated **but** get a new name (like in this PR)?

@jechase, @alexander-irbis, @DarkEld3r , @fpoli, what do you think? I think the (2) is the most user-friendly, but requires a little bit more work, the (1) is the easiest.

tests/jni_global_refs.rs
alexander-irbis: 2. Everything went to this, but I somehow forgot to finish it.

tests/jni_global_refs.rs
alexander-irbis: It seemed to me that this was too obvious, but ok, let it be

src/wrapper/jnienv.rs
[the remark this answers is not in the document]
alexander-irbis: In practice, this call is executed one and a half times longer than creating + deleting a copy of `Arc`. We can really neglect it for now.
```
test tests::jni_get_java_vm                      ... bench:       1,789 ns/iter (+/- 431)
test tests::native_arc                           ... bench:       1,138 ns/iter (+/- 269)
```

src/wrapper/objects/global_ref.rs
alexander-irbis: I agree with this, we can make such methods visible only for this crate itself.

src/wrapper/objects/global_ref.rs
```diff
@@ -1,143 +1,96 @@
...
-        GlobalRef {
-            obj: self.obj,
-            env: env.get_native_interface(),
-        }
-    }
-}
+        let res = match self.vm.get_env() {
+            Ok(env) => drop_impl(&env, self.as_obj()),
+            Err(_) => self.vm
+                .attach_current_thread()
```
dmitry-timofeev: Please consider #69

src/wrapper/objects/global_ref_ad.rs
```diff
@@ -0,0 +1,142 @@
...
+
+    fn drop_impl(&self) -> Result<()> {
+        match self.vm.get_env() {
+            Ok(env) => {
+                let _ = self.attach_impl(&env);
+                Ok(())
+            }
+            Err(Error(ErrorKind::ThreadDetached, _)) => {
+                let env = self.vm.attach_current_thread()?;
+                let _ = self.attach_impl(&env);
```
dmitry-timofeev: Please consider #69

fpoli:
> I wonder what is the deprecation policy:
> 1. Shall the old GlobalRef simply go away?
> 2. Shall it be deprecated first with a #[deprecated(since = "x.y.z", note="Use NewGlobalRef")] and retain its original name so that the existing users are not affected, and then be turned into NewGlobalRef (old GlobalRef removed + NewGlobalRef renamed into GlobalRef)?
> 3. Shall it be deprecated but get a new name (like in this PR)?
>
> @jechase, @alexander-irbis, @DarkEld3r , @fpoli, what do you think? I think the (2) is the most user-friendly, but requires a little bit more work, the (1) is the easiest.

I agree. In my opinion it 's ok to just go with (1) and a (major?) version change.

alexander-irbis:
> When the current native thread is detached, it remains detached, the reference becomes unusable.

@dmitry-timofeev currently, references unusable without `JNIEnv` object. This object keeps a thread attachment and provides access to JNI calls. So, it's not possible to use references without an attachment.

dmitry-timofeev:
> … In other words: it is tested by a type system.

You are right, the second test doesn't make much sense with the second reference being a local one.

src/wrapper/objects/mod.rs
```diff
@@ -36,6 +36,10 @@ pub use self::jbytebuffer::*;
 mod global_ref;
 pub use self::global_ref::*;
 
+// For when you want to store a reference to a java object
```
DarkEld3r: "For storing a reference to a java object" if you want the same comment style as below. I'm not sure if it is needed at all.

src/wrapper/objects/global_ref.rs
```diff
@@ -1,143 +1,104 @@
...
-        GlobalRef {
-            obj: self.obj,
-            env: env.get_native_interface(),
-        }
-    }
-}
+        let res = match self.vm.get_env() {
+            Ok(env) => drop_impl(&env, self.as_obj()),
+            Err(_) => {
+                warn!("`GlobalRef` attaches JNI-thread while dropping its instance.");
```
dmitry-timofeev: Since this message is for (potentially) ignorant users, I'd be more verbose and suggest what they shall do:
`Dropping a GlobalRef in a detached thread. Fix your code if this message appears frequently (see the GlobalRef docs)`

Currently it's unclear from the message whether it shall concern a user at all.

src/wrapper/objects/global_ref.rs
```diff
@@ -1,143 +1,104 @@
...
+use sys;
+
 
 /// A global JVM reference. These are "pinned" by the garbage collector and are
 /// guaranteed to not get collected until released. Thus, this is allowed to
-/// outlive the `JNIEnv` that it came from. Still can't cross thread boundaries
-/// since it requires a pointer to the `JNIEnv` to do anything useful with it.
+/// outlive the `JNIEnv` that it came from and can be used in other threads.
+///
+/// Note that the native thread must be `attach`ed to the jni thread (this is so when you have
```
dmitry-timofeev: ```It is _recommended_ that a native thread that drops the global reference is attached to the Java thread (i.e., has an instance of `JNIEnv`).  If the native thread is *not* attached, the `GlobalRef#drop` will print a warning and implicitly `attach` and `detach` it, which significantly affects performance.```
?

jechase:
I'm good with removing the old GlobalRef and making a new semver-incompatible version. It was going to need to happen for the invocation api branch anyway.

jechase:
If this is ready to go in, I can go ahead and merge it and the invocation api branch, assuming there's nothing major that I've overlooked.

PR #68, opened 2018-01-08
Run integration tests on multiple JVMs.

dmitry-timofeev:
- Add a script that sets up proper environment variables and launches
integration tests.
- Run tests with JDK 8 & JDK 9 on the CI service.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,19 +8,16 @@ matrix:
   allow_failures:
     - rust: nightly
 
-jdk:
-  - oraclejdk8
-
-env:
-  global:
-    - JAVA_HOME=/usr/lib/jvm/java-8-oracle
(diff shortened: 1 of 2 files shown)


dmitry-timofeev:
`cargo test …` doesn't rebuild the executable when `JAVA_HOME` changes :thinking: 

I wonder if it's possible to clean test binaries only? I can currently think only of a hack with `$ touch <some source file>`.

alexander-irbis:
> I wonder if it's possible to clean test binaries only?

```
cargo clean -p jni
```

dmitry-timofeev:
Updated the script and added `--verbose` to make sure it works as expected on Travis

dmitry-timofeev:
Just a flaky build that couldn't download sth from Google: https://travis-ci.org/prevoty/jni-rs/jobs/326824201
@jechase , could you restart please?

PR #70, opened 2018-01-10
A test to ensure `DeleteGlobalRef` is called.

alexander-irbis:
As suggested in a comment https://github.com/prevoty/jni-rs/pull/67#pullrequestreview-87438603 this test checks that `DeleteGlobalRef` was called.

Additionally, few auxilliary functions are placed in a separate module. Perhaps some of these functions should be available in the public API for tests in dependent projects.

I renamed `GlobalRef::new` into `from_raw`, it seems more consistent with `JavaVM`, `JNIEnv`, and common patterns of Rust libs; and I made it public to be usable in tests, but this is not so necessary, we can make a helper method/function available only for tests, so `from_raw` can be a private method. But I think `new` renamed to `from_raw` is less confusing for users and while still `unsafe` it does not look like intended for widespread use.

CC @jechase , @dmitry-timofeev , @DarkEld3r , @fpoli

### DIFF
--- a/tests/util/mod.rs
+++ b/tests/util/mod.rs
@@ -1,0 +1,50 @@
+use std::sync::{Arc, Once, ONCE_INIT};
+
+use error_chain::ChainedError;
+use jni::{InitArgsBuilder, JNIEnv, JNIVersion, JavaVM};
+use jni::errors::Result;
+
+
+pub fn jvm() -> &'static Arc<JavaVM> {
+    static mut JVM: Option<Arc<JavaVM>> = None;
+    static INIT: Once = ONCE_INIT;
+
+
+    INIT.call_once(|| {
+        let jvm_args = InitArgsBuilder::new()
+            .version(JNIVersion::V8)
+            .option("-Xcheck:jni")
+            .option("-Xdebug")
+            .build()
+            .unwrap_or_else(|e| {
+                panic!(format!("{}", e.display_chain().to_string()));
+            });
+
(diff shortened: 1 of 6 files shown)


tests/util/mod.rs
```diff
@@ -0,0 +1,50 @@
+use std::sync::{Arc, Once, ONCE_INIT};
+
+use error_chain::ChainedError;
+use jni::{InitArgsBuilder, JNIEnv, JNIVersion, JavaVM};
+use jni::errors::Result;
+
+
+pub fn jvm() -> &'static Arc<JavaVM> {
```
dmitry-timofeev: Looks cool that individual ITs won't have to duplicate that code.

dmitry-timofeev: I wonder if this guy may be easily migrated to `jvm`: https://github.com/prevoty/jni-rs/blob/master/tests/java_integers.rs

alexander-irbis: Honestly, I just did not want to go beyond the "topic" of the PR. 
But it seems really better to immediately make everything uniform.

dmitry-timofeev: I see, that's reasonable for there are various views on PR and commit atomicity. I generally prefer to perform a system-wide refactoring when introducing a new better API that replaces duplicated code. It makes it likely you won't forget to do it, and other developers will use this single API (they copy-paste code, and might copy-paste the old version of it).

PR #74, opened 2018-01-22
Change function parameter of with_local_frame to FnOnce

mehcode:


jechase:
Can I get a little more background on the problem that this solves or why it's needed?

mehcode:
I personally think that we should always strive to use the least binding `Fn` trait we can. As `with_local_frame` only calls its callback once, it makes logical sense to use `FnOnce`.

As for a specific use case..

```rust
let jvm = env.get_java_vm().unwrap();

on(|event| { // registers callback to be called in a background thread on events
  let env = jvm.get_env().unwrap_or_else(|_| jvm.attach_current_thread_as_daemon().unwrap());
  env.with_local_frame(|| { 
    match event { // ERROR: cannot move out of captured outer variable in a Fn closure
      Event::Something { x, y, z } => { ... }
    }
  });
});
```

Pretty trimmed down but hopefully you get the picture. I could work around it a few ways but it seems a much easier solution to just tell the compiler that the callback passed to `with_local_frame` is only called once.

jechase:
Okie doke, works for me!

PR #78, opened 2018-02-24
Fix broken comment in the document

upsuper:


src/lib.rs
```diff
@@ -117,13 +117,8 @@
...
 //! pub extern "system" fn Java_HelloWorld_hello(env: JNIEnv,
-//! // this is the class that owns
-//! our
-//! // static method. Not going to
-//! be
-//! // used, but still needs to
-//! have an
-//!                                              // argument slot
+//! // this is the class that owns our static method. Not going to be used, but
+//! // still needs to have an argument slot.
```
dmitry-timofeev: I'd explain _why_ it must be here:
`but still must be present to match an expected signature of a static native method.`

jechase:
Could you go ahead and turn wrap_comments off in rustfmt.toml while you're at it? It's likely the culprit here.

upsuper:
I guess it's more likely a bug of rustfmt which changes the meaning of the code.

jechase:
Well, more that rustfmt doesn't really know about the content of the
comments, so it just wraps wherever it wants to. I suppose they could parse
the doc comments in the same way that rustdoc does, but it doesn't
currently.

PR #79, opened 2018-02-24
Added `is_instance_of` method

alexander-irbis:


src/wrapper/jnienv.rs
```diff
@@ -169,6 +169,24 @@ impl<'a> JNIEnv<'a> {
         )
     }
 
+    /// Tests whether an object is an instance of a class.
+    pub fn is_instance_of<T>(&self, object: JObject<'a>, class: T) -> Result<bool>
```
dmitry-timofeev: Since we have a native-to-Java integration tests, I'd add at least one positive and one negative test, e.g.:
 * [String](https://docs.oracle.com/javase/9/docs/api/java/lang/String.html) is instance of CharSequence
 * String is not instance of Runnable.
 * `null` is not instance of anything.

src/wrapper/jnienv.rs
```diff
@@ -169,6 +169,29 @@ impl<'a> JNIEnv<'a> {
         )
     }
 
+    /// Tests whether an object is an instance of a class.
+    ///
+    /// _NB: Unlike the operator `isinstanceof`, function `IsInstanceOf` *returns `true`*
+    /// for all classes *if `object` is `null`.*_
+    ///
+    /// https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/functions.html#IsInstanceOf
```
DarkEld3r: Just a nitpick: I prefer to use something like that:
```
See [Java documentation](https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/functions.html#IsInstanceOf) for details.
```

src/wrapper/jnienv.rs
```diff
@@ -169,6 +169,29 @@ impl<'a> JNIEnv<'a> {
         )
     }
 
+    /// Tests whether an object is an instance of a class.
+    ///
+    /// _NB: Unlike the operator `isinstanceof`, function `IsInstanceOf` *returns `true`*
```
dmitry-timofeev: operator ~~`is`~~`instanceof`

src/wrapper/jnienv.rs
```diff
@@ -169,6 +169,29 @@ impl<'a> JNIEnv<'a> {
         )
     }
 
+    /// Tests whether an object is an instance of a class.
```
dmitry-timofeev: I'd consider using the wording from the docs because there is a significant difference between this method & `instanceof`: Returns true if the object reference can be cast to the given type.

tests/util/mod.rs
```diff
@@ -31,20 +28,23 @@ pub fn jvm() -> &'static Arc<JavaVM> {
     unsafe { JVM.as_ref().unwrap() }
 }
 
+pub fn attach_current_jvm_thread() -> AttachGuard<'static> {
```
dmitry-timofeev: `attach_current_thread`? Or something more verbose?

You attach a _native_ thread to the JVM, not a _JVM_ thread to anything.

tests/jni_api.rs
```diff
@@ -0,0 +1,51 @@
...
+    let env = attach_current_jvm_thread();
+    let obj = AutoLocal::new(&env, unwrap(&env, env.new_object(ARITHMETIC_EXCEPTION_CLASS, "()V", &[])));
+    assert!(!unwrap(&env, env.is_instance_of(obj.as_obj(), ERROR_CLASS)));
+}
+
+#[test]
+pub fn is_instance_of_null() {
+    let env = attach_current_jvm_thread();
+    let obj = JObject::null();
+    assert!(obj.is_null());
```
dmitry-timofeev: This `assert` is redundant here: we don't test `is_null`.

tests/jni_api.rs
```diff
@@ -0,0 +1,51 @@
...
+    let env = attach_current_jvm_thread();
+    let obj = AutoLocal::new(&env, unwrap(&env, env.new_object(EXCEPTION_CLASS, "()V", &[])));
+    assert!(!unwrap(&env, env.is_instance_of(obj.as_obj(), ARITHMETIC_EXCEPTION_CLASS)));
+}
+
+#[test]
+pub fn is_instance_of_not_superclass() {
+    let env = attach_current_jvm_thread();
+    let obj = AutoLocal::new(&env, unwrap(&env, env.new_object(ARITHMETIC_EXCEPTION_CLASS, "()V", &[])));
+    assert!(!unwrap(&env, env.is_instance_of(obj.as_obj(), ERROR_CLASS)));
```
dmitry-timofeev: I would not expect the readers (who might not know Java Exception hierarchy well) to know that some exception is not an error, and replace ERROR_CLASS with something obviously different, e.g., a "java/util/ArrayList"

PR #85, opened 2018-03-13
Fix `DetachGlobalRef` usage in the async example

nbaksalyar:
After the `GlobalRef` reimplementation in https://github.com/prevoty/jni-rs/pull/67, the example provided in `example/mylib` stopped working. This pull request fixes the example by removing the calls to the `detach()` and `attach()` methods.

### DIFF
--- a/example/mylib/src/lib.rs
+++ b/example/mylib/src/lib.rs
@@ -131,11 +131,6 @@ pub extern "system" fn Java_HelloWorld_asyncComputation(
     // it to the thread, to prevent it from being collected by the GC.
     let callback = env.new_global_ref(callback).unwrap();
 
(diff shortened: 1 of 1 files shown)


DarkEld3r:
@dmitry-timofeev Apparently, they aren't. It would be awesome to fix that. :upside_down_face:

nbaksalyar:
@DarkEld3r, @dmitry-timofeev,
I think it should be enough to add an `[[example]]` block to the `Cargo.toml`, e.g.:
```
[[example]]
bench = false
name = "mylib"
```
I can do this as a part of this PR if you'd like :slightly_smiling_face:

dmitry-timofeev:
> I can do this as a part of this PR if you'd like 

It would be awesome!

Issue #87, opened 2018-03-23
call_method does not work when Java method returns null


dmitry-timofeev:
If a Java instance method returns null, `JNIEnv#call_method` panics:

```
        thread 'call_method_returning_null' panicked at 'Error: null pointer in CallObjectMethodA result
', tests/util/mod.rs:48:9
```

jechase:
Fixed and tested in #88

PR #88, opened 2018-03-23
Add a test reproducing call_method failing with a method returning null

dmitry-timofeev:
Reproduces #87


jechase:
Cool, test now passes, thanks for putting this in!

PR #98, opened 2018-06-27
Method to retrieve raw JavaVM pointer

paulrouget:


src/wrapper/java_vm/vm.rs
```diff
@@ -44,6 +44,11 @@ impl JavaVM {
         Ok(JavaVM(ptr))
     }
 
+    /// Returns underlying `sys::JavaVM` interface.
+    pub fn get_native_interface(&self) -> *mut sys::JavaVM {
```
dmitry-timofeev: I think 'native interface' might be confused with the JNI interface pointer (JNIEnv): https://docs.oracle.com/javase/10/docs/specs/jni/design.html#jni-interface-functions-and-pointers

I wonder if anyone can come up with a better name. I can think of `get_raw_vm_pointer` or `get_java_vm_pointer`?

src/wrapper/java_vm/vm.rs
```diff
@@ -44,6 +44,11 @@ impl JavaVM {
...
+        self.0
```
dmitry-timofeev: I wonder if it will benefit from integration tests (e.g., a roundtrip test), or is it too simple to break :-) ?

paulrouget: Like, call `get_java_vm_pointer` follow by a `JavaVM::from_raw()`? And check that I can, for example, get back the env?

paulrouget:
I've renamed the function. I'm not sure how to create a useful test though.

src/wrapper/java_vm/vm.rs
dmitry-timofeev: Yes, I thought of something like that, but a test might get too complicated for such a simple getter.

paulrouget:
Anything else needed to get this to land and be published?

paulrouget:
@jechase ping

paulrouget:
ping?

paulrouget:
@dmitry-timofeev can this be published?

dmitry-timofeev:
I believe we'll publish the next week (I don't have the permissions on crates).

PR #106, opened 2018-09-04
Add CODE_OF_CONDUCT, PR template, CONTRIBUTING.md [ECR-890]

vitvakatu:
This PR adds several documents, that will regulate future contributions to the library. We enforce importance of tests and documentation (including changelog), as well as lack of offensive behavior.

These documents are based on documents used in [Exonum](https://github.com/exonum) project.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,73 @@
+# Jni-rs Contributor Code of Conduct
+
+## Our Pledge
+
+In the interest of fostering an open and welcoming environment, we as
+contributors and maintainers pledge to making participation in our project and
+our community a harassment-free experience for everyone, regardless of age, body
+size, disability, ethnicity, gender identity and expression, level of
+experience, nationality, personal appearance, race, religion, or sexual identity
+and orientation.
+
+## Our Standards
+
+Examples of behavior that contributes to creating a positive environment include:
+
+* Using welcoming and inclusive language
+* Being respectful of differing viewpoints and experiences
+* Gracefully accepting constructive criticism
+* Focusing on what is best for the community
+* Showing empathy towards other community members
+
+Examples of unacceptable behavior by participants include:
+
+* The use of sexualized language or imagery and unwelcome sexual attention or
+  advances
+* Trolling, insulting/derogatory comments, and personal or political attacks
+* Public or private harassment
+* Publishing others' private information, such as a physical or electronic
+  address, without explicit permission
+* Other conduct which could reasonably be considered inappropriate in a
+  professional setting
+
+## Our Responsibilities
+
+Project maintainers are responsible for clarifying the standards of acceptable
+behavior and are expected to take appropriate and fair corrective action in
+response to any instances of unacceptable behavior.
+
+Project maintainers have the right and responsibility to remove, edit, or reject
+comments, commits, code, wiki edits, issues, and other contributions that are
+not aligned to this Code of Conduct, or to ban temporarily or permanently any
+contributor for other behaviors that they deem inappropriate, threatening,
+offensive, or harmful.
+
+## Scope
+
+This Code of Conduct applies both within project spaces and in public spaces
+when an individual is representing the project or its community. Examples of
+representing a project or community include using an official project e-mail
+address, posting via an official social media account, or acting as an appointed
+representative at an online or offline event. Representation of a project may be
+further defined and clarified by project maintainers.
+
+## Enforcement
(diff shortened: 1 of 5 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,105 @@
...
+  Integration tests use `jni-rs` as every other Rust application - by importing 
+  library using `extern crate` keyword.
+  
+  ```rust
+    extern crate jni;
+    use jni::*;
+  ```
+  Integration tests typically require running JVM, so you should add 
+  `#![cfg(feature = "invocation")]` at the top of the file. You can use helper
+  methods from [util module](tests/util/mod.rs) to rum JVM.
```
dmitry-timofeev: to run/start a JVM

CONTRIBUTING.md
```diff
@@ -0,0 +1,105 @@
...
+* Integration tests are in [tests directory](tests). They use the same pattern as 
+  unit tests, but split into several files instead of private modules.
+  Integration tests use `jni-rs` as every other Rust application - by importing 
+  library using `extern crate` keyword.
+  
+  ```rust
+    extern crate jni;
+    use jni::*;
+  ```
+  Integration tests typically require running JVM, so you should add
```
dmitry-timofeev: running a JVM

CONTRIBUTING.md
```diff
@@ -0,0 +1,105 @@
...
+  library using `extern crate` keyword.
+  
+  ```rust
+    extern crate jni;
+    use jni::*;
+  ```
+  Integration tests typically require running JVM, so you should add 
+  `#![cfg(feature = "invocation")]` at the top of the file. You can use helper
+  methods from [util module](tests/util/mod.rs) to rum JVM.
+* Doc tests are rarely used, but they allow to efficiently test some functionality
```
dmitry-timofeev: I’d probably add that only one JVM can be started per process: https://docs.oracle.com/javase/10/docs/specs/jni/invocation.html#jni_createjavavm
therefore, tests that need to launch it with different parameters have to be placed in different source files (test from which will be run in separate processes).

CONTRIBUTING.md
```diff
@@ -0,0 +1,105 @@
...
+
+Rust code follows the [Rust style guide](https://github.com/rust-lang-nursery/fmt-rfcs/blob/master/guide/guide.md).
+[`rustfmt`](https://github.com/rust-lang-nursery/rustfmt) enforces the code style.
+
+After installation, you can run it with
+```$sh
+$ cargo fmt --all -- --write-mode=check
+```
+
+Every public entry of the API must be thoroughly documented and covered with tests if it is possible.
```
dmitry-timofeev: I'd probably add that it's a good idea to use the JNI specification as a reference: https://docs.oracle.com/javase/10/docs/specs/jni/index.html

But one shall not copy it for several reasons:
  - licensing
  - different behaviour of most Rust JNI APIs (e.g., they check exceptions for you).

CONTRIBUTING.md
```diff
@@ -0,0 +1,111 @@
...
```
dmitry-timofeev: Shall we also add that the user guide must be updated if needed?

https://github.com/jni-rs/jni-rs/blob/master/src/lib.rs

vitvakatu: I think it's included in "documentation" thing, isn't it?

dmitry-timofeev: I do not insist, but there is an API documentation that is adjacent to methods; and a documentation of this whole crate which works as a user guide. Do you think we shall communicate that here?

vitvakatu: I'm a bit hesitating just because I've never seen this in other projects. I think I will add this point, not an issue

CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,73 @@
...
+when an individual is representing the project or its community. Examples of
+representing a project or community include using an official project e-mail
+address, posting via an official social media account, or acting as an appointed
+representative at an online or offline event. Representation of a project may be
+further defined and clarified by project maintainers.
+
+## Enforcement
+
+Instances of abusive, harassing, or otherwise unacceptable behavior may be
+reported by contacting the project team at ilya.bogdanov@xdev.re. The project
```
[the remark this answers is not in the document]
dmitry-timofeev: :man_shrugging: , perhaps the email is in the templates because one is likely to feel uncomfortable reporting such things on a _public_ issue tracker.

PR #108, opened 2018-09-08
Add a leaking test [DO-NOT-MERGE]

dmitry-timofeev:
## Overview

Run with ./run_leaking_test.sh and see the memory
size grow in the process manager.


rib:
I've rebased this on master to double check that this leak has been fixed by https://github.com/jni-rs/jni-rs/pull/398

As expected; running `cargo test class_lookup_leaks_local_references --features=invocation -- --nocapture` no longer shows the memory usage increasing.

Issue #109, opened 2018-09-08
Memory leaks in methods looking up Class objects


dmitry-timofeev:
## Overview
`JNIEnv` methods accepting `Desc<JClass>` leak a local reference to the Class object if they create a new local reference.

## Description
If an implementation of `Desc#lookup` that creates a new local reference is passed to the `JNIEnv` methods, the created local reference will be leaked (resulting in a leak of native JVM heap and a (potential) leak of referenced Java object). E.g.,

```rust
env.call_static_method("java/lang/Math", "abs", "(I)I", &[JValue::from(-1 as jint)])
```
will leak a single local reference to an instance of `Class<java.lang.Math>` created [here](https://github.com/jni-rs/jni-rs/blob/a1ebdf8b1b685d31936f8fe0d599530b93dc31ff/src/wrapper/jnienv.rs#L868) each time it is called. 

There are about two dozens of `Desc#lookup`s in the codebase.

### Which client code is NOT affected (benign leaks)
Such leaks can be considered benign if all created local references are freed later:
* Refs created in native methods called from Java — freed by the JVM when the native method returns.
* Refs created in a native method attached briefly to the JVM — freed after the thread is detached.
* Refs created in a lambda passed to `with_local_frame` — freed after the last statement in that lambda.

### Which client code IS affected
Code using affected methods in:
* Native threads attached to the JVM for a long or indefinite time. Although such code shall generally use `with_local_frame`/`auto_local` to manage the references created _by the user_, it must be noted that it's not required if it performs only simple calls that do not expose any references that have to be deleted. In other words, it is reasonable to expect that the library methods do not leak — and that's the specified behaviour of the C APIs:
> To ensure that programmers can manually free local references, JNI functions are not allowed to create extra local references, except for references they return as the result.

* Long-running loops.

### How to reproduce
See #108 

### Possible solutions
1. Always perform lookups in a local frame to ensure we delete refs *if* we create them. *Slight performance regression — need to evaluate.*
2. Have methods with distinct signatures that perform lookups and delete manually in them. *A certain UX regression.*
3. Distinguish (?) somehow when we do create a reference during lookup:
[...]

PR #112, opened 2018-09-08
Add a benchmark of a possible solution to #109 [EXPERIMENTAL]

dmitry-timofeev:
## Overview

Benchmark `call_static_method` with extra Push/PopLocalFrame ensuring there are no leaks.

See #109

### Results
```
call_static_method_safe          ... bench:       2,526 ns/iter (+/- 61)
call_static_method_safe_in_frame ... bench:       2,719 ns/iter (+/- 97)
noop_with_local_frame            ... bench:          96 ns/iter (+/- 0)   <-- Result from #110
```

There is ~ 8% performance hit in wrapping the `call_static` in `with_local_frame`

### DIFF
--- a/benches/api_calls.rs
+++ b/benches/api_calls.rs
@@ -94,6 +94,26 @@ mod tests {
         b.iter(|| jni_abs_safe(&env, -3));
     }
 
+    /// A benchmark of a leak-free way to call methods creating local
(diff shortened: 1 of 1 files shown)


rib:
Closing this now that #109 has been addressed in https://github.com/jni-rs/jni-rs/pull/398 (without needing to forcibly insert an extra local frame)

PR #126, opened 2018-09-14
Ignore the test failing on recent versions of Hotspot

dmitry-timofeev:
## Overview
JVM 10+ aborts the process if an invalid reference is passed.

If we want to test such behaviour, a separate test harness for
crashing processes is required.

### Definition of Done


dmitry-timofeev:
> Shouldn't we create an issue for dealing with this problem (fixing or removing the test)?

That’s a good idea: #127

PR #128, opened 2018-09-14
Detach daemon threads automatically

nbaksalyar:
## Overview

This is a breaking API change that is required to make the library more safe to use. Daemon threads are not special in regard to cleanup: they must be detached in the very same way as common threads (by calling `DetachCurrentThread`). Not managing to do so may result in unexpected behaviour and memory leaks.

### Definition of Done

### DIFF
--- a/src/wrapper/java_vm/vm.rs
+++ b/src/wrapper/java_vm/vm.rs
@@ -1,5 +1,5 @@
-use JNIEnv;
 use errors::*;
+use JNIEnv;
 
 use sys;
 
@@ -67,7 +67,7 @@ impl JavaVM {
     }
 
     /// Attaches the current thread to a Java VM as a daemon.
-    pub fn attach_current_thread_as_daemon(&self) -> Result<JNIEnv> {
+    pub fn attach_current_thread_as_daemon(&self) -> Result<AttachGuard> {
         let mut ptr = ptr::null_mut();
         unsafe {
             let res = java_vm_unchecked!(
@@ -78,7 +78,11 @@ impl JavaVM {
             );
             jni_error_code_to_result(res)?;
 
-            JNIEnv::from_raw(ptr as *mut sys::JNIEnv)
+            let env = JNIEnv::from_raw(ptr as *mut sys::JNIEnv)?;
+            Ok(AttachGuard {
+                java_vm: self,
+                env: env,
+            })
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/wrapper/java_vm/vm.rs
```diff
@@ -67,7 +67,7 @@ impl JavaVM {
     }
 
     /// Attaches the current thread to a Java VM as a daemon.
```
dmitry-timofeev: Shall we clarify here in the docs that the AttachGuard will detach the daemon thread when it goes out of the scope? I guess that this operation is mostly used to attach a thread as a daemon for _some_ time — for longer than a single method execution, hence one will have to remember to keep the guard for that period of time somewhere.

dmitry-timofeev:
Now that I think of it, although a thread being a daemon does not directly imply it’s long-running, but could anyone think of plausible scenarious when a daemon thread shall not run indefinitely (the only diffence between 'regular' and daemon threads is that the latter do not prevent the VM from termination)?

nbaksalyar:
Thank you for the review comments and suggestions, and sorry for taking long to implement this.
I'll update & rebase the PR today.

dmitry-timofeev:
@nbaksalyar, thanks, that's great! Would you please review the recent https://github.com/jni-rs/jni-rs/pull/139/files? It might be helpful as an example, and I think the present code for `attach_as_daemon` might be susceptible to the same problem of nested calls to `attach_x_thread` creating a guard that detaches an earlier-attached thread.

Also, the docs say that a _nested_ attach does not change the daemon status of an already attached thread. I think that makes sense (or there will be surprising changes in a thread status), and must be reflected in our documentation. Probably, the other viable option is to forbid nested attaches altogether, because if they occur in the client code, it might be there is no understanding of when and how the threads are attached/detached.

> If the thread has already been attached via either AttachCurrentThread or AttachCurrentThreadAsDaemon, this routine simply sets the value pointed to by penv to the JNIEnv of the current thread. _In this case neither AttachCurrentThread nor this routine have any effect on the daemon status of the thread._

dmitry-timofeev:
Hi @nbaksalyar 

We are currently considering implementing an automatic detach of native threads by putting guards in TLS — first of all, daemon threads, as there is little reason to use daemons unless they are blocking/long-running. Please see #179 for discussions on the feature.

dmitry-timofeev:
@nbaksalyar , thanks for the patch, closing as superseded by #193 (must be available in 0.13 in a week or so)

PR #129, opened 2018-09-17
Setup projects for Java integration tests

dmitry-timofeev:
## Overview

Setup the projects for Java integration tests.

Layout:
- `java-tests` for Java ITs using the native library
- `java-testlib` with implementations of native methods from ^
- `java-benchmarks` using `java-tests` with JMH setup
- `jni` — is the root for the library
- `example` — is now included in the workspace to avoid `jni` rebuilds

Travis now builds and tests (on top of `jni`):
- `example`
- `java-tests`

### Definition of Done

- [ ] Update the Contribution Guide


mvnw.cmd
```diff
@@ -0,0 +1,161 @@
...
+@REM MAVEN_BATCH_ECHO - set to 'on' to enable the echoing of the batch commands
+@REM MAVEN_BATCH_PAUSE - set to 'on' to wait for a key stroke before ending
+@REM MAVEN_OPTS - parameters passed to the Java VM when running Maven
+@REM     e.g. to debug Maven itself, use
+@REM set MAVEN_OPTS=-Xdebug -Xrunjdwp:transport=dt_socket,server=y,suspend=y,address=8000
+@REM MAVEN_SKIP_RC - flag to disable loading of mavenrc files
+@REM ----------------------------------------------------------------------------
+
+@REM Begin all REM lines with '@' in case MAVEN_BATCH_ECHO is 'on'
+@echo off
```
alexander-irbis: Perhaps it's worth to move this line to the beginning of the file and remove all "@" before REM, setlocal, etc?

dmitry-timofeev: I actually have no idea what that means because this file is added as is by https://github.com/takari/maven-wrapper :-) 

Do you think they must update anything?

alexander-irbis: @ at the beginning of the line suppresses echo of the command in the line.
echo off - do the same for all following lines to not place `@` in each line.
`@echo off` - typically the first line in almost all batch files since ancient DOS times.

alexander-irbis: But if this is autogenerated file, it makes no sense to fix it...

rib:
While this PR is now over 7 years old it's unfortunately very out of date with respect to the current repo.

We can maybe also say that this has been addressed through the addition of the `javac` crate and the addition of tests (mainly for the jni-macros crate) that will compile java code as part of the test which is then loaded via define_class so we can test custom bindings as well as test native method bindings.

PR #136, opened 2018-09-27
Macro usage enhancements

skletsun:
## Overview

A few updates on macro usage for JNI methods calls

### Definition of Done


tests/jni_api.rs
```diff
@@ -143,3 +149,54 @@ pub fn with_local_frame_pending_exception() {
...
+        .expect("JNIEnv#byte_array_from_slice must create a java array from slice");
+    let obj = AutoLocal::new(&env, JObject::from(java_array));
+
+    assert!(!obj.as_obj().is_null());
+    let mut res: [i8; 3] = [0; 3];
+    env.get_byte_array_region(java_array, 0, &mut res).unwrap();
+    assert_eq!(res[0], 1);
+    assert_eq!(res[1], 2);
+    assert_eq!(res[2], 3);
+}
```
DarkEld3r: How about a new line at the end of the file? :upside_down_face:

src/wrapper/jnienv.rs
```diff
@@ -623,6 +624,7 @@ impl<'a> JNIEnv<'a> {
         // TODO clean this up
         Ok(match ret {
             JavaType::Object(_) | JavaType::Array(_) => {
+                // TODO shouldn't it be a non_void instead of non_null here?
```
dmitry-timofeev: Yes, it should — Java methods may return null.

src/wrapper/jnienv.rs
```diff
@@ -636,64 +638,64 @@ impl<'a> JNIEnv<'a> {
             JavaType::Method(_) => unimplemented!(),
             JavaType::Primitive(p) => {
                 let v: JValue = match p {
-                    Primitive::Boolean => (jni_unchecked!(
+                    Primitive::Boolean => (jni_non_void_call!(
```
dmitry-timofeev: Just wondering if a macro expansion affects the code size significantly — does it make sense to add a single `exception_check` below?

src/wrapper/jnienv.rs
```diff
@@ -1048,16 +1050,15 @@ impl<'a> JNIEnv<'a> {
...
-                self.internal,
-                SetByteArrayRegion,
-                bytes,
-                0,
-                length,
-                buf.as_ptr() as *const i8
-            );
-        }
+
+        jni_void_call!(
```
dmitry-timofeev: I believe it is not needed because *we* create an array of the proper size a line above, so no `ArrayIndexOutOfBoundsException` is possible.

src/wrapper/jnienv.rs
```diff
@@ -591,6 +591,7 @@ impl<'a> JNIEnv<'a> {
 
     /// Get the class for an object.
     pub fn get_object_class(&self, obj: JObject) -> Result<JClass<'a>> {
+        non_null!(obj, "get_object_class");
         Ok(jni_non_null_call!(self.internal, GetObjectClass, obj.into_inner()))
```
dmitry-timofeev: This doesn't seem to require exception check (it [does not throw](https://docs.oracle.com/javase/10/docs/specs/jni/functions.html#getobjectclass)).

tests/jni_api.rs
```diff
@@ -12,6 +13,11 @@ static ARRAYLIST_CLASS: &str = "java/util/ArrayList";
 static EXCEPTION_CLASS: &str = "java/lang/Exception";
 static ARITHMETIC_EXCEPTION_CLASS: &str = "java/lang/ArithmeticException";
 static INTEGER_CLASS: &str = "java/lang/Integer";
+static CLASS_MATH: &str = "java/lang/Math";
```
dmitry-timofeev: I'd suggest using the same naming pattern as above:
```
MATH_CLASS
MATH_ABS_METHOD_NAME
MATH_ABS_SIGNATURE
…
```

tests/jni_api.rs
```diff
@@ -143,3 +149,54 @@ pub fn with_local_frame_pending_exception() {
...
+    assert_eq!(val, 10);
+}
+
+#[test]
+pub fn call_static_method_throws() {
+    let env = attach_current_thread();
+
+    let x = JValue::Long(4_000_000_000);
+    env.call_static_method(CLASS_MATH, METHOD_MATH_TO_INT, SIG_MATH_TO_INT, &[x])
+        .expect_err("JNIEnv#call_static_method_unsafe should return error");
```
dmitry-timofeev: Can we check that the error is JavaException?

https://github.com/jni-rs/jni-rs/blob/fec275ffa42353778ad34ecb903c9b45336517b9/src/wrapper/errors.rs#L32-L34

tests/jni_api.rs
```diff
@@ -143,3 +149,54 @@ pub fn with_local_frame_pending_exception() {
...
+pub fn java_byte_array_from_slice() {
+    let env = attach_current_thread();
+    let buf: &[u8] = &[1, 2, 3];
+    let java_array = env.byte_array_from_slice(buf)
+        .expect("JNIEnv#byte_array_from_slice must create a java array from slice");
+    let obj = AutoLocal::new(&env, JObject::from(java_array));
+
+    assert!(!obj.as_obj().is_null());
+    let mut res: [i8; 3] = [0; 3];
+    env.get_byte_array_region(java_array, 0, &mut res).unwrap();
```
dmitry-timofeev: This method implementation looks suspicios (but that may be fixed in a separate PR): won't it cause overflow of the buffer if non-zero start position is passed?

https://github.com/jni-rs/jni-rs/blob/56a14b441ca3f15e9a0db302d16b7af96bb302ee/src/wrapper/jnienv.rs#L1154-L1170

dmitry-timofeev:
This one is breaking — it changes the behaviour of existing code (fixes it to be correct).

src/wrapper/jnienv.rs
skletsun: Yes, and I reflected this in the spreadsheet. Do you want me to correct this within this PR?

src/wrapper/jnienv.rs
skletsun: I don't think it makes sense changing it back to the `jni_unchecked` with an additional line for exceptions checking, `jni_non_void_call` does exactly what we need and handles the return value "for free".

src/wrapper/jnienv.rs
skletsun: Yeap, I was also thinking about this and wasn't actually sure whether this checking is redundant.

src/wrapper/jnienv.rs
dmitry-timofeev: Yes, please.

tests/jni_api.rs
dmitry-timofeev: > Most probably it will trigger the ArrayIndexOutOfBoundsException in this case.

Not necessarily — we do not check the source array length here; but `buf` overflow — most certainly.

src/wrapper/jnienv.rs
```diff
@@ -591,7 +591,8 @@ impl<'a> JNIEnv<'a> {
 
     /// Get the class for an object.
     pub fn get_object_class(&self, obj: JObject) -> Result<JClass<'a>> {
-        Ok(jni_non_null_call!(self.internal, GetObjectClass, obj.into_inner()))
+        non_null!(obj, "get_object_class");
```
dmitry-timofeev: I'd consider adding a test for that.

PR #139, opened 2018-10-02
Fixed the issue with early detaching of a thread by nested AttachGuard.

skletsun:
## Overview

The current implementation of AttachGuard detaches thread unconditionally which can lead to the issue with early thread detach in case of nested AttachGuard. Fixed it by adding checking for the state of the current thread before attaching it to JVM.

### Definition of Done


dmitry-timofeev:
It makes sense to add this one to changelog, too (in Fixed section).

Issue #148, opened 2018-10-08
Unify null pointer handling


dmitry-timofeev:
## Overview

Currently null pointers are handled inconsistently: some methods that are specified as potentially returning `null` convert it to `Err` whilst others return `JObject::null()` (or appropriate wrapper).

### Description

It seems that initially the intention was to always convert `null`s to `Err(ErrorKind::NullPtr)` — but it confused some users (including myself):
  - dd5f3a5ab9809aa1c3fa664161a7deb3ae457955 (jni_non_null_call at that revision does **not** check for nulls) and #87 
  - #136 

The present [documentation](https://docs.rs/jni/0.10.2/jni/struct.JNIEnv.html) says

> Because null pointers are a thing in Java, this also converts them to an Err result with the kind NullPtr. This may occur when either a null argument is passed to a method or when a null would be returned. Where applicable, the null error is changed to a more applicable error type, such as MethodNotFound.

It must be decided how Java `null` pointer must be represented in jni-rs and the present APIs must be updated to use that representation consistently.

skletsun:
The list of methods that communicate return value of Null as Err(NullPtr) for corresponding JNI methods:
- get_superclass()
- define_class()
- find_class()
- alloc_object()
- new_direct_byte_buffer()
- get_method_id()
- get_static_method_id()
- get_field_id()
- get_field_unchecked() for GetObjectField
- get_static_field_id()
- new_object_unchecked()
- get_string_utf_chars()
- new_string()
- new_object_array()
- get_object_array_element()
- new_boolean_array()
- new_byte_array()
- new_char_array()
- new_short_array()
- new_int_array()
- new_long_array()
- new_float_array()
- new_double_array()

dmitry-timofeev:
Thanks, I think there are at least two categories in which the underlying functions could be split up:
1. Returning `null` as a possible reference value (like  `get_object_array_element`, `get_field_unchecked`).
2. Returning `null` in case of an error (e.g., `new_int_array` which never returns `null` if everything is OK).

The first category is the one that triggered some changes (see the issues linked in the descriptions), and it makes sense to make the behaviour of methods in this category consistent _first_. Otherwise, it is rather surprising to see `get_static_field_unsafe -> null` whilst `get_field_unchecked -> Err(NullPtr)`.

Therefore, I'd suggest the following:
1. [ ] Determine where each method belongs.
[...]

PR #150, opened 2018-10-09
Further improvements of macro usage for JNI calls 

### LINKED ISSUE #141: Further improvements of macro usages for JNI calls
Initial improvements have been presented in the #136. There is a [shared spreadsheet](https://docs.google.com/spreadsheets/d/16tfgUxf_A2e7WV504Ml273bPsF_xJc6_zRaGjKrviro/edit#gid=825321690) with results of inspection of macro usages that could be used as a source of further enhancements.

skletsun:
## Overview

Further inspection and improvements of the current implementation of methods that calls JNI functions. Fixes #141. 

### Definition of Done

### DIFF
--- a/src/wrapper/jnienv.rs
+++ b/src/wrapper/jnienv.rs
@@ -1081,12 +1081,6 @@ impl<'a> JNIEnv<'a> {
         Ok(vec)
     }
 
-    // NB: Oracle doc states that the family of New<PrimitiveType>Array routines does not throw any
-    // exception but just return NULL in case of error. In fact, if one calls any of this functions
-    // with negative value as an argument for the length of array it ends up with the pending
-    // java.lang.NegativeArraySizeException exception. That's why we use macro that checks for
-    // exceptions and NULL.
-    // ??? Wouldn't it be more optimal to check for negative argument before calling managed code???
     /// Create a new java boolean array of supplied length.
     pub fn new_boolean_array(&self, length: jsize) -> Result<jbooleanArray> {
         let array: jbooleanArray = jni_non_null_call!(self.internal, NewBooleanArray, length);
--- a/src/wrapper/jnienv.rs
+++ b/src/wrapper/jnienv.rs
@@ -286,8 +286,10 @@ impl<'a> JNIEnv<'a> {
     /// Returns the starting address of the memory of the direct
     /// java.nio.ByteBuffer.
     pub fn get_direct_buffer_address(&self, buf: JByteBuffer) -> Result<&mut [u8]> {
+        non_null!(buf, "get_direct_buffer_address argument");
         let ptr: *mut c_void =
             unsafe { jni_unchecked!(self.internal, GetDirectBufferAddress, buf.into_inner()) };
+        non_null!(ptr, "get_direct_buffer_address return value");
         let capacity = self.get_direct_buffer_capacity(buf)?;
         unsafe { Ok(slice::from_raw_parts_mut(ptr as *mut u8, capacity as usize)) }
     }
--- a/src/wrapper/jnienv.rs
+++ b/src/wrapper/jnienv.rs
@@ -274,14 +274,12 @@ impl<'a> JNIEnv<'a> {
 
     /// Create a new instance of a direct java.nio.ByteBuffer.
     pub fn new_direct_byte_buffer(&self, data: &mut [u8]) -> Result<JByteBuffer> {
-        let obj = unsafe {
-            jni_unchecked!(
-                self.internal,
-                NewDirectByteBuffer,
-                data.as_mut_ptr() as *mut c_void,
-                data.len() as jlong
-            )
-        };
+        let obj: JObject = jni_non_null_call!(
+            self.internal,
+            NewDirectByteBuffer,
+            data.as_mut_ptr() as *mut c_void,
+            data.len() as jlong
+        );
         Ok(JByteBuffer::from(obj))
     }
 
--- a/src/wrapper/jnienv.rs
+++ b/src/wrapper/jnienv.rs
@@ -170,8 +170,7 @@ impl<'a> JNIEnv<'a> {
         T: Desc<'a, JClass<'a>>,
     {
         let class = class.lookup(self)?;
-        // GetSuperclass may return NULL for "java/lang/Object" and this is expected behavior
-        Ok(jni_non_void_call!(self.internal, GetSuperclass, class.into_inner()).into())
+        Ok(jni_non_null_call!(self.internal, GetSuperclass, class.into_inner()))
     }
 
     /// Tests whether class1 is assignable from class2.
--- a/src/wrapper/jnienv.rs
+++ b/src/wrapper/jnienv.rs
@@ -1090,7 +1090,6 @@ impl<'a> JNIEnv<'a> {
             length,
             vec.as_mut_ptr() as *mut i8
         );
-        check_exception!(self.internal);
         Ok(vec)
     }
 
--- a/src/wrapper/jnienv.rs
+++ b/src/wrapper/jnienv.rs
@@ -1081,6 +1081,12 @@ impl<'a> JNIEnv<'a> {
         Ok(vec)
     }
 
+    // NB: Oracle doc states that the family of New<PrimitiveType>Array routines does not throw any
+    // exception but just return NULL in case of error. In fact, if one calls any of this functions
+    // with negative value as an argument for the length of array it ends up with the pending
+    // java.lang.NegativeArraySizeException exception. That's why we use macro that checks for
+    // exceptions and NULL.
+    // ??? Wouldn't it be more optimal to check for negative argument before calling managed code???
     /// Create a new java boolean array of supplied length.
     pub fn new_boolean_array(&self, length: jsize) -> Result<jbooleanArray> {
         let array: jbooleanArray = jni_non_null_call!(self.internal, NewBooleanArray, length);
--- a/src/wrapper/jnienv.rs
+++ b/src/wrapper/jnienv.rs
@@ -144,7 +144,8 @@ impl<'a> JNIEnv<'a> {
         T: Desc<'a, JClass<'a>>,
     {
         let class = class.lookup(self)?;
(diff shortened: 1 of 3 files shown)


tests/jni_api.rs
```diff
@@ -228,3 +227,160 @@ pub fn get_object_class_null_arg() {
...
+    assert!(!result.is_null());
+
+    let capacity = env.get_direct_buffer_capacity(result).unwrap();
+    assert_eq!(capacity, 4);
+}
+
+#[test]
+pub fn get_direct_buffer_capacity_wrong_arg() {
+    let env = attach_current_thread();
+    let capacity = env.get_direct_buffer_capacity(JObject::null().into());
```
dmitry-timofeev: The spec says "must not be null". 

> buf: a direct java.nio.Buffer object, must not be NULL.

I'd consider passing another object

src/wrapper/jnienv.rs
```diff
@@ -274,22 +274,22 @@ impl<'a> JNIEnv<'a> {
...
         Ok(JByteBuffer::from(obj))
     }
 
     /// Returns the starting address of the memory of the direct
     /// java.nio.ByteBuffer.
     pub fn get_direct_buffer_address(&self, buf: JByteBuffer) -> Result<&mut [u8]> {
+        non_null!(buf, "get_direct_buffer_address argument");
         let ptr: *mut c_void =
             unsafe { jni_unchecked!(self.internal, GetDirectBufferAddress, buf.into_inner()) };
+        non_null!(ptr, "get_direct_buffer_address return value");
```
dmitry-timofeev: I wonder if we shall have two macros:
- `non_null` — when `NULL` is strictly not allowed;
- and `null_to_err` — when it is allowed, but must be converted to an `Err` (as in this and many other cases)

They _may_ or may not have the same implementation, but document the intent clearer. The last macro `null_to_err` is also applicable to cases where we check preconditions: the null argument is "not allowed", but `Err` is a natural way to report that in Rust, if I understand correctly.

@vitvakatu , @DarkEld3r , WDYT?

src/wrapper/jnienv.rs
```diff
@@ -1078,6 +1081,12 @@ impl<'a> JNIEnv<'a> {
         Ok(vec)
     }
 
+    // NB: Oracle doc states that the family of New<PrimitiveType>Array routines does not throw any
+    // exception but just return NULL in case of error. In fact, if one calls any of this functions
+    // with negative value as an argument for the length of array it ends up with the pending
+    // java.lang.NegativeArraySizeException exception. That's why we use macro that checks for
+    // exceptions and NULL.
+    // ??? Wouldn't it be more optimal to check for negative argument before calling managed code???
```
dmitry-timofeev: The spec also says they [throw](https://docs.oracle.com/javase/10/docs/specs/jni/functions.html#newprimitivetypearray-routines)

> OutOfMemoryError: if the system runs out of memory.

But in that case you will both get `NULL` and a pending exception.

tests/jni_api.rs
```diff
@@ -228,3 +227,160 @@ pub fn get_object_class_null_arg() {
...
+    assert_pending_java_exception(&env);
+
+    let result = env.new_double_array(WRONG_SIZE);
+    assert!(result.is_err());
+    assert_error(result, "JNIEnv#new_double_array should throw exception");
+    assert_pending_java_exception(&env);
+}
+
+// Helper method that asserts that result is Error and the cause is JavaException.
+fn assert_error(res: Result<jobject, Error>, expect_message: &str) {
```
dmitry-timofeev: I'd pick a more specific name — not any Err is accepted, but the one corresponding to an exception. Maybe, `assert_exception`, that will also delegate to `assert_pending_exception`?

src/wrapper/jnienv.rs
vitvakatu: What do you mean by `allowed, but must be converted to an Err`? I think the only valid behavior would be returning `Err` immediately in case of null.

dmitry-timofeev: > allowed, but must be converted to an Err

By that I mean a situation when a JNI method is specified to return `null` to indicate an error. By NULL being strictly not allowed I mean a method that is specified to **never** return `null` — I doubt there are many of them.

How to treat preconditions — that’s a separate concern. Some arguments are specified to be strictly non-NULL, but I am not sure if we shall panic in that case. 

Finally, any _Java_ method may return `null` (and any field can be `null`), and it is not an indication that it has failed, please see #148 for a more detailed discussion.

src/wrapper/jnienv.rs
skletsun: It's funny, but [spec for version 8](https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/functions.html#New_PrimitiveType_Array_routines) says nothing about throw :-). Thanks, removed the comment.

tests/jni_api.rs
skletsun: Yes, it's more related to the exception, I agree. Thanks, renamed to assert_exception.

tests/jni_api.rs
```diff
@@ -371,6 +371,30 @@ pub fn new_primitive_array_wrong() {
...
+fn get_super_class_null() {
+    let env = attach_current_thread();
+    let class = env.get_superclass("java/lang/Object").unwrap();
+    assert!(class.is_null());
+}
+
+#[test]
+fn get_super_class_wrong() {
+    let env = attach_current_thread();
+    // this triggers the "java.lang.ClassNotFoundException" (even if spec says it doesn't throw)
```
dmitry-timofeev: The thing is — the method described in the spec does not take a string, but a reference to a class instance. Here a lookup of a class fails, not `get_super_class`.

src/wrapper/jnienv.rs
```diff
@@ -1068,17 +1068,15 @@ impl<'a> JNIEnv<'a> {
...
-                self.internal,
-                GetByteArrayRegion,
-                array,
-                0,
-                length,
-                vec.as_mut_ptr() as *mut i8
-            );
-        }
-        check_exception!(self.internal);
+        jni_void_call!(
```
dmitry-timofeev: I believe `jni_unchecked` is OK here because *this method* determines the start and end indices, therefore, `ArrayIndexOutOfBoundsException` is not possible.

skletsun: Yes, but using jni_void! looks more generic here from my POV because it does almost the same.

tests/jni_api.rs
skletsun: Thanks, corrected.

src/wrapper/jnienv.rs
dmitry-timofeev: I'd agree in application code, but in the library code I'd avoid redundant operations if it is possible :-)

src/wrapper/jnienv.rs
```diff
@@ -1068,15 +1068,17 @@ impl<'a> JNIEnv<'a> {
...
+            jni_unchecked!(
+                self.internal,
+                GetByteArrayRegion,
+                array,
+                0,
+                length,
+                vec.as_mut_ptr() as *mut i8
+            );
+        }
+        check_exception!(self.internal);
```
dmitry-timofeev: > I believe jni_unchecked is OK here because this method determines the start and end indices, therefore, ArrayIndexOutOfBoundsException is not possible.

Why adding a check for an impossible exception?

skletsun: Ah, you mean this. Ok, it was introduced more than a year ago within #15 and, most probably, this check is redundant. Going to remove it.

tests/jni_api.rs
```diff
@@ -249,3 +248,201 @@ pub fn get_object_class_null_arg() {
...
+    let result = env.delete_local_ref(null_obj);
+    assert!(result.is_ok());
+}
+
+// Helper method that asserts that result is Error and the cause is JavaException.
+fn assert_exception(res: Result<jobject, Error>, expect_message: &str) {
+    assert!(res.is_err());
+    assert!(res.map_err(|error| match *error.kind() {
+        ErrorKind::JavaException => true,
+        _ => false,
```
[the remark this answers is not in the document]
vitvakatu: Never mind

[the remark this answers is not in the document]
skletsun: 👍

src/wrapper/jnienv.rs
```diff
@@ -330,10 +329,13 @@ impl<'a> JNIEnv<'a> {
...
-        non_null!(obj, "new_global_ref obj argument");
-        let new_ref: JObject = jni_non_null_call!(self.internal, NewGlobalRef, obj.into_inner());
-        let global = unsafe { GlobalRef::from_raw(self.get_java_vm()?, new_ref.into_inner()) };
-        Ok(global)
+        let new_ref: JObject = jni_unchecked!(self.internal, NewGlobalRef, obj.into_inner()).into();
+        if !new_ref.is_null() {
+            let global = unsafe { GlobalRef::from_raw(self.get_java_vm()?, new_ref.into_inner()) };
+            Ok(global)
+        } else {
+            Err(ErrorKind::NullPtr("new_global_ref").into())
```
dmitry-timofeev: This seems like a more verbose version of the previous code that checked the argument for null. The spec allows to "create" and "delete" `null` global refs (it won't create anything, but return a `null` from new and do nothing on delete) — https://docs.oracle.com/en/java/javase/11/docs/specs/jni/functions.html#newglobalref

If some method returns a reference that may be `null`, and a user needs to store that nullable reference in a `GlobalRef`, they will get an error instead.

dmitry-timofeev: Basically, that's the same question of how to deal with `null`s consistently: #148 

I’d go with a symmetric solution to local references, and reconsider in #148

tests/jni_api.rs
```diff
@@ -418,6 +418,28 @@ fn convert_byte_array() {
     assert_eq!(dest.unwrap(), src);
 }
 
+#[test]
+fn local_ref_null() {
```
dmitry-timofeev: Would you please add a test with auto_local to ensure it works correctly with these changes?

PR #152, opened 2018-11-01
Fixed calculation of buffer length provided to Get*ArrayRegion() and Set*ArrayRegion() JNI functions

### LINKED ISSUE #142: get_byte_array_region passes incorrect length
get_byte_array_region will cause overflow of the buffer if non-zero start position is passed.

https://github.com/jni-rs/jni-rs/blob/56a14b441ca3f15e9a0db302d16b7af96bb302ee/src/wrapper/jnienv.rs#L1154-L1170

See also: https://github.com/jni-rs/jni-rs/pull/136#discussion_r221935935

koutheir:
## Overview

Fixed calculation of buffer length provided to `Get*ArrayRegion()` and `Set*ArrayRegion()` JNI functions.

Fixes #142.

### Definition of Done

### DIFF
--- a/src/wrapper/jnienv.rs
+++ b/src/wrapper/jnienv.rs
@@ -1136,6 +1136,30 @@ impl<'a> JNIEnv<'a> {
         Ok(array)
     }
 
+    /// Calculate the correct buffer length to provide to the `Get*ArrayRegion()` and
+    /// `Set*ArrayRegion()` JNI functions, based on the array size, the start index, and the length
+    /// of the provided buffer.
+    fn proper_length_for_primitive_array_region(
+        &self,
+        array: jarray,
+        start: jsize,
+        buf_len: usize,
+    ) -> Result<jsize> {
+        let array_len: jsize = jni_unchecked!(self.internal, GetArrayLength, array);
+
+        // buffer length should fit in a `jsize`.
+        let usable_buf_len = usize::min(buf_len, jsize::max_value() as usize) as jsize;
+
+        let len = if start >= 0 && start <= array_len {
+            // `start` is a valid index into the array. Calculate a proper `len`.
+            jsize::min(usable_buf_len, array_len - start)
+        } else {
+            // The JVM will throw an `ArrayIndexOutOfBoundsException` due to `start`.
+            usable_buf_len
+        };
+        Ok(len)
+    }
+
     /// Copy elements of the java boolean array from the `start` index to the
     /// `buf` slice.
     pub fn get_boolean_array_region(
@@ -1145,12 +1169,13 @@ impl<'a> JNIEnv<'a> {
         buf: &mut [jboolean],
     ) -> Result<()> {
         non_null!(array, "get_boolean_array_region array argument");
+        let len = self.proper_length_for_primitive_array_region(array, start, buf.len())?;
         jni_void_call!(
             self.internal,
             GetBooleanArrayRegion,
             array,
             start,
-            buf.len() as jsize,
+            len,
             buf.as_mut_ptr()
         );
         Ok(())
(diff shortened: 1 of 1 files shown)


koutheir:
For information, the Hotspot JVM in OpenJDK 8 defines the condition to throw an `ArrayIndexOutOfBoundsException` as follows:
```c++
start < 0 || len < 0 || (start > src->length() - len)
```
See [`hotspot/src/share/vm/prims/jni.cpp:3836`](http://hg.openjdk.java.net/jdk8u/jdk8u/hotspot/file/9ce27f0a4683/src/share/vm/prims/jni.cpp#l3836) and [`hotspot/src/share/vm/prims/jni.cpp:3885`](http://hg.openjdk.java.net/jdk8u/jdk8u/hotspot/file/9ce27f0a4683/src/share/vm/prims/jni.cpp#l3885).

koutheir:
If nobody is going to review this pull request, then I will remove it entirely. This repository seems dead.

DarkEld3r:
@koutheir Thank you for the pull request and sorry for the delay with response. I'm going to look at this thoroughly tomorrow.

DarkEld3r:
OK, sorry again for the late response. Motivations seem adequate, but I'm not really a Java-guy. Perhaps @dmitry-timofeev or @jrobsonchase will give a more detailed feedback.

Could you please add some tests that will show the flaws of the previous implementation?

koutheir:
**[PLEASE DO NOT MERGE YET]**

koutheir:
From a correctness perspective, the previous implementation works because it relies on the JNI implementation to check for parameters correctness and to throw an `ArrayIndexOutOfBoundsException` in case of invalid parameters.

The idea of this modification is to avoid calling the JNI in the first place when the library knows for a fact that the parameters it passes are invalid. JNI calls are an order of magnitude more expensive than normal function calls, because of JVM state management on entry and on exit of JNI implementations, and because throwing a Java exception is a heavy task. Therefore, this is an effciency modification, which relies on the JNI specifications instead of the JNI implementation.

koutheir:
> Could you please add some tests that will show the flaws of the previous implementation?

Here follows a simple case that would cause failure.
Consider this **Java** boolean array:
```java
boolean[] java_array = new boolean[8];
```
Now consider this **Rust** `jboolean` array:
```rust
let mut rust_buffer: [jboolean; 9] = [0; 9];
```
The following call would **always** result in a **Java** `ArrayIndexOutOfBoundsException` to be thrown by the JVM:
```rust
get_boolean_array_region(java_array, 0, &mut rust_buffer);
```
because the following condition is `true` in the **JNI implementation**:
```c++
(start > src->length() - len)
```
where `start = 0`, `len = 9` and `src->length() = 8`.
This happens because the previous implementation passes `len = rust_buffer.len()` and `src = java_array`.

koutheir:
Actually, now that I think of it, this modification is incomplete, because the caller of `get_boolean_array_region()` needs to know how many elements have been actually transferred into `buf`. I will need to make the changes before this pull request becomes viable again.

koutheir:
Given the approach errors are reported, and the fact that some errors require Java exceptions to be generated (such as `errors::JavaException`), I think my modification will require a lot of redesign in the library. I am cancelling this pull request for the moment.

dmitry-timofeev:
Hi, thanks for the PR! It seems the issue is not correct, as these methods _do_ support copying an arbitrary range of Java array into a Rust slice (or sub-slice, using the range syntax: `&mut dst[start..end]`). They just expect the caller to size that slice correctly. Sorry about the confusion, perhaps, the documentation of these method needs to be improved to avoid such confusion in the future.

Regarding the efficiency: I'd consider a situation where an incorrectly sized slice or a wrong offset is passed as a programming error that needs to be fixed. It must not appear in a correct application. Therefore, it is fine if the precondition violation causes an exception because it won't happen in a correct application (exception is converted in `Err` by this library). An explicit check here would duplicate the implicit check made by the JVM in GetXArrayRegion.

dmitry-timofeev:
Submitted #169 to improve the docs.

PR #157, opened 2018-12-20
throw_new should call ThrowNew

### LINKED ISSUE #156: throw_new calls Throw instead of ThrowNew
`throw_new` (Rust)  is currently implemented in a way that it instantiates a throwable and calls `throw` (Rust) new which calls `Throw` (JNI)
I would expect `throw_new` (Rust) to call `ThrowNew` (JNI) which also avoids the need to create a new throwable.

marschall:
Fixes #156

## Overview

Change throw_new to call ThrowNew directly rather can calling throw.

### Definition of Done

- [X] Change is covered by automated [tests](https://github.com/jni-rs/jni-rs/blob/master/CONTRIBUTING.md#tests)
- [X] This change is not breaking **or** mentioned in the Changelog
- [X] The [continuous integration build](https://www.travis-ci.org/jni-rs/jni-rs) passes


tests/jni_api.rs
```diff
@@ -452,6 +452,17 @@ fn auto_local_null() {
...
 }
 
+
+#[test]
+pub fn throw_new() {
+    let env = attach_current_thread();
+
+    env.throw_new("java/lang/RuntimeException", "Test Exception")
+        .unwrap();
+    env.exception_occurred().unwrap();
```
dmitry-timofeev: Shan't it assert the returned throwable is not `null`? Or use a stronger assertion that it is of the requested type ("java/lang/RuntimeException"), and, possibly, has the requested message?

src/wrapper/jnienv.rs
```diff
@@ -248,7 +248,14 @@ impl<'a> JNIEnv<'a> {
         S: Into<JNIString>,
         T: Desc<'a, JClass<'a>>,
     {
-        self.throw((class, msg))
+        let class = class.lookup(self)?;
+        let msg = msg.into();
+        let res: i32 = jni_unchecked!(self.internal, ThrowNew, class.into_inner(), msg.as_ptr());
+        if res < 0 {
```
dmitry-timofeev: Could you please change that to better match the [spec](https://docs.oracle.com/en/java/javase/11/docs/specs/jni/functions.html#thrownew):

> Returns 0 on success; a negative value on failure.

```
if res == 0 → Ok
else → Err(…)
```

marschall: Sure, no problem. I took that from `throw`, should I update `throw` as well? Should I panic on a positive result or also simply return `Err`?

dmitry-timofeev: > should I update throw as well

Would be wonderful!

> Should I panic on a positive result or also simply return Err?

I think it's fine to simply return `Err` as we can expect the JNI interface functions to follow the specification.

PR #163, opened 2019-02-15
Some methods are allowed to return NULL if not error

skletsun:
## Overview

Some of JNI functions can return NULL as a possible value but current implementation treats this as an error. This PR allows such methods to return a NULL object.

### Definition of Done

### DIFF
--- a/src/wrapper/jnienv.rs
+++ b/src/wrapper/jnienv.rs
@@ -76,15 +76,17 @@ use JavaVM;
 /// will _not_ clear the exception - it's up to the caller to decide whether to
 /// do so or to let it continue being thrown.
 ///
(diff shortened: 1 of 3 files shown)


src/wrapper/jnienv.rs
```diff
@@ -170,7 +170,7 @@ impl<'a> JNIEnv<'a> {
         T: Desc<'a, JClass<'a>>,
     {
         let class = class.lookup(self)?;
-        Ok(jni_non_null_call!(self.internal, GetSuperclass, class.into_inner()))
+        Ok(jni_non_void_call!(self.internal, GetSuperclass, class.into_inner()).into())
```
dmitry-timofeev: Documentation for this method does not mention that/when `null` is possible.

PR #165, opened 2019-03-25
Allow JNIEnv lifetimes to be more flexible

jvff:
## Overview

In some cases, it was only possible to call certain methods if the
lifetime of the JNIEnv borrow was the same of the JNIEnv type (i.e., in
`&'a JNIEnv<'b>`, `'a` had to be equal to `'b`). This PR changes this so
that `'a: 'b` (i.e., the type lives at least as much as the borrow).

This lead to some changes to certain wrapper types (`JList`, `JMap` and
`JavaStr`), which now require both lifetimes to be explicit.

Another requirement was that some implicit lifetimes had to become
explicit, otherwise they would be infered to be the lifetime of `&self`,
i.e., the borrow, not the type.

### Definition of Done

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Changed
+- `JString`, `JMap` and `JavaStr` and their respective iterators now require an extra lifetime so
+  that they can now work with `&'b JNIEnv<'a>`, where `'a: 'b`.
+
 ## [0.11.0]
 
 ### Highlights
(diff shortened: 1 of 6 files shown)


DarkEld3r:
Could you please add a test demonstrating this new flexibility?

jvff:
I added two tests. I apologize for making them more complicated than the tests that already exist, but I didn't find a simpler way to show the separate lifetimes.

If the tests are executing on the current `master` branch (with the change to remove the extra `JList` lifetime), compilation will fail saying that the lifetimes must match. With the fixes in this PR, that is no longer necessary and the tests compile normally.

dmitry-timofeev:
@jvff Thanks again for the PR, the new release is on the crates.io: https://github.com/jni-rs/jni-rs/releases/tag/v0.12.0

jvff:
Awesome, thanks!

PR #170, opened 2019-05-01
Use repr(transparent) and not repr(C) for transparent wrappers

thomcc:
## Overview

`#[repr(C)] struct Foo(T)` isn't guaranteed to behave identically to `T`, but `#[repr(transparent)]` is. In practice, these two are often different, although for pointers (which is how they're used in `jni-rs`) they're _typically_ the same.

This changes the uses of `#[repr(C)]` where identical behavior is desired/required to use `#[repr(transparent)]`.

In practice this _I think_ this won't matter for most modern systems, since these are wrappers around pointers (it might matter on arm64 if you have a method with many arguments), and they're only present as function arguments. However, this is the reason `#[repr(transparent)]` exists, and it seems like tempting fate not to use it.

This should be a non-breaking change.

### Definition of Done

    - Covered by existing tests, not practical to add new tests specifically for this since platforms where it matters are not common.
    - Not a public API
    - non-breaking

### DIFF
--- a/src/wrapper/jnienv.rs
+++ b/src/wrapper/jnienv.rs
@@ -112,7 +112,7 @@ use JavaVM;
 /// Calling unchecked methods with invalid arguments and/or invalid class and
 /// method descriptors may lead to segmentation fault.
 #[derive(Clone)]
-#[repr(C)]
+#[repr(transparent)]
 pub struct JNIEnv<'a> {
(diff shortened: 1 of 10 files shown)


thomcc:
Nice catch, the only reason is that apparently I don't know how to save all the changed files in my text editor before submitting a PR 😅. (fixed)

PR #172, opened 2019-05-06
Adds support for setting a global class loader

bkchr:
New functions:
- `register_class_loader` to register a global class loader
- `unregister_class_loader` to unregister a global class loader

The global class loader will be used by `find_class`, instead of
`JNIEnv::FindClass`.

## Overview

Please describe your changes here and list any open questions you might have.

### Definition of Done

### DIFF
--- a/src/wrapper/class_loader.rs
+++ b/src/wrapper/class_loader.rs
@@ -1,0 +1,59 @@
+use std::sync::Mutex;
+
+use errors::*;
+
+use JNIEnv;
+
+use objects::{GlobalRef, JObject, JClass, JValue};
+
+use strings::JNIString;
+
+use lazy_static::lazy_static;
+
+/// The `loadClass` function name.
+const LOAD_CLASS: &str = "loadClass";
+/// The `loadClass` signature.
+const LOAD_CLASS_SIG: &str = "(Ljava/lang/String;)Ljava/lang/Class;";
+
+lazy_static! {
+    /// The global class loader instance.
+    static ref CLASS_LOADER: Mutex<Option<GlobalRef>> = Mutex::default();
+}
+
+/// Register the given object as the global class loader instance.
+pub fn register_class_loader<'a>(env: &JNIEnv<'a>, class_loader: JObject<'a>) -> Result<()> {
+    // Check that the `loadClass` function is present.
+    env.get_method_id(class_loader, LOAD_CLASS, LOAD_CLASS_SIG)?;
+
+    *CLASS_LOADER.lock().unwrap() = Some(env.new_global_ref(class_loader)?);
+
+    Ok(())
+}
+
+/// Unregister the global class loader instance.
+pub fn unregister_class_loader() {
+    *CLASS_LOADER.lock().unwrap() = None;
+}
+
+/// Look up a class by name.
+///
+/// Either it uses the registered `CLASS_LOADER` or it falls back to use the
+/// JNI env function `FindClass`.
+pub(crate) fn load_class<'a>(env: &JNIEnv<'a>, name: JNIString) -> Result<JClass<'a>> {
+    match *CLASS_LOADER.lock().unwrap() {
+        Some(ref class_loader) => {
+            let name = env.new_string(name)?;
+            let res = env.call_method(
+                class_loader.as_obj(),
+                LOAD_CLASS,
+                LOAD_CLASS_SIG,
+                &[JValue::Object(name.into())]
+            )?;
+            res.l().map(Into::into)
+        },
+        None => {
+            let class = jni_non_null_call!(env.get_native_interface(), FindClass, name.as_ptr());
+            Ok(class)
+        }
+    }
+}
--- a/src/wrapper/jnienv.rs
+++ b/src/wrapper/jnienv.rs
@@ -60,6 +60,8 @@ use signature::JavaType;
 use signature::Primitive;
 use signature::TypeSignature;
 
+use class_loader;
+
 use JNIVersion;
 use JavaVM;
 
@@ -78,11 +80,11 @@ use JavaVM;
 ///
 /// ## `null` Java references
 /// `null` Java references are handled by the following rules:
-///   - If a `null` Java reference is passed to a method that expects a non-`null` 
+///   - If a `null` Java reference is passed to a method that expects a non-`null`
 ///   argument, an `Err` result with the kind `NullPtr` is returned.
 ///   - If a JNI function returns `null` to indicate an error (e.g. `new_int_array`),
-///     it is converted to `Err`/`NullPtr` or, where possible, to a more applicable 
-///     error type, such as `MethodNotFound`. If the JNI function also throws 
+///     it is converted to `Err`/`NullPtr` or, where possible, to a more applicable
+///     error type, such as `MethodNotFound`. If the JNI function also throws
 ///     an exception, the `JavaException` error kind will be preferred.
 ///   - If a JNI function may return `null` Java reference as one of possible reference
 ///     values (e.g., `get_object_array_element` or `get_field_unchecked`),
@@ -165,8 +167,7 @@ impl<'a> JNIEnv<'a> {
         S: Into<JNIString>,
     {
         let name = name.into();
-        let class = jni_non_null_call!(self.internal, FindClass, name.as_ptr());
-        Ok(class)
+        class_loader::load_class(&self, name)
     }
 
     /// Returns the superclass for a particular class OR `JObject::null()` for `java.lang.Object` or
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,7 @@ cesu8 = "1.1.0"
 combine = "3.4.1"
 jni-sys = "0.3.0"
 log = "0.4.4"
+lazy_static = "1"
 
 [build-dependencies]
 walkdir = "2"
@@ -28,9 +29,6 @@ walkdir = "2"
 [dependencies.error-chain]
 default-features = false
 version = "0.12.0"
-
-[dev-dependencies]
-lazy_static = "1"
 
 
 [features]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -228,6 +228,8 @@ extern crate combine;
 
 extern crate cesu8;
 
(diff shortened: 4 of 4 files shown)


bkchr:
@dmitry-timofeev 
1. Yeah every time you call `find_class`, it will use the global class loader if present.
2. Hmm, I'm not sure about surprising effects. It is also only used when the user sets a global class loader, otherwise the default behavior is always to use `FindClass`.
3. I require this on Android. I have multipe threads and they need to be able to call into the Kotlin application. When using the default class loader, the threads don't have the correct class paths set.

bkchr:
Ping @dmitry-timofeev

dmitry-timofeev:
OK, there seem to be the following ways to deal with this in user code:
1. Use the classloader explicitly at each call site where a class is needed.
    - Repetitive.
    - Error-prone: if you forget and rely on auto string->class lookup, you get an error.
2. Put together a cache of class references, populate it manually in, say, [JNI_OnLoad](https://docs.oracle.com/en/java/javase/11/docs/specs/jni/invocation.html#jni_onload):
    - More efficient than the lookup at each call site — obtaining the class object becomes effectively free.
    - Android tips suggest that it [receives the classloader](https://developer.android.com/training/articles/perf-jni#faq:-why-didntfindclassfind-my-class) that loaded the library (the right one)
    - Error-prone: can still forget to use.
    - Must be hand-written — not the easiest thing to get right, and not necessarily *needed* if the application doesn't do frequent lookups.
    - See a [similar thing in Exonum Java](https://github.com/exonum/exonum-java-binding/blob/159fb887aacb50b7291a1c97131a972efe50a0a9/exonum-java-binding/core/rust/src/utils/jni_cache.rs#L50), made for efficiency — but please mind that it relies on some app-specific assumptions (unload never happening) — hence, not the best example for _any_ application.

As both things are somewhat error-prone and require some extra coding, I think we can support overriding the FindClass behaviour *in* the library, if  it does not bring any significant overhead, is documented and tested. @jni-rs/developers , are there any other consideration re this feature?

> … on Android.

I wonder if setting the global classloader might require some cooperation on the user side. If the classloader that loaded the library is garbage-collected (may it happen on Android?) and `JNI_OnUnload` is called — then does the global reference that is kept by the library need to be released or would JVM do that for it? In other words, is the JVM required to keep track of which native library created which global reference so that it can automatically clear the ones that were created in a particular native library?

bkchr:
Ping

dmitry-timofeev:
Hi @bkchr , I think I'll be able to get to the code tomorrow. Have you considered the questions above re unloading process? Is there a definite answer?

bkchr:
How should the class loader be garbage-collected, when we store an instance in a global reference?

dmitry-timofeev:
> If the classloader **that loaded the library** is garbage-collected [...] and JNI_OnUnload is called

I might have been unclear, but I assumed that the one that loaded the library != the global, because otherwise find_class would use that one.

src/wrapper/class_loader.rs
```diff
@@ -0,0 +1,59 @@
...
+const LOAD_CLASS: &str = "loadClass";
+/// The `loadClass` signature.
+const LOAD_CLASS_SIG: &str = "(Ljava/lang/String;)Ljava/lang/Class;";
+
+lazy_static! {
+    /// The global class loader instance.
+    static ref CLASS_LOADER: Mutex<Option<GlobalRef>> = Mutex::default();
+}
+
+/// Register the given object as the global class loader instance.
```
dmitry-timofeev: I would also describe the effects of this method on any subsequent `find_class` invocations.

src/wrapper/class_loader.rs
```diff
@@ -0,0 +1,59 @@
...
+pub fn register_class_loader<'a>(env: &JNIEnv<'a>, class_loader: JObject<'a>) -> Result<()> {
+    // Check that the `loadClass` function is present.
+    env.get_method_id(class_loader, LOAD_CLASS, LOAD_CLASS_SIG)?;
```
dmitry-timofeev: I'd suggest a little clearer way: `env.is_instance_of`.

However, it would also require a not null check :thinking:

src/wrapper/class_loader.rs
```diff
@@ -0,0 +1,59 @@
...
+pub fn unregister_class_loader() {
+    *CLASS_LOADER.lock().unwrap() = None;
+}
+
+/// Look up a class by name.
+///
+/// Either it uses the registered `CLASS_LOADER` or it falls back to use the
+/// JNI env function `FindClass`.
+pub(crate) fn load_class<'a>(env: &JNIEnv<'a>, name: JNIString) -> Result<JClass<'a>> {
+    match *CLASS_LOADER.lock().unwrap() {
```
dmitry-timofeev: This code serializes all class lookups. I think we must ensure that normal `find_class` clients do not pay any price for this feature and are not serialized.

Possibly, one way to achieve that is to acquire the lock only to read the optional because there is no requirement of a mutual exclusion between the class lookups and `register_class_loader`/`unregister_class_loader` (i.e., it is totally OK to let "the last" lookup to proceed). GlobalRefs can be cloned and safely accessed from multiple threads — the last one dropping it will free it.

src/wrapper/class_loader.rs
```diff
@@ -0,0 +1,59 @@
...
+        Some(ref class_loader) => {
+            let name = env.new_string(name)?;
+            let res = env.call_method(
+                class_loader.as_obj(),
+                LOAD_CLASS,
```
dmitry-timofeev: As this is library code, I think it *must* cache the method id of `Classloader#loadClass`
and use `call_method_unchecked` for adequate performance. Otherwise each find_class (that may
be implicit) will also entail (implicit here) `get_object_class` + `get_method_id`.

FYI, there is already an order of magnitude difference between calls that use string identifiers and implicit lookups (absolutely not a problem if you do that once) and the code that caches method ids:
 - https://docs.rs/jni/0.12.3/jni/struct.JNIEnv.html#checked-and-unchecked-methods
 - https://github.com/jni-rs/jni-rs/blob/master/CONTRIBUTING.md#run-benchmarks

src/wrapper/jnienv.rs
```diff
@@ -165,8 +167,7 @@ impl<'a> JNIEnv<'a> {
         S: Into<JNIString>,
```
dmitry-timofeev: I think this documentation must mention that the default behaviour of using FindClass can be overridden and reference the method that allows to do that.

dmitry-timofeev:
> I suspect it does not work at the moment, because loadClass and FindClass use different formats of name argument.

@bkchr , 

Assuming you have tested this change and it worked with binary class names ("com.example.Foo") — have you tried if the default "find_class" works in your case with class names "com/example/Foo", as required by its specification? Maybe the classpath is actually OK?

bkchr:
I just use the com/example/Foo class names and that works.

src/wrapper/class_loader.rs
skletsun: I'm wouldn't it be more optimal to use `RwLock` instead of `Mutex` here? 🤔

dmitry-timofeev: To read a couple of words, [unlikely](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/locks/ReadWriteLock.html) — but the only way to definitely tell is to measure. The problem is coming with a _realistic_ scenario — the worst-case one we can easily put together.

dmitry-timofeev:
> I just use the com/example/Foo class names and that works.

Good, does it work with the default implementation of `find_class`? Or do you still need to use the custom classloader?

rib:
As I recently started helping with maintaining the project I've been going through some of the older PRs and issues and this one stood out since I'm also working with Android and think I've come across the same kind of issue that probably lead to this PR.

Sorry if you don't really recall much of the context for this old issue @bkchr but I figured I'd follow up and then I can hopefully determine whether it make sense to pursue the original idea / PR, possibly consider a different solution or perhaps close this PR.

@bkchr do you maybe remember whether there was a realization in the end that the wrong syntax was being used with `find_class` which meant you didn't end up needing this? (regarding this last question: https://github.com/jni-rs/jni-rs/pull/172#issuecomment-499119430)

When I've had this issue before I ended up manually getting the `ClassLoader` from the the Android `Activity` and using `loadClass` directly, instead of `env.find_class()`. This way you could call the regular `env.find_class()` for system classes or else call your own utility for non-system classes. The disadvantage is that you have to know in advance whether you're looking up a system class or not - but I guess that would be ok in many situations (well if you didn't know I suppose you could default to the `loadClass` path and it might just be a little less optimized perhaps).

One of the notable challenges with being able to hook into `env.find_class()` is that there isn't really a mechanism for `jni-rs` to specifically attach private state to a `JNIEnv` or `JavaVM`- which is something that also constrains the current ergonomics of APIs like `JList` and `JMap` that can't currently cache method IDs to re-use.

The patch here ends up introducing a global static as a workaround for that which is notably quite awkward since the lifetime of that static is not related to the current `JavaVM`. This is probably what's most contentious about this current PR.

If it was instead somehow just easier to find a `ClassLoader` from an `Activity` pointer on Android (e.g. got from the [`ndk-context` crate](https://crates.io/crates/ndk-context) and then easily call `classLoad` to find a class, maybe that could also solve the original problem without needing to hook into `env.find_class()`?

bkchr:
@rib sadly I really don't remember the details :( I think it was something around loading a class in a thread, but more isn't left in my brain.

rib:
yeah, no worries @bkchr - sorry that the issue didn't get resolved in this time.

I think I'll probably end up closing this PR while I guess the momentum behind it got lost some time ago.

I can imagine there could be various things that could be done to try and help access the right classLoader on Android more easily (e.g. for standalone Rust applications) and I had had some early thoughts about ways of helping via the `android-activity` crate in some cases, and/or `ndk-context` crate perhaps.

Without a specific case to focus on for an issue though it's hard to determine for now whether having a mechanism to hook into `find_class()` is the right thing to do - and the need for the static global state is something that's technically a bit contentious to decide about here.

This issue can always be re-opened if it makes sense to in the future.

PR #173, opened 2019-05-08
Pass `jboolean` directly to `JValue` instead of using bool

bkchr:
## Overview

So I had this bug with calling a `hasNext` that always returned `false`. After debugging it for a while, I found that adding `println!` fixed the error:
```
let res = jni_non_void_call!(
  self.internal,
  CallBooleanMethodA,
  obj,
  method_id,
  jni_args
);

println!("HELLO {}", res);

(res == sys::JNI_TRUE).into()
```

This pr solves this by forwarding the `jboolean` to `JValue` directly, instead of converting it first into `bool` and then into `jboolean` again. The compiler probably generated some incorrect code or the `bool as jboolean` failed, I'm not quite sure. However, it now works for me and make the code cleaner.

### Definition of Done

### DIFF
--- a/src/wrapper/objects/jvalue.rs
+++ b/src/wrapper/objects/jvalue.rs
@@ -99,7 +99,7 @@ impl<'a> JValue<'a> {
     /// Try to unwrap to a boolean.
     pub fn z(self) -> Result<bool> {
         match self {
-            JValue::Bool(b) => Ok(b != 0),
+            JValue::Bool(b) => Ok(b == JNI_TRUE),
             _ => Err(ErrorKind::WrongJValueType("bool", self.type_name()).into()),
         }
     }
@@ -175,10 +175,16 @@ impl<'a> From<JObject<'a>> for JValue<'a> {
     }
 }
 
-// jbool
 impl<'a> From<bool> for JValue<'a> {
     fn from(other: bool) -> Self {
-        JValue::Bool(other as jboolean)
+        JValue::Bool(if other { JNI_TRUE } else { JNI_FALSE })
+    }
+}
+
+// jbool
+impl<'a> From<jboolean> for JValue<'a> {
+    fn from(other: jboolean) -> Self {
+        JValue::Bool(other)
     }
 }
 
--- a/tests/jmap.rs
+++ b/tests/jmap.rs
@@ -19,7 +19,8 @@ pub fn jmap_push_and_iterate() {
     let env = attach_current_thread();
     let data = &["hello", "world", "from", "test"];
 
-    let map_object = unwrap(&env, env.new_object("java/util/HashMap", "()V", &[]));
+    // Create a new map. Use LinkedHashMap to have predictable iteration order
+    let map_object = unwrap(&env, env.new_object("java/util/LinkedHashMap", "()V", &[]));
     let map = unwrap(&env, JMap::from_env(&env, map_object));
 
     // Push all strings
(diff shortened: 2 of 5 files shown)


src/wrapper/objects/jvalue.rs
```diff
@@ -175,10 +175,16 @@ impl<'a> From<JObject<'a>> for JValue<'a> {
     }
 }
 
-// jbool
 impl<'a> From<bool> for JValue<'a> {
     fn from(other: bool) -> Self {
-        JValue::Bool(other as jboolean)
+        JValue::Bool(if other { JNI_TRUE } else { JNI_FALSE })
```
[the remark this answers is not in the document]
bkchr: I think this is needed. Otherwise the user needs to add this `if other { JNI_TRUE } else { JNI_FALSE }` in their code, when they want to create a `JValue` from a `bool`. The other types are okay, as you can use them directly, but for `bool` you mostly use `bool` and don't want to use `jboolean`.

bkchr:
This was in my Android application using `JMapIterator`. Are there any tests that are executed with a JVM? But I'm not sure if that can be reproduced in a normal test.

bkchr:
Okay, I added a test, but it worked without my fixes as well :( So, maybe some weird bug in relationship with Android.

tests/jmap.rs
```diff
@@ -0,0 +1,50 @@
...
+    attach_current_thread,
+    unwrap,
+};
+
+#[test]
+pub fn jmap_push_and_iterate() {
+    let env = attach_current_thread();
+    let data = &["hello", "world", "from", "test"];
+
+    let map_object = unwrap(&env, env.new_object("java/util/HashMap", "()V", &[]));
```
dmitry-timofeev: If the test requires a particular order of keys to simplify comparison, I'd suggest either:
1. Using [`java/util/TreeMap`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/TreeMap.html) — it keeps keys ordered.
2. Or, better, [`LinkedHashMap`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/LinkedHashMap.html) — it preserves the insertion order, hence neither the source nor the `collected` need to be sorted below.

tests/jmap.rs
```diff
@@ -0,0 +1,50 @@
...
+        }),
+    );
+
+    let mut collected = Vec::new();
+    unwrap(
+        &env,
+        map.iter().and_then(|mut iter| {
+            iter.try_for_each(|e| {
+                env.get_string(e.0.into())
+                    .map(|s| collected.push(String::from(s)))
```
dmitry-timofeev: \+ @vitvakatu @DarkEld3r

Is it idiomatic in Rust? In Java streams, for instance, it is encouraged to use collectors instead of manually creating and updating a destination collection: 

```java
List<Result> dst = src.stream()
    .map(Foo::transform)
    .collect(Collectors.toList());
```

bkchr: I did not used `collect` here, because we deal with a lot of `Result`s here. And to not write `unwrap(&env, ...)` every time, I chain all the results.

tests/jmap.rs
bkchr: The test does not require any particular order. I just tried to replicate the error I had seen on Android and this was with a `HashMap`.

bkchr:
@dmitry-timofeev anymore questions?

PR #176, opened 2019-05-22
Release 0.12.3

dmitry-timofeev:


dmitry-timofeev:
Yes, I think we can delay the release a bit to include it. Also, @skletsun might have planned to include some additional fields in the output 🤷‍♂

Issue #179, opened 2019-05-24
Detach threads automatically on exit


dmitry-timofeev:
Detach attached through Invocation API threads automatically on their exit.

In the present version:
 - The daemon threads cannot be detached (see #128), potentially, leaking VM memory; 
 - Regular threads are detached as soon as the `AttachGuard` returned by `JavaVM#attach_current_thread` is dropped, which, albeit a leak-safe option, might not be the desired behaviour if the same native thread needs to repeatedly and frequently call into Java, because of high overhead of attach/detach operation;
- There is no reliable **and** efficient way of managing attachment of 'foreign' native threads — the one your app does not control.

The goal of this issue is to provide a mechanism to attach the native threads 'permanently' — till they exit; and automatically detach them on exit. That can be achieved utilising thread-local storage, by placing there a guard object similar to `AttachGuard`.

There is a recommendation to use such approach with examples using pthreads in [Android tips](https://developer.android.com/training/articles/perf-jni#threads_1) and on [SO](https://stackoverflow.com/a/30026231/1792188).

There are some questions on the API:
1. What is the default for `attach_current_thread`: attach locally, or permanently?
2. Shall we support both modes for `attach…_as_daemon`, or only permanently?
3. Shall we also take care of the local frame (as a user failure to manage locals properly is likely to be more severe than with briefly attached threads), for instance, with such interface: 
```rust
vm->with_attached_as_daemon(|env| -> { 
  /* Use env, the local frame has already been allocated */
});
// That, however, still leaves some potential for misuse:
let env = vm->get_env()?;
// Yay, let's leak some locals, forgetting of frames/auto_locals
while true {
  env.new_object()?;
}
```
or leave it for an abstraction atop JavaVM (e.g., an `Executor` as in #83 )?

vitvakatu:
Implementation in #180 answers the above questions the following way:
1. `attach_current_thread` preserves old behaviour, thread is attached until returned `AttachGuard` dropped.
2. `attach_current_thread_as_daemon` attaches the thread permanently, no other option provided
3. Local frames handling is not implemented

The following questions arose: 
1. Should we detach the thread in `JavaVM::new`? Current implementation does it, though original JNI implementation doesn't.
[...]

PR #180, opened 2019-05-30
Implement TLS-based AttachGuard

vitvakatu:
## Overview

See #179 

Questions:
1. Is there a better name than InternalAttachGuard?
2. What is the default for attach_current_thread: attach locally, or permanently? In this implementation, current behavior is preserved, `attach_current_thread_permanently` added.
3. Shall we support both modes for attach…_as_daemon, or only permanently? In this implementation, `attach_current_thread_as_daemon` supports only permanent attach.
4. Shall we also take care of the local frame (as a user failure to manage locals properly is likely to be more severe than with briefly attached threads). In this implementation, no actions were made by intent.
5. Tests organization is not nice. We can solve it by running _all_ tests in a single thread, but it will need environment variable/special cargo flag on each `cargo test` run.
6. What other test cases can be added?
7. Should we detach thread in JavaVM::new()?

### Definition of Done


vitvakatu:
Additional question: should we handle thread names (output them in logs)?

vitvakatu:
🤔 Commit history is awful, going to force push

tests/threads_detach.rs
```diff
@@ -0,0 +1,59 @@
...
+use jni::{
+    objects::JValue,
+    sys::jint,
+};
+
+mod util;
+use util::jvm;
+use std::time::Duration;
+
+// We forced to combine several tests in one function, because every test function is running in
```
dmitry-timofeev: Does cargo support 'single threaded' mode for a single test file via an annotation?

If it doesn't, I'd consider either splitting into indy files (that would hurt navigation/readability, but provide perfect isolation and better reporting as individual tests), or moving them all in a single file in separate test methods. Currently a single test is in vm.rs, three — in here, which makes little sense.

vitvakatu:
> InternalAttachGuard::create_and_attach seems to be the main function that is suppossed to attach/detach threads. Why then VM::attach_current_thread uses InternalAttachGuard::attach_current_thread directly, then creates an AttachGuard, that, in turn, creates a new InternalAttachGuard?

Agreed, I did some refactoring and now it looks better I think. The initial reason was attempt to keep different behavior of `attach_current_thread` and `attach_current_thread_permanently` without code duplication.

>Also, why does InternalAttachGuard include 'shall_detach' flag? Shan't it always detach the attached thread? That flag exists so that attach_current_thread can return a non-detaching AttachGuard if called twice. It seems sensible to keep that logic inside AttachGuard, and not propagate it to InternalAttachGuard, unless I'm missing something.

Detaching the thread is clearly the role of `InternalAttachGuard`, and its `detach` is called from `Drop` implementation. Without check for `should_detach`, we will get an invalid behavior for nested attaches. However, we cannot send `should_detach` to `Drop` implementation without storing it directly inside `InternalAttachGuard`. The other option is to have different `Drop` for `AttachGuard` and reimplement it without even using `InternalAttachGuard`, but this will lead us to unnecessary code duplication, so I don't think we can do smth here.

> Also, does it make sense to move the common pattern "if has env — return, else — attach" inside the guards? Would it clarify the code and reduce the duplication?

Some code duplication, but I think that it separates roles clearly, doesn't it?

tests/threads_detach.rs
vitvakatu: No support afaik, so I split these tests into several files, yes.

tests/vm.rs
```diff
@@ -26,7 +26,7 @@ pub fn nested_attach_guard_should_not_detach_thread() {
         assert_eq!(val, 2);
```
dmitry-timofeev: I'd check that the number of attached threads remains the same as before the second call (1?).

dmitry-timofeev:
I think the code is still too complicated and we must ponder on a simpler design. Would it be clearer if there is a *single* internal function that performs attachment (i.e., `IAG::create_and_attach`) and leaves the IAG inside TLS, and an AttachGuard that may or may not clear the TLS in its `Drop` implementation depending on the 'shall_detach' flag, but does not contain the `IAG` instance?

```whatever
VM::attach_locally:
  if attached:
    return AttachGuard::new_nested(env)
  else:
    env = IAG::create_and_attach(REGULAR)
    return AttachGuard::new(env)

VM::attach_permanently:
  if attached:
    return env
  else:
    return IAG::create_and_attach(REGULAR)

VM::attach_as_daemon (implies permanently)
  if attached:
    return env
  else:
    return IAG::create_and_attach(DAEMON)

VM::detach_current_thread: clears the TLS (might be delegated to an internal function).

Drop for AttachGuard: if not nested → clears the TLS (might be delegated to an internal function).

Drop for IAG: performs the actual detachment.
```

dmitry-timofeev:
>    Is there a better name than InternalAttachGuard?

I think it's good + it is an internal structure.

>    What is the default for attach_current_thread: attach locally, or permanently? In this implementation, current behavior is preserved, attach_current_thread_permanently added.

:+1: 

>    Shall we support both modes for attach…_as_daemon, or only permanently? In this implementation, attach_current_thread_as_daemon supports only permanent attach.

I think not (as suggested in the issue).

>    Shall we also take care of the local frame (as a user failure to manage locals properly is likely to be more severe than with briefly attached threads). In this implementation, no actions were made by intent.

I'd discuss that separately from this PR in the issue.

> Tests organization is not nice. We can solve it by running all tests in a single thread, but it will need environment variable/special cargo flag on each cargo test run.

See above — maybe a separate module/directory or naming pattern would help.

> What other test cases can be added?

- Nested attaches in various combinations not having effect on the outer guard?
  - [x] attach_local -> attach_local
  - attach_local -> attach_permanently. This one is questionable — if I have an already attached threads, shall attach_permanently do what it says, or have a clause similar to the one in 'attach_as_daemon' saying that it does not change the status of the thread?
  - attach_local -> attach_as_daemon.
  - attach_permanently -> attach_local
  - attach_as_daemon -> attach_local
  - …
- Explicit detach with variously attached threads? Also is it dangerous to use if the user has `attach_current_thread` with a Guard :thinking: ? Probably, yes, as the JNIEnv cannot be used after the thread has been detached.
- All detaches happening.

> Should we detach thread in JavaVM::new()?

:thinking: :man_shrugging: :hourglass:

src/wrapper/java_vm/vm.rs
```diff
@@ -131,7 +178,102 @@ impl<'a> Deref for AttachGuard<'a> {
...
+            AttachCurrentThread,
+            &mut env_ptr,
+            ptr::null_mut()
+        );
+        jni_error_code_to_result(res)?;
+
+        ATTACHED_THREADS.fetch_add(1, Ordering::SeqCst);
+
+        debug!("Attached thread {:?}. {} threads attached",
+               current().id(),
```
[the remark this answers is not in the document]
skletsun: Nit: I feel that `ThreadId` with the `{:?}` formatter will be printed like this:
```
ThreadId(
    42
)
```
Perhaps it's worth converting it to `int` an use `{}`

src/wrapper/java_vm/vm.rs
```diff
@@ -132,9 +190,128 @@ impl<'a> Deref for AttachGuard<'a> {
...
+            self.java_vm,
+            AttachCurrentThread,
+            &mut env_ptr,
+            ptr::null_mut()
+        );
+        jni_error_code_to_result(res)?;
+
+        ATTACHED_THREADS.fetch_add(1, Ordering::SeqCst);
+
+        debug!("Attached thread {:?}. Name: {:?}. {} threads attached",
```
dmitry-timofeev: Nit (here and elsewhere): would the following format be more human-readable:

`Attached thread foo (id=1234).`

?

src/wrapper/java_vm/vm.rs
```diff
@@ -132,9 +190,128 @@ impl<'a> Deref for AttachGuard<'a> {
...
+
+        Ok(env_ptr as *mut sys::JNIEnv)
+    }
+
+    fn detach(&mut self) -> Result<()> {
+        unsafe {
+            java_vm_unchecked!(self.java_vm, DetachCurrentThread);
+        }
+        ATTACHED_THREADS.fetch_sub(1, Ordering::SeqCst);
+        debug!("Detached thread {:?}. Name: {:?}. {} threads attached",
```
dmitry-timofeev: Nit: x threads _remain_ attached.

src/wrapper/java_vm/vm.rs
```diff
@@ -132,9 +190,128 @@ impl<'a> Deref for AttachGuard<'a> {
...
+        );
+
+        Ok(env_ptr as *mut sys::JNIEnv)
+    }
+
+    fn detach(&mut self) -> Result<()> {
+        unsafe {
+            java_vm_unchecked!(self.java_vm, DetachCurrentThread);
+        }
+        ATTACHED_THREADS.fetch_sub(1, Ordering::SeqCst);
```
dmitry-timofeev: Is that correct that it won't go below zero even with a public  `detach_current_thread` because **all** detaches happen through this method?

src/wrapper/java_vm/vm.rs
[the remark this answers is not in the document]
dmitry-timofeev: @vitvakatu does it really take 3 lines?

src/wrapper/java_vm/vm.rs
skletsun: > Nit (here and elsewhere): would the following format be more human-readable:
> 
> `Attached thread foo (id=1234).`
> 
> ?

Thread is not required to have a name, that's why `.name()` returns `Option<>`. I guess we could use "Attached thread ["NAME_" or ""] with id: ID".

vitvakatu: Btw, Debug implementation for ThreadId really sucks - it will be printed as `ThreadId(<number>)`, and it seems like we can't make it nicer

src/wrapper/java_vm/vm.rs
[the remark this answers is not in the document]
vitvakatu: First of all, `ThreadId` does not have conversions to numbers, so we cannot do smth here. Secondly, `{:?}` debug formatter prints `ThreadId` as `ThreadId(<number>)`. `{:#?}` formatter will print it in three lines, probably.

src/wrapper/java_vm/vm.rs
vitvakatu: Yes, all detaches somehow triggers IAG::drop, which is the only entry points for this method

src/wrapper/java_vm/vm.rs
dmitry-timofeev: It's ok — you don't have to put `id=` :-)

src/wrapper/java_vm/vm.rs
```diff
@@ -132,9 +190,129 @@ impl<'a> Deref for AttachGuard<'a> {
...
+
+#[derive(Debug)]
+struct InternalAttachGuard {
+    java_vm: *mut sys::JavaVM,
+}
+
+impl InternalAttachGuard {
+    fn create_and_attach(
+        java_vm: *mut sys::JavaVM,
+        ty: ThreadType,
```
skletsun: Nit: `ty` -> `thread_type`?

src/wrapper/java_vm/vm.rs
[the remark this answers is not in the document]
skletsun: Yeap, I mixed it with `{:#?}` formatter. And I agree that `ThreadId(<number>)` output is OK.

skletsun:
👍

src/wrapper/java_vm/vm.rs
```diff
@@ -49,45 +52,88 @@ impl JavaVM {
...
+    pub fn attach_current_thread_permanently(&self) -> Result<JNIEnv> {
+        match self.get_env() {
+            Ok(env) => Ok(env),
+            Err(_) => {
+                unsafe {
+                    let env_ptr = InternalAttachGuard::create_and_attach(
+                        self.get_java_vm_pointer(),
+                        ThreadType::Normal
+                    )?;
+                    JNIEnv::from_raw(env_ptr)
```
DarkEld3r: Only `from_raw` is unsafe here, right? Lets reduce `unsafe` scope.

src/wrapper/java_vm/vm.rs
```diff
@@ -49,45 +52,88 @@ impl JavaVM {
...
             },
-        };
+            Err(_) => {
+                let env_ptr = InternalAttachGuard::create_and_attach(
+                    self.get_java_vm_pointer(),
+                    ThreadType::Normal
+                )?;
+                let env = unsafe {
+                    JNIEnv::from_raw(env_ptr)?
+                };
```
DarkEld3r: I see a very similar code in `attach_current_thread_permanently`, `attach_current_thread` and `attach_current_thread_as_daemon`. Can it be moved into a function?

src/wrapper/java_vm/vm.rs
```diff
@@ -49,45 +52,88 @@ impl JavaVM {
...
+                    ThreadType::Daemon
+                )?;
+                unsafe { JNIEnv::from_raw(env_ptr) }
+            }
         }
     }
 
+    /// Returns current number of threads attached to the JVM.
+    pub fn threads_attached(&self) -> usize {
+        ATTACHED_THREADS.load(Ordering::SeqCst)
```
DarkEld3r: Just curious: is it only needed for debug purpose or do you build some logic on top of this?

src/wrapper/java_vm/vm.rs
```diff
@@ -132,9 +190,129 @@ impl<'a> Deref for AttachGuard<'a> {
...
+        let mut env_ptr = ptr::null_mut();
+        let res = java_vm_unchecked!(
+            self.java_vm,
+            AttachCurrentThread,
+            &mut env_ptr,
+            ptr::null_mut()
+        );
+        jni_error_code_to_result(res)?;
+
+        ATTACHED_THREADS.fetch_add(1, Ordering::SeqCst);
```
DarkEld3r: Maybe it makes sense to use value returned by `fetch_add`?

src/wrapper/java_vm/vm.rs
dmitry-timofeev: I think for debug/diagnostic purposes + easier testing. I considered to add that to documentation separately.

BTW, we can add a fancier test that ensures that a thread is _indeed_ detached: in a loop, create a thread, attach permanently, allocate a java byte array and leak it (= don't use auto_locals or local_frame), join. Create that many threads so that the total allocated Java heap exceeds the configured limit.

src/wrapper/java_vm/vm.rs
vitvakatu: It returns previous value, which is not much useful for us. We can increment it, but I think it isn't worth the effort.

src/wrapper/java_vm/vm.rs
vitvakatu: Decided to move `create_and_attach` into JavaVM as internal method - it reduces the code duplication significantly, but maybe brings some mess. What do you think?

src/wrapper/java_vm/vm.rs
vitvakatu: Done

PR #185, opened 2019-06-14
Hide `detach_current_thread` docs

vitvakatu:
## Overview

Hides `detach_current_thread` method to prevent its usage except for exceptional situations. See #179 

### Definition of Done

### DIFF
--- a/src/wrapper/java_vm/vm.rs
+++ b/src/wrapper/java_vm/vm.rs
@@ -73,7 +73,11 @@ unsafe impl Send for JavaVM {}
 unsafe impl Sync for JavaVM {}
 
 impl JavaVM {
-    /// Launch a new JavaVM using the provided init args
+    /// Launch a new JavaVM using the provided init args.
+    ///
+    /// Unlike original JNI API, the main thread (the thread from which this method is called) will
+    /// not be attached to JVM. You must explicitly use `attach_current_thread…` methods (refer
+    /// to [Attaching Native Threads section](#attaching-native-threads)).
     #[cfg(feature = "invocation")]
     pub fn new(args: InitArgs) -> Result<Self> {
         use std::os::raw::c_void;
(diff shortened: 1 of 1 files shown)


dmitry-timofeev:
Probably, it is the better middle ground between "no support at all" and "here you go, go detach threads". But people will still see it in their IDEs, therefore, I think a more detailed documentation is needed both at JavaVM level and at this method.

vitvakatu:
@dmitry-timofeev added detailed comment to #179. 🤔 We could not hide part of docs, so `JavaVM`-level docs might be confusing for users from `docs.rs`, where method is invisible.

src/wrapper/java_vm/vm.rs
```diff
@@ -76,8 +76,8 @@ impl JavaVM {
     /// Launch a new JavaVM using the provided init args.
     ///
     /// Unlike original JNI API, the main thread (the thread from which this method is called) will
-    /// not be attached to JVM. You must explicitly use `attach_current_thread...` methods (refer
-    /// to [Attaching Native Threads section](#attaching-native-threads)).
+    /// not be attached to JVM. You must explicitly use `attach_current_thread…` methods (refer
+    /// to [Attaching Native Threads section](#attaching-native-threads).
```
vitvakatu: missing closing brace

src/wrapper/java_vm/vm.rs
```diff
@@ -154,20 +154,22 @@ impl JavaVM {
...
     ///
-    /// Detaching a non-attached thread is no-op.
+    /// Detaching a non-attached thread is a no-op.
     ///
-    /// Calling this method is an equivalent for calling `drop()` for `AttachGuard`. However, the
-    /// actual guard will not be dropped. __Any existing `JNIEnv` and `AttachGuard` will be
-    /// considered invalid in the current thread after calling this method.__ Any operation
-    /// on `JNIEnv` will return an error.
+    /// __Any existing `JNIEnv`s and `AttachGuard`s created in the calling thread
+    /// will be invalidated after this method completes. It is the__ caller’s __responsibility
```
vitvakatu: I wonder if this will render correctly, it seems like underscores on this line are redundant

vitvakatu: Oh well, it will not render correctly, because it will not render at all :D
We hide this method, so it seems we should not use markdown formatting here, as it is useless

dmitry-timofeev: They are not redundant — they highlight caller’s from the rest of the text, which is in bold. You will see highlighting if you use "Show quick documentation" in your IDEA (E.g., Ctrl + Q).

vitvakatu: Okay, I returned them back

dmitry-timofeev:
Merged CI -> master -> auto-detach-threads, going to rebase this branch onto the most recent auto-detach-threads.

dmitry-timofeev:
Moved VM example out as it breaks the doc tests currently

Issue #83, opened 2018-03-01
Consider Executors providing JNIEnv


dmitry-timofeev:
## Overview
This is a _feature request_ for an abstraction enabling any code to execute a lambda in some context with JNIEnv (i.e., in a native thread that is guaranteed to be attached to the JVM).

Such abstraction is useful when impelementing native proxies of Java objects:

```rust
trait Executor {
  // Executes the given function in a native thread 
  // that is attached to the JVM (= has JNIEnv).
  def <T> with_context(Function<&JNIEnv, T> function);
}

// A native proxy of a Java Foo class.
struct JavaFooProxy {
  // A global reference to an instance of the Java Foo class.
  foo: GlobalRef;

  // An executor, allowing to call a Java method in an arbitrary moment.
  executor: Executor;
  
  // Calls Java `Foo#bar` method.
  def bar() {
    executor.with_context((env: &JNIEnv) -> {
        env.call_method(foo, "bar", BAR_SIGNATURE, …);
    }
  }
}
```

## Possible implementations
How `Executor` provides a context (JNIEnv) — is an implementation detail. 
There might be various implementations of such an interface:
1. SnailExecutor: always attaches/detaches the current thread, which is VERY costly :snail: 
2. CurrentThreadExecutor: will work when the current thread is already attached, uses JNIEnv directly.
3. HackyExecutor: relies on the fact that your app never creates more than N threads 
   and _never_ `detach`es them (a benign leak).
4. DefaultExecutor: keeps a couple of attached threads and passes the lambda to them. 
   Reliable, but requires some time and care to get right.

In our code base, we currently have a `SnailExecutor` only.

## Questions
Certainly not an exaustive list, please post yours:
- [ ] Q0: Is it useful to the target audience? Shall it be:
    - In the core library?
    - In a separate satellite crate, e.g., `jni-native-proxies`?
    - Outside the lib?
- [ ] Q1: Which implementations the library must provide, if any?
- [ ] Q2: (Might be out-of-scope) Can one write a proxy that can work with both `AutoLocal` and `GlobalRef`?

## See also
- Auto-generation of native proxies: #82

PR #186, opened 2019-06-18
Add Executor abstraction

vitvakatu:
## Overview

Executor is a simple thread attachment manager. It allows to execute closures in attached threads with automatic local references managment done with `with_local_frame`.

See also #83 

### Definition of Done


src/wrapper/java_vm/vm.rs
```diff
@@ -36,6 +36,12 @@ use InitArgs;
 /// an appropriately-sized local frame for larger code fragments
 /// (see [`with_local_frame`](struct.JNIEnv.html#method.with_local_frame))
 /// and [auto locals](struct.JNIEnv.html#method.auto_local) in loops.
+///
+/// Jni-rs provides a helper struct to deal with such local references -
```
dmitry-timofeev: I struggle to put in a clear and concise way :-) Basically, it provides JNIEnv to any native thread, combining the performance benefits of permanent attaches whilst removing the risk of local references leaks if used consistently. Is that correct that it is its main value proposition?

tests/executor.rs
```diff
@@ -0,0 +1,73 @@
...
+        assert_eq!(1, atomic.increment_and_get().unwrap());
+        assert_eq!(3, atomic.add_and_get(2).unwrap());
+        atomic
+    });
+    let mut atomic = jh.join().unwrap();
+    assert_eq!(3, atomic.get().unwrap());
+}
+
+fn test_concurrent_threads(executor: Executor, thread_num: usize) {
+    const ITERS_PER_THREAD: usize = 10_000;
```
dmitry-timofeev: How long does this test take? :upside_down_face:

src/wrapper/executor.rs
```diff
@@ -0,0 +1,55 @@
...
+
+/// The capacity of local frames, allocated for attached threads by default. Same as the default
+/// value Hotspot uses when calling native Java methods.
+pub const DEFAULT_LOCAL_FRAME_CAPACITY: i32 = 32;
+
+/// Thread attachment manager. Attaches threads as daemons, hence they do not block
+/// JVM exit. Finished threads detach automatically.
+#[derive(Clone)]
+pub struct Executor {
+    vm: Arc<JavaVM>,
```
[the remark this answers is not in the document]
skletsun: > Is it correct that Executor is not Send + Sync whilst JavaVM — is? Isn't it appropriate to make Executor such?

JavaVM is Send & Sync --> Arc<JavaVM> is Send & Sync --> Executor is Send & Sync.

tests/executor.rs
vitvakatu: About one second, I think it's affordable.

src/wrapper/executor.rs
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    /// Executes a provided closure, making sure that the current thread
+    /// is attached to the JVM. Additionally ensures that local object references are freed after
+    /// call.
+    ///
+    /// Allocates a local frame with the specified capacity.
+    pub fn with_attached_capacity<F, R>(&self, capacity: i32, f: F) -> Result<R>
+    where
+        F: FnOnce(&JNIEnv) -> Result<R>,
```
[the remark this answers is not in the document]
vitvakatu: Yes. `JObject` can't outlive `JNIEnv` it was created with, and `JNIEnv` will be destroyed at closure exit because it's not clonable.

src/wrapper/executor.rs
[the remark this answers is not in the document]
vitvakatu: 1. Executor is Send + Sync, it's visible in docs. @skletsun described it correctly
изображение
2. The main reason of putting `JavaVM` in `Arc` is to make it clonable and avoid dangling pointers. Are there any problems with `Arc` here?
3. It's interesting option. We can do it and it will simplify the code a bit, but probably in a separate PR.

src/wrapper/java_vm/vm.rs
```diff
@@ -42,9 +41,10 @@ use InitArgs;
 /// and [auto locals](struct.JNIEnv.html#method.auto_local) in loops.
 ///
 /// Jni-rs provides a helper struct to deal with such local references -
-/// [`Executor`](struct.Executor.html). It is a simple abstraction that hides the necessity
-/// of handling local references behind the
-/// [`Executor::with_attached`](struct.Executor.html#method.with_attached) method.
+/// [`Executor`](struct.Executor.html). The `Executor` provides `JNIEnv` to any native thread with
+/// reasonably high performance, making use of permanent attaches. It also removes the risk of local
+/// references leaks if used consistently. `Executor` is recommended to use when the scope of
+/// necessary attachment is distinct.
```
dmitry-timofeev: What do you mean by 'distinct' scope?

src/wrapper/executor.rs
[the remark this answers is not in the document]
dmitry-timofeev: I don't know if there are any problems, it just seems somewhat strange to use reference-counting with basically a _pointer_ to a foreign object (= as a pointer it has no Drop implementation, therefore, its atomic destruction, which Arc handles, is not an issue).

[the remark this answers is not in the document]
vitvakatu: It's a safe way to clone pointer. Otherwise you can't guarantee that JVM will not be destroyed before the last use of the pointer.

src/wrapper/java_vm/vm.rs
vitvakatu: When the scope is relatively small and well defined (e.g. we're not using JNIEnv during the whole program execution). Perhaps it's not the best wording here

dmitry-timofeev: I don't think 'distinct' works here. I also don't understand why the scope has to be 'relatively small and well defined' — Exonum uses it in an unbounded scope.

vitvakatu: The scope of attachment is always restricted by closure passed into `with_attached` method, I mean this. Scope is `unbounded` when we attach the main thread and do not detach it to the end of the program execution, this is my view of the definition.

dmitry-timofeev: Well, there is a scope of thread attachment — where the native thread remains attached, and a scope in which the JNI interface pointer is accessible. The first may be unbounded for Executors, the second is limited.

PR #194, opened 2019-07-01
Use unchecked exception by default

### LINKED ISSUE #119: Consider unchecked exception as the default for Desc<JThrowable>
## Overview
Checked `j.l.Exception` is used as the default in `Desc<JThrowable>` implementations. Such exception cannot be thrown unless the Java native method declares it.

`j.l.RuntimeException` might be more appropriate, on the other hand, it might be reasonable to always require the callers to specify the exception type explicitly.

## Questions
1. [ ] Shall we allow to create an exception of a default type (e.g., `j.l.RuntimeException`) or always require to specify its type?

skletsun:
## Overview

This fixes #119 by using unchecked `java.lang.RuntimeException` as the default in `Desc<JThrowable>` implementations .

### Definition of Done

### DIFF
--- a/tests/jni_api.rs
+++ b/tests/jni_api.rs
@@ -4,11 +4,14 @@ extern crate error_chain;
 extern crate jni;
 
 use jni::{
+    descriptors::Desc,
     errors::{Error, ErrorKind},
     objects::{AutoLocal, JByteBuffer, JList, JObject, JValue},
     signature::JavaType,
     sys::{jint, jobject, jsize},
     JNIEnv,
+    objects::JThrowable,
+    strings::JNIString
 };
 use std::str::FromStr;
 
@@ -572,22 +575,11 @@ pub fn throw_new() {
 
     let result = env.throw_new("java/lang/RuntimeException", "Test Exception");
     assert!(result.is_ok());
-    let ex: JObject = env
-        .exception_occurred()
-        .expect("Exception should be thrown")
-        .into();
-    assert_pending_java_exception(&env);
-
-    assert!(env
-        .is_instance_of(ex, "java/lang/RuntimeException")
-        .unwrap());
-    let message = env
-        .call_method(ex, "getMessage", "()Ljava/lang/String;", &[])
-        .unwrap()
-        .l()
-        .unwrap();
-    let msg_rust: String = env.get_string(message.into()).unwrap().into();
-    assert_eq!(msg_rust, "Test Exception");
+    assert_pending_java_exception_detailed(
+        &env,
+        Some("java/lang/RuntimeException"),
+        Some("Test Exception"),
+    );
 }
 
 #[test]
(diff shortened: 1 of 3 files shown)


tests/jni_api.rs
```diff
@@ -611,8 +624,37 @@ fn assert_exception(res: Result<jobject, Error>, expect_message: &str) {
...
 }
 
-// Helper method that asserts there is a pending Java exception and clears if any
+// Shortcut to `assert_pending_java_exception_detailed()` without checking for expected  type and
+// message of exception.
 fn assert_pending_java_exception(env: &JNIEnv) {
+    assert_pending_java_exception_detailed(env, None, None)
+}
+
+// Helper method that asserts there is a pending Java exception of `expected_class` with
```
dmitry-timofeev: expected_type

tests/jni_api.rs
```diff
@@ -600,6 +592,27 @@ pub fn throw_new_fail() {
     assert_pending_java_exception(&env);
 }
 
+#[test]
+pub fn throw_defaults() {
+    let env = attach_current_thread();
+
+    assert_throws(&env, "Default exception thrown");
```
dmitry-timofeev: Why don't we instead of throwing just use `JThrowable.lookup(env, "foo")` and then assert on the exception object (type + message)?

tests/jni_api.rs
```diff
@@ -600,6 +592,27 @@ pub fn throw_new_fail() {
...
+    assert_throws(&env, "Default exception thrown".to_owned());
+    assert_throws(&env, JNIString::from("Default exception thrown"));
+}
+
+fn assert_throws<'a, D>(env: &JNIEnv<'a>, descriptor: D)
```
dmitry-timofeev: This method does not only assert, it also performs the test itself, hence the name is slightly misleading. Also, it does not accept the expected values, but has them as locals, which is implicit and duplicates what is in `throw_defaults`.

I suggest making the expected value explicit, and passing the thing to assert on directly (or renaming this method to `test_throwable_descriptors_with_default_type(env, descriptor, expected_message)`). Or use constants.

tests/jni_api.rs
skletsun: Well, I wanted also to cover the `.throw()` itself (we have tests for `.throw_new()`, but nothing for `.throw()`), but probably it's not the best idea.

tests/jni_api.rs
skletsun: It looks better with constants.

tests/jni_api.rs
```diff
@@ -25,6 +28,8 @@ static MATH_ABS_METHOD_NAME: &str = "abs";
 static MATH_TO_INT_METHOD_NAME: &str = "toIntExact";
 static MATH_ABS_SIGNATURE: &str = "(I)I";
 static MATH_TO_INT_SIGNATURE: &str = "(J)I";
+static DEFAULT_EXCEPTION_TYPE: &str = "java/lang/RuntimeException";
+static TEST_EXCEPTION_MESSAGE: &str = "Default exception thrown";
```
DarkEld3r: It is slightly disturbing to see this duplicated constant. :upside_down_face:

skletsun: Duplicated?.. 🤨

DarkEld3r: There are `DEFAULT_EXCEPTION_CLASS` and `DEFAULT_EXCEPTION_TYPE` with the same value.

tests/jni_api.rs
```diff
@@ -572,22 +577,11 @@ pub fn throw_new() {
...
-    let message = env
-        .call_method(ex, "getMessage", "()Ljava/lang/String;", &[])
-        .unwrap()
-        .l()
-        .unwrap();
-    let msg_rust: String = env.get_string(message.into()).unwrap().into();
-    assert_eq!(msg_rust, "Test Exception");
+    assert_pending_java_exception_detailed(
+        &env,
+        Some("java/lang/RuntimeException"),
```
DarkEld3r: Should `DEFAULT_EXCEPTION_TYPE` be used here?

skletsun: It's a different test and it is just a coincidence that it uses j.l.RuntimeException. Well... I see no reason making the `DEFAULT_EXCEPTION_CLASS` public. What if I define `RUNTIME_EXCEPTION_CLASS` (in analogue to the `EXCEPTION_CLASS` & `ARITHMETIC_EXCEPTION_CLASS`) and change all string representations to it (with no emphasis on default type of exception)?

skletsun: Implemented in the suggested way.

PR #196, opened 2019-07-22
Update Cargo.toml for better documentation

Enet4:
This updates the Cargo manifest with two improvements:

- Instruct docs.rs to build docs with the `invocation` feature, so that all necessary parts of the Java invocation API appear on docs.rs.
- Remove the `homepage` field from the manifest, as this one should only be used for dedicated project websites and should not be redundant with the repository or documentation ([C-METADATA](https://rust-lang-nursery.github.io/api-guidelines/documentation.html#c-metadata)).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,6 @@
 authors = ["Josh Chase <josh@prevoty.com>"]
 description = "Rust bindings to the JNI"
 documentation = "https://docs.rs/jni"
-homepage = "https://docs.rs/jni"
 readme = "README.md"
 keywords = [
     "ffi",
@@ -37,3 +36,6 @@ lazy_static = "1"
 invocation = []
 backtrace = ["error-chain/backtrace"]
 default = ["backtrace"]
+
+[package.metadata.docs.rs]
+features = ["invocation"]
--- a/src/wrapper/java_vm/init_args.rs
+++ b/src/wrapper/java_vm/init_args.rs
@@ -16,7 +16,10 @@ error_chain! {
     }
 }
 
-/// Builder for JavaVM InitArgs
+/// Builder for JavaVM InitArgs.
+/// 
+/// *This API requires "invocation" feature to be enabled, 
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -37,3 +36,6 @@ lazy_static = "1"
 invocation = []
 backtrace = ["error-chain/backtrace"]
 default = ["backtrace"]
+
+[package.metadata.docs.rs]
+features = ["invocation"]
```
dmitry-timofeev: Hm, I expected it to fail because building with invocation feature [requires a valid path to libjvm in LD_LIBRARY_PATH](https://github.com/jni-rs/jni-rs/blob/master/CONTRIBUTING.md#setup-environment), but `cargo doc --features=invocation --open` worked!

Is it a good practice to show the APIs that require a feature to be enabled? The generated docs unfortunately don't label such APIs at all:
image

dmitry-timofeev:
> because building with invocation feature requires a valid path to libjvm in LD_LIBRARY_PATH

Actually, building works (probably, auto-detecting the installed Java), but tests do not.

Cargo.toml
Enet4: Although the usual guidelines do not talk much about this mechanism, I usually include here all features that would extend the API, since at least they'll become visible to everyone who might be looking for them in docs.rs. Tests are not executed in this context, nor are dynamically linked objects required, which is a plus when some features require very specific resources.

dmitry-timofeev: I agree that it is worth including this API, as it is essential for one of the important use-cases — launching the VM from Rust. It is unfortunate that Rustdoc cannot label such methods automatically. In order to avoid confusion, I think we can add a note as the last paragrapth to the documentation of JavaVM#new, InitArgs and InitArgsBuilder:
```
This API requires "invocation" feature to be enabled, 
see ["Launching JVM from Rust"](struct.JavaVM.html#launching-jvm-from-rust).
```

DarkEld3r: I also think that adding such note is a good idea.

Enet4:
My apologies for the delay. I have added the extra piece of documentation, as suggested.

PR #206, opened 2019-10-22
Improve objects conversion.

skletsun:
## Overview

- Improved conversion from object wrappers to `JValue` - they can now be converted into `JValue` without intermediate `JObject`.

### Definition of Done


skletsun:
@dmitry-timofeev Is it worth updating JNIEnv API (by changing the type of argument for target object from JObject to impl AsObj) within this PR? On the one hand, it changes the public API, but on the other hand, it remains compatible with the existing client's code.

vitvakatu:
> On the one hand, it changes the public API, but on the other hand, it remains compatible with the existing client's code.

Will it actually break the API, given that `AsObj` is implemented for `JObject` itself?

PR #209, opened 2019-10-28
Separate descriptor lifetimes in `get_static_field`

jvff:
## Overview

This PR loosens the lifetime constraints for the `get_static_field` and `get_static_field_unchecked` methods. Previously, the class and static field ID descriptors had a very strict lifetime requirement, which led the borrow checker to think that borrowing a `GlobalRef` (to use as a class descriptor) would require the returned `JValue` to live as long as the `GlobalRef` borrow. By creating new separate lifetimes for the descriptors, this is no longer the case.

These changes should not be breaking, because previously accepted code would adhere to stricter requirements, which are still accepted.

### Definition of Done

### DIFF
--- a/src/wrapper/jnienv.rs
+++ b/src/wrapper/jnienv.rs
@@ -1634,15 +1634,15 @@ impl<'a> JNIEnv<'a> {
 
     /// Get a static field without checking the provided type against the actual
     /// field.
-    pub fn get_static_field_unchecked<T, U>(
+    pub fn get_static_field_unchecked<'c, 'f, T, U>(
         &self,
         class: T,
         field: U,
         ty: JavaType,
     ) -> Result<JValue<'a>>
     where
-        T: Desc<'a, JClass<'a>>,
-        U: Desc<'a, JStaticFieldID<'a>>,
+        T: Desc<'a, JClass<'c>>,
+        U: Desc<'a, JStaticFieldID<'f>>,
     {
         let class = class.lookup(self)?.into_inner();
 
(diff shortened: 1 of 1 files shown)


jvff:
I realized that there are also some other methods in `JNIEnv` that are in a similar situation. Should I also apply similar changes to them and include them in this PR?

src/wrapper/jnienv.rs
```diff
@@ -1634,15 +1634,15 @@ impl<'a> JNIEnv<'a> {
...
+    pub fn get_static_field_unchecked<'c, 'f, T, U>(
         &self,
         class: T,
         field: U,
         ty: JavaType,
     ) -> Result<JValue<'a>>
     where
-        T: Desc<'a, JClass<'a>>,
-        U: Desc<'a, JStaticFieldID<'a>>,
+        T: Desc<'a, JClass<'c>>,
```
dmitry-timofeev: Does such modification make sense in all the other usages in JNIEnv, where `Desc<'a, JClass/Field/MethodId<'a>>` is accepted, so that they are compatible with global refs  and field/method ids (which have distinct from JNIEnv lifetimes and are allowed to outlive JNIEnv)?

jvff: Yes, I believe they are also candidates for similar changes. Should they be included in this PR?

dmitry-timofeev:
@jvff We've noticed at the same time :-) If it's not too much trouble for you to update them in the remaining cases, it would be wonderful!

jvff:
Yeah, perfect timing :smile: 

Sure! I'll push an update to the PR with the extra changes :slightly_smiling_face:

jvff:
I've pushed more commits to add the same lifetime separation changes to the other methods I found that had similar restrictions.

PR #216, opened 2019-12-04
Remove logging facilities from jni-rs

maplant:
## Overview

I'm not sure if this commit is _exactly_ ready to be merged, but I'd like to put it here to pose the question.
I do believe that logging should be removed from jni-rs. It's not a great way to report information, and pollutes logs.
However I can see it being useful for conveying drop errors. Maybe the more sensible solution is to gate logging behind a feature flag?

Anyway, here is my branch for anyone that does not want jni-rs to log. cheers.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,6 @@ edition = "2018"
 cesu8 = "1.1.0"
 combine = "4.1.0"
 jni-sys = "0.3.0"
-log = "0.4.4"
 
 [build-dependencies]
 walkdir = "2"
--- a/src/wrapper/java_vm/vm.rs
+++ b/src/wrapper/java_vm/vm.rs
@@ -3,10 +3,7 @@ use std::{
     ops::Deref,
     ptr,
     sync::atomic::{AtomicUsize, Ordering},
-    thread::current,
 };
-
-use log::{debug, error};
 
 use crate::{errors::*, sys, JNIEnv};
 
@@ -381,13 +378,6 @@ impl InternalAttachGuard {
 
         ATTACHED_THREADS.fetch_add(1, Ordering::SeqCst);
 
-        debug!(
-            "Attached thread {} ({:?}). {} threads attached",
-            current().name().unwrap_or_default(),
-            current().id(),
-            ATTACHED_THREADS.load(Ordering::SeqCst)
-        );
-
         Ok(env_ptr as *mut sys::JNIEnv)
     }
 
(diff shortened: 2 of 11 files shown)


vitvakatu:
Hello.

What logging library are you using in your project? All of them have some methods to restrict logs from different log levels/modules. For example, `env_logger` allows you to set environment variable `RUST_LOG`:

`export RUST_LOG="${RUST_LOG-jni=error}"`

This will log only `error`-level messages from `jni` crate.

maplant:
@vitvakatu I'm using log4rs and indeed it is able to filter these messages. However, my opinion is that some of these log messages are extraneous. For a library, I would argue there should be no log messages less important than warn

DarkEld3r:
As for me, logging can be useful even for the users of a library. Though I'm not sure what best practices on this topic in Rust are. I don't know if jni-rs is a specific case either. I'm going to check popular libraries to form an opinion.

dmitry-timofeev:
Hi, thanks for bringing it up,

I agree with @vitvakatu that it is the responsibility of the application to configure the logging according to its needs. From my Java perspective, the usual configuration is: "Log my application messages however I see fit — debug/info, log other messages from the dependencies at 'warn'". In Rust `env_logger` words, `RUST_LOG=warn,myapp=info ./myapp`.

I agree that the users of a library/framework shall not _see_ logs below warn, but I don't think it shall be achieved by forbidding the library code to use info and debug — that is what for the  logging backend configuration. Is that different in Rust ecosystem?
 
The way I see the separation of logs by levels for a library in terms of target audience: 
- above 'warn' — end-users (to provide extra context when things go bad, to provide _any_ information when it cannot be communicated in other way, to warn about mis-use or misconfiguration. For instance, when a user runs a 1 second operation on an event loop, it is nice of the event loop library to warn that that is likely a terrible mistake)
- below 'warn' — mostly library developers _or_ particularly interested in library operation users, e.g., troubleshooting a bug.

Therefore, I'd focus on those warning+ messages of jni that do not seem appropriate and consider what to do with them on a case-by-case basis: 
* remove, 
* throttle, if easily possible, 
* change the level, etc.

Are there any inappropriate warnings that you think shall be reconsidered?

lucaszanella:
I don't get it. Should `log!` simply works on the Rust side and it would appear in logcat? On my case it does not appear, but I also have no way of setting the environment variable for proper logging.

I just wanted to use Log.d on Rust's side

PR #223, opened 2020-02-20
Fix clippy warning

vitvakatu:
### Definition of Done

### DIFF
--- a/src/wrapper/jnienv.rs
+++ b/src/wrapper/jnienv.rs
@@ -1800,7 +1800,7 @@ impl<'a> JNIEnv<'a> {
             // attempt to acquire the lock. This prevents us from consuming the
             // mutex if there's an outstanding lock. No one else will be able to
             // get a new one as long as we're in the guarded scope.
-            let _ = mbox.try_lock()?;
+            std::mem::drop(mbox.try_lock()?);
 
             self.set_field_unchecked(
                 obj,
--- a/src/wrapper/jnienv.rs
+++ b/src/wrapper/jnienv.rs
@@ -1800,7 +1800,7 @@ impl<'a> JNIEnv<'a> {
             // attempt to acquire the lock. This prevents us from consuming the
             // mutex if there's an outstanding lock. No one else will be able to
             // get a new one as long as we're in the guarded scope.
-            std::mem::drop(mbox.try_lock()?);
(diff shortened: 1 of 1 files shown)


src/wrapper/jnienv.rs
```diff
@@ -1800,7 +1800,7 @@ impl<'a> JNIEnv<'a> {
             // attempt to acquire the lock. This prevents us from consuming the
             // mutex if there's an outstanding lock. No one else will be able to
             // get a new one as long as we're in the guarded scope.
-            let _ = mbox.try_lock()?;
+            std::mem::drop(mbox.try_lock()?);
```
dmitry-timofeev: What is happening here? Why does it immediately unlock the acquired mutex? 😕

vitvakatu: As we receiving this mutex via raw pointer, we're checking that the mutex isn't locked by anyone right now. Actually I'm not sure if it's a valid behaviour or not

dmitry-timofeev: Ok, so it
1. Locks Java object (using its monitor)
2. Tries to lock the native object (using the mutex it is wrapped in)
3. If it is unlocked — clears the "Rust field" of the Java object and returns the unlocked mutex.
4. Finally , unlocks the Java monitor.

DarkEld3r:
```suggestion
            drop(mbox.try_lock()?);
```
Isn't drop imported by the prelude?

dmitry-timofeev: Would it make sense to clarify anything a little?

vitvakatu: Well, it is exactly as you wrote it. I think additional clarification is not necessary, docs of used methods and this comment are enough to understand the purpose of the code.

dmitry-timofeev: >  I think additional clarification is not necessary, docs of used methods and this comment are enough to understand the purpose of the code. 

I described what happens, but not _why_ it happens. If the Java lock is used, then how Rust lock can be 'locked'? Does it acquisition outlive the Java monitor acquisition?

dmitry-timofeev:
OK, I think we'll improve this documentation some time later 🙃

dmitry-timofeev:
Thanks!

PR #228, opened 2020-03-09
add wrappers for java.io.{OutputStream, InputStream}

dlesl:
## Overview

I was thinking this would be nice to have in jni-rs. It's useful to have if you want to use a Rust parsing library on a Java stream, for example. 

### Definition of Done

### DIFF
--- a/src/wrapper/objects/jinputstream.rs
+++ b/src/wrapper/objects/jinputstream.rs
@@ -1,0 +1,86 @@
+use crate::{
+    errors::Result,
+    objects::{AutoLocal, JMethodID, JObject},
+    signature::{JavaType, Primitive},
+    sys::jsize,
+    JNIEnv,
+};
+use std::io;
+
+const DEFAULT_BUF_SIZE: usize = 8 * 1024; // same as std::io
+
+/// Wrapper for JObjects that implement `java.io.InputStream`.
+pub struct JInputStream<'a: 'b, 'b> {
+    internal: JObject<'a>,
+    buffer: AutoLocal<'a, 'b>,
+    read: JMethodID<'a>,
+    env: &'b JNIEnv<'a>,
+}
+
+impl<'a: 'b, 'b> std::ops::Deref for JInputStream<'a, 'b> {
+    type Target = JObject<'a>;
+
+    fn deref(&self) -> &Self::Target {
+        &self.internal
+    }
+}
+
+impl<'a: 'b, 'b> From<JInputStream<'a, 'b>> for JObject<'a> {
+    fn from(other: JInputStream<'a, 'b>) -> JObject<'a> {
+        other.internal
+    }
+}
+
+impl<'a: 'b, 'b> JInputStream<'a, 'b> {
+    /// Wrap an environment and an object that implements
+    /// `java.io.InputStream`, specifying the capacity of the internal
+    /// buffer.
+    pub fn from_env_with_capacity(
+        capacity: usize,
+        env: &'b JNIEnv<'a>,
+        obj: JObject<'a>,
+    ) -> Result<JInputStream<'a, 'b>> {
+        let class = env.auto_local(env.find_class("java/io/InputStream")?);
+
+        let read = env.get_method_id(&class, "read", "([B)I")?;
+
+        let buffer = env.auto_local(env.new_byte_array(capacity as jsize)?);
+
+        Ok(JInputStream {
+            internal: obj,
+            buffer,
+            read,
+            env,
+        })
+    }
+    /// Wrap an environment and an object that implements `java.io.InputStream`.
+    pub fn from_env(env: &'b JNIEnv<'a>, obj: JObject<'a>) -> Result<JInputStream<'a, 'b>> {
+        Self::from_env_with_capacity(DEFAULT_BUF_SIZE, env, obj)
+    }
+}
+
+impl<'a: 'b, 'b> io::Read for JInputStream<'a, 'b> {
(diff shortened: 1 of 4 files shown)


src/wrapper/objects/jinputstream.rs
```diff
@@ -0,0 +1,86 @@
+use crate::{
+    errors::*,
```
vitvakatu: nit: Blob imports are considered bad practice, as it's impossible to see what exactly came from this module.

src/wrapper/objects/jinputstream.rs
```diff
@@ -0,0 +1,86 @@
...
+use std::io;
+
+const DEFAULT_BUF_SIZE: usize = 8 * 1024; // same as std::io
+
+/// Wrapper for JObjects that implement `java.io.InputStream`.
+pub struct JInputStream<'a: 'b, 'b> {
+    internal: JObject<'a>,
+    buffer: AutoLocal<'a, 'b>,
+    read: JMethodID<'a>,
+    env: &'b JNIEnv<'a>,
```
vitvakatu: JNIEnv is actually clonable. Is there a reason to use reference here?

src/wrapper/objects/jinputstream.rs
```diff
@@ -0,0 +1,86 @@
...
+}
+
+impl<'a: 'b, 'b> ::std::ops::Deref for JInputStream<'a, 'b> {
```
vitvakatu: style nit: In modern Rust, I can drop `::` at the beginning of the import, so use `std::ops::Deref`.

dlesl:
Hi @dmitry-timofeev,
Thanks for taking a look at this. I agree that it's arguable whether this is within the scope of jni-rs. On the other hand, as with the `List` and `Map` wrappers, the wrapper is very simple.
I thought about whether it might be better in a separate crate, however as far as I could tell this would be problematic because the jni-rs dependency would constantly need to be kept up to date?

Regarding performance, while it's probably best to avoid excessive interop, in this case it seems like the overhead is quite low. A [quick unscientific benchmark](https://github.com/dlesl/jni_stream_bench/blob/master/src/main.rs):
```
$ dd if=/dev/zero bs=1000000 count=10000 | target/release/jni_stream_bench native >/dev/null
10000+0 records in
10000+0 records out
10000000000 bytes (10 GB, 9.3 GiB) copied, 8.57863 s, 1.2 GB/s
$ dd if=/dev/zero bs=1000000 count=10000 | target/release/jni_stream_bench java-in >/dev/null
10000+0 records in
10000+0 records out
10000000000 bytes (10 GB, 9.3 GiB) copied, 12.6121 s, 793 MB/s
$ dd if=/dev/zero bs=1000000 count=10000 | target/release/jni_stream_bench java-out >/dev/null
10000+0 records in
10000+0 records out
10000000000 bytes (10 GB, 9.3 GiB) copied, 10.7353 s, 932 MB/s
```
This is using the default 8 KB buffer size.

src/wrapper/objects/jinputstream.rs
dlesl: Thanks! Fixed

src/wrapper/objects/jinputstream.rs
dlesl: Actually I just copied and pasted `jmap.rs` to begin with 😄, and this is the way it's done there. I agree that there's nothing to be gained by taking a reference since it's just a pointer internally, but I suppose it's good to do it consistently in all the wrappers?

src/wrapper/objects/jinputstream.rs
dlesl: Done

src/wrapper/objects/jinputstream.rs
vitvakatu: I think `jmap.rs` was there before `JNIEnv` was made clonable :)

It looks you will still need second lifetime parameter, so we can leave it as it is.

dmitry-timofeev:
> I thought about whether it might be better in a separate crate, however as far as I could tell this would be problematic because the jni-rs dependency would constantly need to be kept up to date?

Yes, quite likely. It definitely is easier to refactor this code if it lives with the main JNI crate. Would something like a satellite crate in the same repository with some Java proxies work? It won't solve the issues with unclear scope, but will allow experimenting with this feature without affecting the main crate.

> Regarding performance, while it's probably best to avoid excessive interop, in this case it seems like the overhead is quite low.

It certainly depends on the use case. The last time I measured, native -> Java with proper caching of classes and method IDs gave about 5–8M ops/s, Java->native about 80M ops/s. My point is including proxies of Java classes will encourage their direct use and more active interop when some alternatives could be better (e.g., copying a value object from Java to Rust and then operating on the Rust object). Such proxies still have value — as they give a ready to use thing, but that option is not universally best.

rib:
Hi,

Sorry that this PR / discussion went quiet for over two years without a resolution.

I'm going to close this for now, partly because it's quite an old pull request that will involve work to rebase, but also because:
- There have been some recent discussions about how to improve safety with some changes to make JObject references non-`Copy` and potentially also affect how the `JNIEnv` type is accessed (ref: https://github.com/jni-rs/jni-rs/issues/392)
- I've been trying to think of ways of supporting improved ergonomics generally for proxy types - since it's very unfortunate currently that they can't leverage any kind of method ID cache.

... and so I tend to think that minimizing the number of built-in proxy APIs should hopefully help keep it more practical to explore how the `jni-rs` crate / API can be improved - at least for now.

This can be re-considered and potentially re-opened in the future.

PR #237, opened 2020-06-14
Get/release byte array elements. Get/release primitive array critical

maurolacy:
## Overview

Hello,

I've implemented wrappers around [Get/ReleaseByteArrayElements](https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/functions.html#Get_PrimitiveType_ArrayElements_routines), and [Get/ReleasePrimitiveArrayCritical](https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/functions.html#GetPrimitiveArrayCritical_ReleasePrimitiveArrayCritical).

I understand these functions are `unsafe`, as they return raw pointers. But they can still be convenient for performance reasons, given precautions are taken when allocating / releasing the pointed-to data. There are a couple examples on how to do this in the unit tests.

I have a series of questions regarding these functions:

- Why haven't you implemented these wrappers already?

- Do you know of better ways to use these functions, than the ones documented in the unit tests? With `get_byte_array_elements()` in particular, I haven't been able to avoid copying the array (i.e. `is_copy` always return true). At least with OpenJDK 8, and the default GC. I plan to test other VMs and other GCs, so I can document these eventually.
With `get_primitive_array_critical()`, copy is avoided (i.e. `is_copy` set to false). But the cost overall is relatively high, as it's not recommended to call other JNI functions during critical sections. The current test cases in fact throw a Warning. I suppose, because test cases are running in parallel. Do you know of specific problems if these recommendations are violated? I've been unable to detect them, and in fact all unit tests are still passing, despite the warning.

- Are you interested in these implementations? In particular, there's a family of related functions (`Get<PrimitiveType>ArrayElements`, and `Get/ReleaseStringCritical`) that can still be wrapped up.

- Regarding benchmarks, I've detected a ~15% performance improvement using `get_primitive_array_critical()`, compared to `get_byte_array_region()` (which always copies the array), and depending on array length. I can reflect / document these in the benchmarks if you want.

I've searched the issues for this and found nothing related. If you'd rather I create an issue to follow up on this, say so and I will.

Regards,
Mauro

### Definition of Done

### DIFF
--- a/src/wrapper/jnienv.rs
+++ b/src/wrapper/jnienv.rs
@@ -2020,8 +2020,7 @@ impl<'a> JNIEnv<'a> {
         mode: ReleaseMode,
     ) -> Result<AutoByteArray> {
         let (ptr, is_copy) = self.get_byte_array_elements(array)?;
-        let res = AutoByteArray::new(self, array.into(), ptr, mode, is_copy);
-        Ok(res)
+        AutoByteArray::new(self, array.into(), ptr, mode, is_copy)
     }
 
     /// Return a tuple with a pointer to elements of the given Java primitive array as first
@@ -2133,8 +2132,7 @@ impl<'a> JNIEnv<'a> {
         mode: ReleaseMode,
     ) -> Result<AutoPrimitiveArray> {
         let (ptr, is_copy) = self.get_primitive_array_critical(array)?;
-        let res = AutoPrimitiveArray::new(self, array.into(), ptr, mode, is_copy);
-        Ok(res)
+        AutoPrimitiveArray::new(self, array.into(), ptr, mode, is_copy)
     }
 }
 
--- a/src/wrapper/jnienv.rs
+++ b/src/wrapper/jnienv.rs
@@ -1933,7 +1933,7 @@ impl<'a> JNIEnv<'a> {
         jni_error_code_to_result(res)
     }
 
-    /// Return elements of the given Java byte array.
+    /// Return a pointer to elements of the given Java byte array.
     ///
     /// The result is valid until the corresponding release_byte_array_elements() function is
     /// called. Since the returned array may be a copy of the Java array, changes made to the
@@ -1983,6 +1983,63 @@ impl<'a> JNIEnv<'a> {
         jni_void_call!(
             self.internal,
             ReleaseByteArrayElements,
+            array,
+            elems,
+            mode
+        );
+        Ok(())
+    }
+
+    /// Return a pointer to elements of the given Java array.
+    ///
+    /// The semantics of this function is very similar to the existing
+    /// get_<primitivetype>_array_elements functions. If possible, the VM returns a pointer to the
+    /// primitive array; otherwise, a copy is made. However, there are significant restrictions
+    /// on how these functions can be used.
+    ///
+    /// After calling get_primitive_array_critical, the native code should not run for an extended
+    /// period of time before it calls release_primitive_array_critical. We must treat the code
+    /// inside this pair of functions as running in a "critical region." Inside a critical region,
+    /// native code must not call other JNI functions, or any system call that may cause the
+    /// current thread to block and wait for another Java thread. (For example, the current
+    /// thread must not call read on a stream being written by another Java thread.)
+    ///
+    /// These restrictions make it more likely that the native code will obtain an uncopied version
+    /// of the array, even if the VM does not support pinning. For example, a VM may temporarily
+    /// disable garbage collection when the native code is holding a pointer to an array obtained
+    /// via get_primitive_array_critical.
+    ///
+    /// Note that get_primitive_array_critical might still make a copy of the array if the VM
+    /// internally represents arrays in a different format. Therefore, we need to check its
+    /// return value against NULL for possible out of memory situations.
+    pub fn get_primitive_array_critical(
+        &self,
+        array: jarray,
+        is_copy: *mut jboolean,
+    ) -> Result<*mut c_void> {
+        non_null!(array, "get_primitive_array_critical array argument");
+        let res = jni_non_void_call!(
+            self.internal,
+            GetPrimitiveArrayCritical,
+            array,
+            is_copy
+        );
+        Ok(res)
+    }
+
+    /// Release elements of the given array.
+    ///
+    /// See get_primitive_array_critical for a discussion on how these functions can be used.
+    pub fn release_primitive_array_critical(
+        &self,
+        array: jarray,
+        elems: *mut c_void,
+        mode: jint
+    ) -> Result<()> {
+        non_null!(array, "release_primitive_array_critical array argument");
+        jni_void_call!(
+            self.internal,
+            ReleasePrimitiveArrayCritical,
             array,
             elems,
             mode
--- a/src/wrapper/jnienv.rs
+++ b/src/wrapper/jnienv.rs
@@ -8,12 +8,14 @@ use std::{
 
 use log::warn;
 
+use crate::wrapper::objects::ReleaseMode;
 use crate::{
     descriptors::Desc,
     errors::*,
     objects::{
-        AutoLocal, GlobalRef, JByteBuffer, JClass, JFieldID, JList, JMap, JMethodID, JObject,
-        JStaticFieldID, JStaticMethodID, JString, JThrowable, JValue,
+        AutoByteArray, AutoLocal, AutoPrimitiveArray, GlobalRef, JByteBuffer, JClass, JFieldID,
+        JList, JMap, JMethodID, JObject, JStaticFieldID, JStaticMethodID, JString, JThrowable,
+        JValue,
     },
     signature::{JavaType, Primitive, TypeSignature},
     strings::{JNIString, JavaStr},
@@ -435,6 +437,38 @@ impl<'a> JNIEnv<'a> {
             AllocObject,
             class.into_inner()
         ))
+    }
+
+    /// Creates a new auto-released byte array.
+    ///
+    /// See also [`get_byte_array_elements_auto`](struct.JNIEnv.html#method.get_byte_array_elements_auto)
+    pub fn auto_byte_array<'b, O>(
+        &'b self,
+        obj: O,
+        ptr: *mut jbyte,
+        mode: ReleaseMode,
+        is_copy: bool,
+    ) -> AutoByteArray<'a, 'b>
+    where
+        O: Into<JObject<'a>>,
+    {
+        AutoByteArray::new(self, obj.into(), ptr, mode, is_copy)
+    }
+
+    /// Creates a new auto-released primitive array.
+    ///
+    /// See also [`get_primitive_array_critical_auto`](struct.JNIEnv.html#method.get_primitive_array_critical_auto)
+    pub fn auto_primitive_array<'b, O>(
+        &'b self,
+        obj: O,
+        ptr: *mut c_void,
+        mode: ReleaseMode,
+        is_copy: bool,
+    ) -> AutoPrimitiveArray<'a, 'b>
+    where
+        O: Into<JObject<'a>>,
+    {
+        AutoPrimitiveArray::new(self, obj.into(), ptr, mode, is_copy)
     }
 
     /// Common functionality for finding methods.
@@ -1960,9 +1994,8 @@ impl<'a> JNIEnv<'a> {
     /// The mode argument provides information on how the array buffer should be released. mode
     /// has no effect if elems is not a copy of the elements in array. Otherwise, mode has the
     /// following impact:
-    /// 0: Copy back the content and free the elems buffer.
-    /// JNI_COMMIT: Copy back the content but do not free the elems buffer.
-    /// JNI_ABORT: Free the buffer without copying back the possible changes.
+    /// Copy: Copy back the content and free the elems buffer.
+    /// NoCopy: Free the buffer without copying back the possible changes.
     ///
     /// In most cases, programmers pass “0” to the mode argument to ensure consistent behavior
     /// for both pinned and copied arrays. The other options give the programmer more control
@@ -1971,16 +2004,61 @@ impl<'a> JNIEnv<'a> {
         &self,
         array: jbyteArray,
         elems: &mut jbyte,
-        mode: jint,
+        mode: ReleaseMode,
     ) -> Result<()> {
         non_null!(array, "release_byte_array_elements array argument");
-        jni_void_call!(self.internal, ReleaseByteArrayElements, array, elems, mode);
+        jni_void_call!(
+            self.internal,
+            ReleaseByteArrayElements,
+            array,
+            elems,
+            mode as i32
+        );
         Ok(())
     }
 
-    /// Return a pointer to elements of the given Java array.
+    /// Commit elements of the given byte array.
     ///
-    /// The semantics of this function is very similar to the existing
+    /// This function has no effect if elems is not a copy of the elements in array. Otherwise,
+    /// this function copies back the content of the array (and does not free the elems buffer).
+    pub fn commit_byte_array_elements(&self, array: jbyteArray, elems: &mut jbyte) -> Result<()> {
+        non_null!(array, "commit_byte_array_elements array argument");
+        jni_void_call!(
+            self.internal,
+            ReleaseByteArrayElements,
+            array,
+            elems,
+            jni_sys::JNI_COMMIT
+        );
+        Ok(())
+    }
+
+    /// Return an AutoByteArray of the given Java byte array.
+    ///
+    /// The result is valid until the corresponding AutoByteArray object goes out of scope, when the
+    /// release happens automatically according to the mode parameter.
+    ///
+    /// Since the returned array may be a copy of the Java array, changes made to the
+    /// returned array will not necessarily be reflected in the original array until
+    /// release_array_elements() is called.
+    /// AutoByteArray has a commit() method, to force a copy of the array if needed (and without
+    /// releasing it).
+    ///
+    /// See also [`get_byte_array_elements`](struct.JNIEnv.html#method.get_byte_array_elements)
+    pub fn get_byte_array_elements_auto(
+        &self,
+        array: jbyteArray,
+        mode: ReleaseMode,
+    ) -> Result<AutoByteArray> {
+        let mut is_copy: jboolean = 0xff;
+        let ptr = self.get_byte_array_elements(array, &mut is_copy).unwrap();
+        let res = self.auto_byte_array(array, ptr, mode, is_copy != 0);
+        Ok(res)
+    }
+
+    /// Return a pointer to elements of the given Java primitive array.
+    ///
+    /// The semantics of this function are very similar to the existing
     /// get_<primitivetype>_array_elements functions. If possible, the VM returns a pointer to the
     /// primitive array; otherwise, a copy is made. However, there are significant restrictions
     /// on how these functions can be used.
@@ -2017,7 +2095,7 @@ impl<'a> JNIEnv<'a> {
         &self,
         array: jarray,
         elems: &mut c_void,
-        mode: jint,
+        mode: ReleaseMode,
     ) -> Result<()> {
         non_null!(array, "release_primitive_array_critical array argument");
         jni_void_call!(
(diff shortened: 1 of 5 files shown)


DarkEld3r:
Thanks for the pull request!

I'm going to comment on the changes themselves, but at first I would like to answer your questions:

> Why haven't you implemented these wrappers already?

Historically this library has evolved to suit the needs of its users, so it doesn't cover all JNI interface. So if you need something it is right to submit a pull request or at least an issue.

> Do you know of better ways to use these functions, than the ones documented in the unit tests?

I have mostly C++ and Rust experience, so unfortunately I'm unable to answer this question. Perhaps @dmitry-timofeev can add something.

> Are you interested in these implementations?

Personally, I think that the more parts of JNI we have covered, the better. Sometimes there are caveats: it is non-trivial to expose some parts of the JNI api through a safe interface, so I would prefer to omit them. At least till someone really needs them and can explain the usage. But I don't think this is a case.

> I can reflect / document these in the benchmarks if you want.

It would be a nice addition to the documentation!

> If you'd rather I create an issue to follow up on this, say so and I will.

Don't worry, I don't think that we need an issue if we already have the pull request.

src/wrapper/jnienv.rs
```diff
@@ -1932,6 +1932,103 @@ impl<'a> JNIEnv<'a> {
...
+    /// called. Since the returned array may be a copy of the Java array, changes made to the
+    /// returned array will not necessarily be reflected in the original array until
+    /// release_array_elements() is called.
+    ///
+    /// is_copy is set to JNI_TRUE if a copy is made; or it is set to JNI_FALSE if no copy is made.
+    pub fn get_byte_array_elements(
+        &self,
+        array: jbyteArray,
+        is_copy: &mut jboolean,
+    ) -> Result<*mut jbyte> {
```
DarkEld3r, line 1948: Perhaps, we can return some wrapper similar to `AutoLocal` instead of the raw pointer?

src/wrapper/jnienv.rs
```diff
@@ -1932,6 +1932,103 @@ impl<'a> JNIEnv<'a> {
...
+    /// JNI_ABORT: Free the buffer without copying back the possible changes.
+    ///
+    /// In most cases, programmers pass “0” to the mode argument to ensure consistent behavior
+    /// for both pinned and copied arrays. The other options give the programmer more control
+    /// over memory management, and should be used with extreme care.
+    pub fn release_byte_array_elements(
+        &self,
+        array: jbyteArray,
+        elems: &mut jbyte,
+        mode: jint,
```
DarkEld3r, line 1974: I prefer to create an enum for the mode instead of using `jint`.

src/wrapper/jnienv.rs
maurolacy: Sure. I was thinking on returning a `Vec` instead of a raw pointer, but I think your suggestion is better / more general.

src/wrapper/jnienv.rs
maurolacy: OK.

src/wrapper/jnienv.rs
maurolacy: I've implemented the auto-release wrappers for byte and primitive arrays. I've left the original manual / unsafe functions in place, but I'm OK with removing them, if you prefer so.

Was trying to generalize the `AutoArray` containers using generics (as the problem looks highly generalizable). I _almost_ succeeded in doing so, but was stumped by some issues regarding the subjacent JNI API, which has type-specific methods for the `*ArrayElements` family of functions. The tokens in the wrappers need to be type dependant, and I currently don't know how to do that.
So, in the end, I resorted to using two different array containers, 'AutoByteArray`, and `AutoPrimitiveArray` :-/ . Maybe you can suggest better names for these containers, and/or an elegant approach at generalization. I'm OK with iterating over this again, although my Rust knowledge is rapidly reaching its limit here.

Finally, some unit tests for the auto variants are still missing. They will be pushed soon, no later than early next week.

src/wrapper/jnienv.rs
maurolacy: Done. As part of this, I created separate functionality for releasing and commiting the array, as I didn't want to introduce a bogus "release" mode (like in the original API :-).

src/wrapper/jnienv.rs
maurolacy: Update: unit tests added.

src/wrapper/jnienv.rs
```diff
@@ -1932,6 +1966,192 @@ impl<'a> JNIEnv<'a> {
...
+    ///
+    /// See also [`get_byte_array_elements`](struct.JNIEnv.html#method.get_byte_array_elements)
+    pub fn get_byte_array_elements_auto(
+        &self,
+        array: jbyteArray,
+        mode: ReleaseMode,
+    ) -> Result<AutoByteArray> {
+        let mut is_copy: jboolean = 0xff;
+        let ptr = self.get_byte_array_elements(array, &mut is_copy).unwrap();
+        let res = self.auto_byte_array(array, ptr, mode, is_copy != 0);
```
dmitry-timofeev, line 2055: 0 -> JNI_FALSE ?

src/wrapper/jnienv.rs
```diff
@@ -1932,6 +1966,192 @@ impl<'a> JNIEnv<'a> {
...
+            elems,
+            jni_sys::JNI_COMMIT
+        );
+        Ok(())
+    }
+
+    /// Return an AutoPrimitiveArray of the given Java primitive array.
+    ///
+    /// The result is valid until the corresponding AutoPrimitiveArray object goes out of scope,
+    /// when the release happens automatically according to the mode parameter.
```
dmitry-timofeev, line 2134: Do I understand correctly that this API might make the unsuspecting user non-intentionally break the conditions of "GetCritical" by using this method in a big scope: get for as short period as possible, release as soon as possible? E.g., 

```
var array = GetPrimitiveArrayCriticalAuto();
// Use array

// Do lots of other, potentially, long-running, stuff

// Finally, release it when it goes out of scope
```

So the questions are: 
- Shall we provide the auto flavour for critical arrays?
- If we do, shall we advise to consider explicit std::mem::drop if they are acquired in biggish scopes?

src/wrapper/objects/auto_byte_array.rs
```diff
@@ -0,0 +1,115 @@
...
+
+    /// Forget the wrapper, returning the original pointer.
+    ///
+    /// This prevents `release_byte_array_elements` from being called when the `AutoArray`
+    /// gets dropped. You must remember to release the array manually.
+    pub fn forget(self) -> *mut jbyte {
+        let ptr = self.ptr;
+        mem::forget(self);
+        ptr
+    }
```
dmitry-timofeev, line 64: Is there a use-case when this is needed? I.e., is there a reason to use AutoArray instead of a pair of get/release when you also `forget` it?

src/wrapper/objects/auto_byte_array.rs
```diff
@@ -0,0 +1,115 @@
...
+
+    /// Commits the result of the array, if it is a copy
+    pub fn commit(&self) {
+        if !self.is_copy {
+            return;
+        }
+        let res = self
+            .env
+            .commit_byte_array_elements(*self.obj, unsafe { self.ptr.as_mut() }.unwrap());
```
dmitry-timofeev, line 84: We may remove the field `is_copy` to keep the code slightly simpler, as I think `release_...` already handles that possibility (committing to a non-copy).

dmitry-timofeev:
What's interesting is that Get*Elements appeared to be no better than `Get*Region` for most sizes (though, on JDK 8, on 14 the picture might be different): https://habr.com/ru/post/222997/ The post is in Russian, but the diagram with the results is in English.

src/wrapper/jnienv.rs
```diff
@@ -1932,6 +1966,192 @@ impl<'a> JNIEnv<'a> {
...
+    /// is a pointer derived from array using the corresponding get_byte_array_elements() function.
+    /// If necessary, this function copies back all changes made to elems to the original array.
+    ///
+    /// The mode argument provides information on how the array buffer should be released. mode
+    /// has no effect if elems is not a copy of the elements in array. Otherwise, mode has the
+    /// following impact:
+    /// Copy: Copy back the content and free the elems buffer.
+    /// NoCopy: Free the buffer without copying back the possible changes.
+    ///
+    /// In most cases, programmers pass “0” to the mode argument to ensure consistent behavior
```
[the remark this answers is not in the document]
maurolacy: Yes, sorry, I've noticed documentation is still outdated / incomplete in some parts. Will work on that as soon as possible.

src/wrapper/jnienv.rs
```diff
@@ -1932,6 +1966,192 @@ impl<'a> JNIEnv<'a> {
...
+    /// Release elements of the given byte array.
+    ///
+    /// Informs the VM that the native code no longer needs access to elems. The elems argument
+    /// is a pointer derived from array using the corresponding get_byte_array_elements() function.
+    /// If necessary, this function copies back all changes made to elems to the original array.
+    ///
+    /// The mode argument provides information on how the array buffer should be released. mode
+    /// has no effect if elems is not a copy of the elements in array. Otherwise, mode has the
+    /// following impact:
+    /// Copy: Copy back the content and free the elems buffer.
```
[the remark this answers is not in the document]
maurolacy: Perfect. I'll add a reference in the documentation. As well as thoroughly review the docs / comments.

src/wrapper/jnienv.rs
```diff
@@ -437,6 +439,38 @@ impl<'a> JNIEnv<'a> {
         ))
     }
 
+    /// Creates a new auto-released byte array.
+    ///
+    /// See also [`get_byte_array_elements_auto`](struct.JNIEnv.html#method.get_byte_array_elements_auto)
+    pub fn auto_byte_array<'b, O>(
```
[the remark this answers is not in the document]
maurolacy: No, not really. They just need to be accessed from the same module / file. I'' fix it.

src/wrapper/jnienv.rs
maurolacy: Right, thanks.

src/wrapper/jnienv.rs
maurolacy: Good points. I think we should provide the API, and specifically point to the dangers / limitations of Critical functions. `std::mem::drop` can be used, as you mention, but also scopes. I'll mention both.

src/wrapper/objects/auto_byte_array.rs
maurolacy: Not really. It's there just because I copied it from `auto_local`. I'll happily remove it.

src/wrapper/objects/auto_byte_array.rs
maurolacy: Sure. That was a doubt I had, and now that I think of it, the more simple / generic the code the better.

maurolacy:
> What's interesting is that Get*Elements appeared to be no better than `Get*Region` for most sizes (though, on JDK 8, on 14 the picture might be different): https://habr.com/ru/post/222997/ The post is in Russian, but the diagram with the results is in English.

Yes. My own tests indicate the same. This is because `Get*Elements` is basically always copying the array. I'm in fact using `Get*Critical` to get an advantage in performance.

I think we should keep the wrapper, though, as `Get*Elements` could work better in the future. Or in different VMs / with different GCs / settings.

src/wrapper/jnienv.rs
dmitry-timofeev: Perfect, thanks!

src/wrapper/objects/auto_byte_array.rs
dmitry-timofeev: Makes sense, as with the 'auto' flavour of these methods it is not possible to get that information on creation.

maurolacy:
Guys, if you don't have more issues, for me this is ready to merge. I could still add the benchmarks, but I would prefer to do that as a separate task / PR, as I have some more pressing issues to work on right now.

src/wrapper/jnienv.rs
```diff
@@ -1932,6 +1967,202 @@ impl<'a> JNIEnv<'a> {
...
+    ///
+    /// Since the returned array may be a copy of the Java array, changes made to the
+    /// returned array will not necessarily be reflected in the original array until
+    /// release_primitive_array_critical() is called, which happens at AutoPrimitiveArray
+    /// destruction.
+    /// AutoPrimitiveArray has a commit() method, to force a copy of the array if needed (without
+    /// releasing it).
+    ///
+    /// See also [`get_primitive_array_critical`](struct.JNIEnv.html#method.get_primitive_array_critical)
+    pub fn get_primitive_array_critical_auto(
```
DarkEld3r, line 2154: I'm slightly worried about the naming inconsistency: we have `auto_local` and `get_primitive_array_critical_auto`.

Additionally, in Rust `get_` prefix is usually omitted, but I think that these arguments aren't very relevant in our case, because this library already has a lot functions with such prefix.

Personally, I would prefer to rename this function to `auto_primitive_array_critical` (or `get_auto_primitive_array_critical`) and use `AutoPrimitiveArray::new()` directly, so the `auto_primitive_array` function can be removed. Same for other `_auto` functions.

tests/jni_api.rs
```diff
@@ -13,6 +15,7 @@ use jni::{
 };
 
 mod util;
+use jni::objects::ReleaseMode::Copy;
```
DarkEld3r, line 18: A nitpick: I think that it is usually better to import an enum instead of its members, especially for such commonly used things as "copy".

tests/jni_api.rs
```diff
@@ -316,6 +319,194 @@ pub fn java_byte_array_from_slice() {
...
+
+    // Modify
+    vec[0] += 1;
+    vec[1] += 1;
+    vec[2] += 1;
+
+    // Release
+    // First, make sure vec's destructor doesn't free the data it thinks it owns when it goes out
+    // of scope (so release_primitive_array_critical() can properly free it)
+    std::mem::ManuallyDrop::new(vec);
```
DarkEld3r, line 449: I think that it is more idiomatic to use `ManuallyDrop` when creating a vector. Otherwise it is easier to use `mem::forget` here.

src/wrapper/objects/auto_byte_array.rs
```diff
@@ -0,0 +1,94 @@
...
+/// ReleaseMode
+///
+/// This defines the release mode of AutoByteArray (and AutoPrimitiveArray) resources, and
+/// related release array functions.
+#[derive(Clone, Copy)]
+pub enum ReleaseMode {
+    /// Copy back the content and free the elems buffer.
+    Copy = 0,
+    /// Free the buffer without copying back the possible changes.
+    NoCopy = JNI_ABORT as isize,
```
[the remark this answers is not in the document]
maurolacy: Yes, I can change the `enum` representation to avoid the cast:
```
#[derive(Clone, Copy)]
#[repr(i32)]
pub enum ReleaseMode {
    /// Copy back the content and free the elems buffer.
    Copy = 0,
    /// Free the buffer without copying back the possible changes.
    NoCopy = JNI_ABORT,
}
```
Will do.

src/wrapper/jnienv.rs
```diff
@@ -1932,6 +1967,202 @@ impl<'a> JNIEnv<'a> {
...
+    /// disable garbage collection when the native code is holding a pointer to an array obtained
+    /// via get_primitive_array_critical.
+    ///
+    /// Note that get_primitive_array_critical might still make a copy of the array if the VM
+    /// internally represents arrays in a different format. Therefore, we need to check its
+    /// return value against NULL for possible out of memory situations.
+    pub fn get_primitive_array_critical(
+        &self,
+        array: jarray,
+        is_copy: &mut jboolean,
```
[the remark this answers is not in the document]
maurolacy: A tuple / struct doesn't make much sense, as the function is returning only one thing. In fact, if you're OK with that I can return the raw pointer directly, instead of wrapping it into a `Result`. If I recall correctly, I only wrapped it because that's what I saw in other functions / examples.

src/wrapper/jnienv.rs
maurolacy: OK.

tests/jni_api.rs
maurolacy: Good point.

Another issue I have with this enum, is that I'm not sure if it's better to rename these members adding the `Back` suffix, i.e. `CopyBack` and `NoCopyBack`. That's to emphasize these modes apply at object's release, not creation.

tests/jni_api.rs
maurolacy: OK. Thanks.

src/wrapper/jnienv.rs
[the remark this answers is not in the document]
maurolacy: Update: Or are you're talking about removing the `is_copy` parameter entirely, and returning a tuple / struct with the pointer and the `is_copy` flag? I'm OK with that.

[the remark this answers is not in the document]
DarkEld3r: Yes, exactly. I think that "output parameters" are confusing, because in that case it is non-obvious if it is out or in-out parameter.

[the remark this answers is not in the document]
maurolacy: Changed.

As part of this I also added a non-null check to `get_primitive_array_critical`'s return value, which was missing.

Also changed / fixed the documentation.

Thanks for your review. The code is now cleaner, and more elegant.

maurolacy:
> Fantastic job, thanks a lot!

My pleasure. Thank you both for your insightful suggestions.

src/wrapper/objects/auto_byte_array.rs
```diff
@@ -0,0 +1,95 @@
...
+
+/// Auto-release wrapper for pointer-based byte arrays.
+///
+/// This wrapper is used to wrap pointers returned by get_byte_array_elements.
+///
+/// These arrays need to be released through a call to release_byte_array_elements.
+/// This wrapper provides automatic array release when it goes out of scope.
+pub struct AutoByteArray<'a: 'b, 'b> {
+    obj: JObject<'a>,
+    ptr: *mut jbyte,
```
DarkEld3r, line 27: I think that `NonNull` can be used here: calling `as_mut` wouldn't require unsafe and unwrap. Additionally, I think that `AutoByteArray::new` should return `Result` because `ptr` should be null.

src/wrapper/jnienv.rs
```diff
@@ -1932,6 +1933,209 @@ impl<'a> JNIEnv<'a> {
...
+    /// AutoByteArray has a commit() method, to force a copy of the array if needed (and without
+    /// releasing it).
+    ///
+    /// See also [`get_byte_array_elements`](struct.JNIEnv.html#method.get_byte_array_elements)
+    pub fn get_auto_byte_array_elements(
+        &self,
+        array: jbyteArray,
+        mode: ReleaseMode,
+    ) -> Result<AutoByteArray> {
+        let (ptr, is_copy) = self.get_byte_array_elements(array).unwrap();
```
DarkEld3r, line 2022:
```suggestion
        let (ptr, is_copy) = self.get_byte_array_elements(array)?;
```

src/wrapper/jnienv.rs
```diff
@@ -1932,6 +1933,209 @@ impl<'a> JNIEnv<'a> {
...
+    /// AutoPrimitiveArray also has a commit() method, to force a copy of the array if needed
+    /// (without releasing it).
+    ///
+    /// See also [`get_primitive_array_critical`](struct.JNIEnv.html#method.get_primitive_array_critical)
+    pub fn get_auto_primitive_array_critical(
+        &self,
+        array: jarray,
+        mode: ReleaseMode,
+    ) -> Result<AutoPrimitiveArray> {
+        let (ptr, is_copy) = self.get_primitive_array_critical(array).unwrap();
```
DarkEld3r, line 2135:
```suggestion
        let (ptr, is_copy) = self.get_primitive_array_critical(array)?;
```

src/wrapper/objects/auto_byte_array.rs
maurolacy: I think that instead of using `NonNull`, it's simpler to change the the `release` and `commit` prototypes to directly take a raw pointer instead of a ref. Please see my next commit.

Regarding `AutoByteArray::new` returning a `Result`, do you think it's really needed? (I guess you mean 'could' instead of 'should' above). `GetByteArrayElements` will never return null (because we're passing a valid java object to it). And if somebody uses `AutoByteArray` poorly, it's their problem.

But if you really think it's better / safer, I'll change it.

maurolacy: Update: No, it's not possible; `clippy` complains about raw pointer dereferences (which makes sense). OK, I'll go with the `NonNull`changes.

maurolacy: Update2: Even with `ptr` being a`NonNull<jbyte>`, converting to a mutable reference still requires an `unsafe` block. So (unless I'm missing something) it´s not possible to avoid the `unsafe` blocks. In fact this is consistent with what the [NonNull](https://doc.rust-lang.org/std/ptr/struct.NonNull.html) doc says.

I can commit these changes if you want, but I don't think they add much in terms of safety. And in fact, they add a little more clutter, so I would prefer not to add them.

DarkEld3r: You are right, I was sure that dereferencing `NonNull` was a safe operation, but it isn't. Though it wouldn't require unwrapping the return value, so personally I still think it is better. Am I missing something or are there any other difficulties except creating a `NonNull` wrapper?

As for `AutoByteArray::new` returning a `Result` - I see your point, but generally (rust) api should either be unsafe or validate its arguments. Panic instead of `Result` would also be OK, something like that (if we agree on using `NonNull`):
```rust
ptr: NonNull::new(ptr).expect("Non-null pointer expected")
```
Though I think that returning `Result`/`Option` makes api more flexible and we already have `NullPtr` error that can be used here.

maurolacy: OK, I've committed the `NonNull` changes. Basically, we're trading two `unwrap`s by a `new` (and an `&mut`).

I'll now add the `Result` wrapper as a separate commit.

maurolacy: Done.

These latest changes were more cosmetic than anything else in my opinion. But I agree, moving the `Result` wrapper "down the line" makes sense. And using `NonNull` adds some robustness (and helps with the consistency in the error reporting).

DarkEld3r:
`SIGSEGV` in the tests is disturbing, but it shouldn't be related to your changes.

maurolacy:
> `SIGSEGV` in the tests is disturbing, but it shouldn't be related to your changes.

Yeah, that's what I thought. I cannot reproduce it locally, by the way. And not for lack of trying ;-)

dmitry-timofeev:
Restarted the failing job and it passed. Merged into master. Thanks a lot, @maurolacy !

dmitry-timofeev:
@maurolacy , released 0.17.0 with your patch on crates.io 🎉

PR #240, opened 2020-07-02
Try github actions

stanislav-tkach:
## Overview

I have positive experience with Github actions and would like to try them with this project.

### Definition of Done

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,55 +3,55 @@ name: CI
 on: [pull_request, push]
 
 jobs:
-#  # Check code formatting.
-#  fmt:
-#    name: Rustfmt
-#    runs-on: ubuntu-latest
-#    steps:
-#      - name: Checkout sources
-#        uses: actions/checkout@v2
-#      - name: Install rust
-#        uses: actions-rs/toolchain@v1
-#        with:
-#          toolchain: stable
-#          components: rustfmt
-#          profile: minimal
-#          override: true
-#      - name: Run rustfmt
-#        uses: actions-rs/cargo@v1
-#        with:
-#          command: fmt
-#          args: --all -- --check
-#
-#  # Static analyzer.
-#  clippy:
-#    name: Clippy
-#    runs-on: ubuntu-latest
-#    steps:
-#      - name: Checkout sources
-#        uses: actions/checkout@v2
-#      - name: Install rust
-#        uses: actions-rs/toolchain@v1
-#        with:
-#          toolchain: stable
-#          components: clippy
-#          profile: minimal
-#          override: true
-#      - name: Run clippy
-#        uses: actions-rs/clippy-check@v1
-#        with:
-#          token: ${{ secrets.GITHUB_TOKEN }}
-#          args: -- -D warnings
-#
-#  # Security audit.
-#  audit:
-#    name: Security audit
-#    runs-on: ubuntu-latest
-#    steps:
-#      - uses: actions/checkout@v2
-#      - uses: actions-rs/audit-check@v1
-#        with:
-#          token: ${{ secrets.GITHUB_TOKEN }}
+  # Check code formatting.
+  fmt:
+    name: Rustfmt
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v2
+      - name: Install rust
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          components: rustfmt
+          profile: minimal
+          override: true
+      - name: Run rustfmt
+        uses: actions-rs/cargo@v1
+        with:
+          command: fmt
+          args: --all -- --check
+
+  # Static analyzer.
+  clippy:
+    name: Clippy
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v2
+      - name: Install rust
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          components: clippy
+          profile: minimal
+          override: true
+      - name: Run clippy
+        uses: actions-rs/clippy-check@v1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          args: -- -D warnings
+
+  # Security audit.
+  audit:
+    name: Security audit
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions-rs/audit-check@v1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
 
   # Tests.
   test:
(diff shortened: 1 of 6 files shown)


DarkEld3r:
The results can be seen here: https://github.com/DarkEld3r/jni-rs/actions/runs/155858041

.github/workflows/ci.yml
```diff
@@ -0,0 +1,122 @@
...
+          token: ${{ secrets.GITHUB_TOKEN }}
+          args: -- -D warnings
+
+  audit:
+    name: Security audit
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions-rs/audit-check@v1
```
dmitry-timofeev, line 56: Interesting that this step takes about 6 minutes :fearful: 

https://github.com/DarkEld3r/jni-rs/runs/832278844?check_suite_focus=true

.github/workflows/ci.yml
```diff
@@ -0,0 +1,122 @@
...
+    strategy:
+      matrix:
+        build: [Linux, macOS, Win32-msvc, Win32-gnu, Win64-msvc, Win64-gnu]
+
+        include:
+          - build: Linux
+            os: ubuntu-latest
+          - build: Linux-nightly
+            os: ubuntu-latest
+            rust: nightly
```
dmitry-timofeev, line 73: Is there anything like "allow_failures" in Travis?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,122 @@
...
+          args: --examples --all
+
+      - name: Test
+        uses: actions-rs/cargo@v1
+        with:
+          command: test
+          args: --all
+
+      - name: Test invocation
+        if: matrix.rust == 'nightly'
```
dmitry-timofeev, line 118: I think it must be the other way around: we test invocation feature on all platforms, but may skip on nightly (though it doesn't seem to make much sense).

Also, it requires setting the `JAVA_HOME` env variable, hence the job currently fails: https://github.com/DarkEld3r/jni-rs/runs/832279070?check_suite_focus=true#step:6:18

.github/workflows/ci.yml
DarkEld3r: There is `continue-on-error` option, but unfortunately it isn't supported at the job level. Here is an example of a workaround: https://github.com/ozkriff/zemeroth/blob/master/.github/workflows/ci.yml#L85

.github/workflows/ci.yml
DarkEld3r: Yes, for some reason it is the longest check.

.github/workflows/ci.yml
DarkEld3r: https://github.com/actions/setup-java should install Java and set `JAVA_HOME`. I'm going to look at this closely.

stanislav-tkach:
- Java is installed using [this](https://github.com/actions/setup-java) github action.
- On my projects github actions are faster than Travis (and much faster than Appveyor). As far as I can see, the second check is much faster, but it makes sense to compare timing when I am able to fix everything. :upside_down_face: 
- Yes, it is possible to use shell scripts.

stanislav-tkach:
Currently blocked by https://github.com/actions/setup-java/issues/99.

stanislav-tkach:
Finally I managed to get GA working: https://github.com/stanislav-tkach/jni-rs/actions

.github/workflows/ci.yml
```diff
@@ -3,119 +3,107 @@ name: CI
...
             os: macOS-latest
-          - build: Win32-msvc
-            os: windows-latest
-            rust: stable-i686-pc-windows-msvc
-            target: i686-pc-windows-msvc
-          - build: Win32-gnu
-            os: windows-latest
-            rust: stable-i686-pc-windows-gnu
-            target: i686-pc-windows-gnu
+            libjvm: /Users/runner/hostedtoolcache/jdk/8.0.265/x64/jre/lib/jli/
```
dmitry-timofeev, line 68: I see there is an exact version hardcoded. If a JDK is updated (?in the image), will it break our actions? Can we detect the exact path using our `tests_profile` script?

.github/workflows/ci.yml
```diff
@@ -3,119 +3,107 @@ name: CI
...
           target: ${{ matrix.target }}
           profile: minimal
           override: true
-
+      - name: Install java
+        uses: actions/setup-java@v1
+        with:
+          java-version: '1.8.0'
+      - name: Add libjvm to LD_LIBRARY_PATH
+        run: echo "::set-env name=LD_LIBRARY_PATH::${{ matrix.libjvm }}"
```
dmitry-timofeev, line 94: Reminds me of this tweet: https://twitter.com/pchiusano/status/1300243277652856834 :upside_down_face:

.github/workflows/ci.yml
stanislav-tkach: I don't like the hard-coded path either. :upside_down_face: 

I guess the script can be used, but unfortunately we need a different script for Windows. I will try that approach.

.github/workflows/ci.yml
stanislav-tkach: It now takes less than a minute. I guess Github does some caching, so initial checks run longer.

stanislav-tkach:
Finally all checks are green: https://github.com/stanislav-tkach/jni-rs/actions/runs/290080253
I have managed to use the `test_profile` script and added the similar script for Windows.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,103 @@
...
+        with:
+          toolchain: stable
+          components: clippy
+          profile: minimal
+          override: true
+      - name: Run clippy
+        uses: actions-rs/clippy-check@v1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          args: -- -D warnings
```
dmitry-timofeev, line 44: There seems to be a difference in arguments: 
cargo clippy --all --tests --all-features -- -D warnings

.github/workflows/ci.yml
```diff
@@ -0,0 +1,103 @@
...
+        include:
+          - build: Linux
+            os: ubuntu-latest
+          - build: macOS
+            os: macOS-latest
+          - build: Win64-msvc
+            os: windows-latest
+            rust: stable-x86_64-pc-windows-msvc
+            target: x86_64-pc-windows-msvc
+          - build: Win64-gnu
```
dmitry-timofeev, line 71: Do we need both windows targets?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,103 @@
...
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+  # Tests.
+  test:
+    name: ${{ matrix.build }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        include:
```
dmitry-timofeev, line 62: We no longer test on nightly with "allow_failures"?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,103 @@
...
+      - name: Install java
+        uses: actions/setup-java@v1
+        with:
+          java-version: '1.8.0'
+      - name: Build
+        uses: actions-rs/cargo@v1
+        with:
+          command: build
+          args: --examples --all
+      - name: Test
```
dmitry-timofeev, line 94: Test default features?

dmitry-timofeev (commented):
Generally, looks good. Do these actions for clippy, etc complete? Can we do pass everything we can as CLI flags using these actions? Or do they always lag behind the corresponding commands?

.github/workflows/ci.yml
stanislav-tkach: I'm not sure. On the one hand, we don't have windows-specific code that should be thoroughly checked. On the other hand, checks for each target take about three minutes, so I think it is not a big deal.

.github/workflows/ci.yml
stanislav-tkach: Github actions have the `continue-on-error` option, but it completely ignores all errors - the build will be "successful" and the only way to see actual errors is to check the logs. I'm not sure if it is useful in this state.

stanislav-tkach:
> Can we do pass everything we can as CLI flags using these actions? 

Sure. Additional arguments can be specified (using `args`) and are passed to the underlying tool.

.github/workflows/ci.yml
dmitry-timofeev: I see, let's skip nightly then.

dmitry-timofeev (approved):
LGTM, thanks! 

Do you plan to update badges/instructions/remove travis config separately?

Shellcheck is also gone :)

.github/workflows/ci.yml
dmitry-timofeev: I'd consider removing one then, so that there is less stuff.

stanislav-tkach:
Yes, I planned to deal with Travis and badges in a separate pull request.

As for `shellcheck`: I have moved it into a separate CI step, because we now use the same script for Linux and Macos and the latter doesn't have the `shellcheck` tool installed.

dmitry-timofeev (approved):

PR #244, opened 2020-08-03
Add method to define a class using an AutoByteArray

x4e:
## Overview

1. Adds `len` method to `AutoByteArray`
2. Adds `define_class_bytearr` method that takes an `AutoByteArray` rather than a `&[u8]`

I was struggling working out how to feed defineClass a `jbyteArray` and this seemed like the best way to do it.
Im new to rust, so please let me know if this is a dumb change/there is a better way to do it.

### Definition of Done

- [*] There are no TODOs left in the code
- [*] Change is covered by automated [tests](https://github.com/jni-rs/jni-rs/blob/master/CONTRIBUTING.md#tests)
- [*] The [coding guidelines](https://github.com/jni-rs/jni-rs/blob/master/CONTRIBUTING.md#the-code-style) are followed
- [*] Public API has documentation
- [*] This change is not breaking **or** mentioned in the Changelog

### DIFF
--- a/src/wrapper/jnienv.rs
+++ b/src/wrapper/jnienv.rs
@@ -143,6 +143,29 @@ impl<'a> JNIEnv<'a> {
         Ok(unsafe { JClass::from_raw(class) })
     }
 
+    /// Load a class from a buffer of raw class data. The name of the class must match the name
+    /// encoded within the class file data.
+    pub fn define_class_bytearray<S>(
+        &self,
+        name: S,
+        loader: JObject<'a>,
+        buf: AutoArray<'_, jbyte>,
+    ) -> Result<JClass<'a>>
+    where
+        S: Into<JNIString>,
+    {
+        let name = name.into();
+        let class = jni_non_null_call!(
+            self.internal,
+            DefineClass,
+            name.as_ptr(),
+            loader.into_raw(),
+            buf.as_ptr(),
+            buf.size()?
+        );
+        Ok(unsafe { JClass::from_raw(class) })
+    }
+
     /// Look up a class by name.
     ///
     /// # Example
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - `JavaStr::from_raw()` which takes ownership of a raw string pointer to create a `JavaStr` ([#374](https://github.com/jni-rs/jni-rs/pull/374))
 - `JNIEnv::get_string_unchecked` is a cheaper, `unsafe` alternative to `get_string` that doesn't check the given object is a `java.lang.String` instance. ([#328](https://github.com/jni-rs/jni-rs/issues/328))
 - `WeakRef` and `JNIEnv#new_weak_ref`. ([#304](https://github.com/jni-rs/jni-rs/pull/304))
+- `define_class_bytearray` method that takes an `AutoArray<jbyte>` rather than a `&[u8]`
 
 ### Changed
 - `JNIEnv::get_string` checks that the given object is a `java.lang.String` instance to avoid undefined behaviour from the JNI implementation potentially aborting the program. ([#328](https://github.com/jni-rs/jni-rs/issues/328))


cookiedragon234:
Should I modifiy it to pass the code formatting check? I was using the formatting in the existing code as a reference...

maurolacy:
Hello,

I don't know much about `DefineClass`, but this looks like a sensible use for an `AutoByteArray`.

In particular, the (admittedly scarce) documentation for [DefineClass](https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/functions.html#DefineClass) says: 
_The buffer containing the raw class data is not referenced by the VM after the DefineClass call returns, and it may be discarded if desired._

So, this could be a convenient way to automatically release the buffer.

maurolacy:
> Should I modifiy it to pass the code formatting check? I was using the formatting in the existing code as a reference...

Yes, you should. Take a look at running `cargo fmt` and `cargo clippy` locally.

src/wrapper/objects/auto_byte_array.rs
```diff
@@ -73,6 +73,11 @@ impl<'a, 'b> AutoByteArray<'a, 'b> {
     pub fn is_copy(&self) -> bool {
         self.is_copy
     }
+	
+	/// Returns the size of this byte array as a jsize
+	pub fn len(&self) -> jsize {
+		self.env.get_array_length(*self.obj).expect("Couldn't fetch byte array length")
+	}
```
maurolacy, line 80: I don't think this method is a good idea, as it's relatively inefficient, and it's for obtaining data that you probably have already

Better pass the `len` as a parameter in the main (`define_class_bytearr`) method.

src/wrapper/jnienv.rs
```diff
@@ -117,6 +117,23 @@ impl<'a> JNIEnv<'a> {
         );
         Ok(class)
     }
+	
+	/// Define a new java class using an AutoByteArray
+	pub fn define_class_bytearr<S>(&self, name: S, loader: JObject<'a>, buf: AutoByteArray) -> Result<JClass<'a>>
```
maurolacy, line 122: Better naming here? Not sure, maybe something like `define_class_autobytearray`?

cookiedragon234:
Squashed commits and tests should pass now

src/wrapper/jnienv.rs
```diff
@@ -118,6 +118,29 @@ impl<'a> JNIEnv<'a> {
         Ok(class)
     }
 
+    /// Define a new java class using an AutoByteArray
+    pub fn define_class_autobytearray<S>(
+        &self,
+        name: S,
+        loader: JObject<'a>,
+        buf: AutoByteArray,
+        buf_length: jsize,
```
maurolacy, line 127: I think this should be an `usize` for generality, and you cast it to `jsize` (`buf_length as jsize`) below.

This looks good to me in general. But, I'm not an official reviewer. You'll have to convince them, no me. :-)

cookiedragon234: Really? jsize is the official parameter in the documentation

maurolacy: Let's wait for the feedback of the reviewers.

DarkEld3r: I think that using `jsize` is fine: as far as I can see, all `JNIEnv` methods use this type. Additionally, we are still limited by `jsize` in the call to `DefineClass` function.

cookiedragon234:
@DarkEld3r good?

DarkEld3r:
@cookiedragon234 Yes, thanks!

src/wrapper/jnienv.rs
rib: Having the separate `buf_length` argument seems to make this a potentially `unsafe` API compared to `define_class` because it would be easy to trigger a buffer overrun by telling `DefineClass` that the buffer is larger than it really is.

rib:
Hi @x4e 

Sorry that this PR has languished for quite some time here.

I did try to review this in passing while I was looking at what to merge for 0.20 since I recently started helping with maintaining the project.

My main observations are that:
- The API this is based on got re-worked / renamed to `AutoArray<jbyte>` so it unfortunately needs a bit of extra work to rebase.
- Having the separate `buf_length` argument means that `define_class_autobytearray` would need to be marked as `unsafe` as it is currently because it would be possible to easily overrun the length of the buffer by passing in a `buf_length` that's too large.
- To make this safe it could either check the `buf_length` against `buf.size()` or else it could drop the `buf_length` argument and instead automatically refer to `buf.size()` internally.

One other possibility is that the existing `define_class` API could also be used with a bit of `unsafe` code that would create a temporary slice from an `AutoArray<jbyte>` something like:
```rust
let slice = unsafe { slice::from_raw_parts(byte_array.as_ptr() as *const u8, byte_array.size() as usize) };
env.define_class(env, loader, slice);
```
but considering that's not very ergonomic then it would probably still be nicer to have an API like `define_class_autobytearray`.

rib:
Okey, I haven't tested this but I've done the following:

- rebased on master
- renamed the api from `define_class_autobytearray` to `define_class_bytearray` (consistent with changelog)
- the api takes an `AutoArray<jbyte>` now
- removed the `buf_length` argument in favour of referring to `buf.size()` internally

rib:
Even though it's been long enough that the original author likely no longer recalls the details surrounding this contribution, it seems like a fairly non-controversial, orthogonal, small addition so I figure this is reasonable to land, including the fixups I made above.

PR #247, opened 2020-08-11
Add 'define_unnamed_class' function

### LINKED ISSUE #245: `define_class` with null name
At the moment it is impossible to provide a null name to `env.define_class`, as is allowed in JNI, which means the class name is found from the contents of the provided class bytes.

A potential solution could be to change the constraint on define class from:
```Rust
where
    S: Into<JNIString>,
```
To:
```Rust
where
	S: Into<*const c_char>
```

Would this change be acceptable? JNIString could be modified to implement `Into<*const c_char>`.

This would allow code like the following:
```Rust
define_class(null(), ...);
define_class(CString::new("ClassName").unwrap().as_ptr(), ...);
```

stanislav-tkach:
## Overview

Closes #245.

### Definition of Done

### DIFF
--- a/src/wrapper/jnienv.rs
+++ b/src/wrapper/jnienv.rs
@@ -110,7 +110,8 @@ impl<'a> JNIEnv<'a> {
         self.define_class_impl(name.as_ptr(), loader, buf)
     }
 
-    /// Load a class from a buffer of raw class data.
+    /// Load a class from a buffer of raw class data. The name of the class is inferred from the
+    /// buffer.
     pub fn define_unnamed_class<S>(&self, loader: JObject<'a>, buf: &[u8]) -> Result<JClass<'a>>
     where
         S: Into<JNIString>,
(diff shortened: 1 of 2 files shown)


cookiedragon234:
Thank you!

src/wrapper/jnienv.rs
```diff
@@ -100,17 +100,34 @@ impl<'a> JNIEnv<'a> {
...
+    /// encoded within the class file data.
     pub fn define_class<S>(&self, name: S, loader: JObject<'a>, buf: &[u8]) -> Result<JClass<'a>>
     where
         S: Into<JNIString>,
     {
         let name = name.into();
+        self.define_class_impl(name.as_ptr(), loader, buf)
+    }
+
+    /// Load a class from a buffer of raw class data.
```
dmitry-timofeev, line 113: Shall we add that the VM _infers_ the class name from the buffer, as the issue suggests? As "unnamed class" might sound confusing.

dmitry-timofeev (approved):
:+1:

src/wrapper/jnienv.rs
stanislav-tkach: I have updated the documentation, but I agree with you that `define_unnamed_class` isn't the best name for the method. I suppose it would be better to have `define_class` and `define_named_class` functions, but it would be a breaking change.

dmitry-timofeev (approved):

PR #249, opened 2020-09-04
Add set_static_field method

### LINKED ISSUE #226: Expose SetStatic*Field APIs
It looks like this crate exposes the `GetStatic*Field` APIs:
https://docs.rs/jni/0.14.0/jni/struct.JNIEnv.html#method.get_static_field

but not the corresponding `SetStatic*Field` APIs:
https://docs.oracle.com/javase/7/docs/technotes/guides/jni/spec/functions.html#wp20829

It would be nice to support those as well.

stanislav-tkach:
## Overview

Closes #226.

### Definition of Done

### DIFF
--- a/src/wrapper/jnienv.rs
+++ b/src/wrapper/jnienv.rs
@@ -1693,51 +1693,47 @@ impl<'a> JNIEnv<'a> {
         T: Desc<'a, JClass<'c>>,
         U: Desc<'a, JStaticFieldID<'f>>,
     {
+        use JavaType::Primitive as JP;
+
         let class = class.lookup(self)?.into_inner();
+        let field = field.lookup(self)?.into_inner();
 
-        let field_id = field.lookup(self)?.into_inner();
-
-        // TODO clean this up
-        Ok(match ty {
+        let result = match ty {
             JavaType::Object(_) | JavaType::Array(_) => {
-                let obj: JObject =
-                    jni_non_void_call!(self.internal, GetStaticObjectField, class, field_id).into();
-                obj.into()
+                jni_non_void_call!(self.internal, GetStaticObjectField, class, field).into()
             }
-            // JavaType::Object
             JavaType::Method(_) => {
                 return Err(ErrorKind::WrongJValueType("Method", "see java field").into())
             }
-            JavaType::Primitive(p) => match p {
-                Primitive::Boolean => {
-                    jni_unchecked!(self.internal, GetStaticBooleanField, class, field_id).into()
-                }
-                Primitive::Char => {
-                    jni_unchecked!(self.internal, GetStaticCharField, class, field_id).into()
-                }
-                Primitive::Short => {
-                    jni_unchecked!(self.internal, GetStaticShortField, class, field_id).into()
-                }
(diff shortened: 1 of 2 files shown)


src/wrapper/jnienv.rs
```diff
@@ -1693,51 +1693,47 @@ impl<'a> JNIEnv<'a> {
...
+                jni_unchecked!(self.internal, GetStaticDoubleField, class, field).into()
+            }
+            JP(Primitive::Byte) => {
+                jni_unchecked!(self.internal, GetStaticByteField, class, field).into()
+            }
+            JP(Primitive::Void) => {
+                return Err(ErrorKind::WrongJValueType("void", "see java field").into())
+            }
+        };
+        return Ok(result);
```
maurolacy, line 1736: `Ok(result)` would work as well here.

src/wrapper/jnienv.rs
```diff
@@ -1757,6 +1753,42 @@ impl<'a> JNIEnv<'a> {
         self.get_static_field_unchecked(class, (class, field, sig), ty)
     }
 
+    /// Set a static field. Requires a class lookup and a field id lookup internally.
+    pub fn set_static_field<'c, 'f, T, U>(&self, class: T, field: U, value: JValue) -> Result<()>
+    where
+        T: Desc<'a, JClass<'c>>,
+        U: Desc<'a, JStaticFieldID<'f>>,
+        //U: Into<JNIString>,
```
maurolacy, line 1761: \<nit\> Remove commented code.

src/wrapper/jnienv.rs
```diff
@@ -1757,6 +1753,42 @@ impl<'a> JNIEnv<'a> {
...
+    {
+        let class = class.lookup(self)?.into_inner();
+        let field = field.lookup(self)?.into_inner();
+
+        match value {
+            JValue::Object(v) => jni_unchecked!(
```
maurolacy, line 1767: What happened to `JavaType::Array()`?

stanislav-tkach: As far as I can see, there is no way to do something like `SetStaticArrayField`. I'm not a Java expert, so I can be missing something.

stanislav-tkach:
@maurolacy Thanks for the review! By the way, I think that pointing the the commented code isn't a nitpick. :+1:

src/wrapper/jnienv.rs
dmitry-timofeev: That's correct: arrays are object references in Java, and are represented in JValue enum as JValue::Object (there is no JValue::Array). The C JNI API accepts them as an argument to `SetStaticObjectField`

dmitry-timofeev (approved):
Looks good. Regarding testing: I do not happen to know a class in the standard libary wit h a public static non-final field that we can set. The JNI methods, however, allow getting field IDs of private fields, but I do not think it's good to make a test depending on a private field in a std. library class. On top of that, VMs may be free to forbid such access to platform classes, but I haven't checked that.

Ideally, we could have a separate Java module with some classes for testing, but we are not there yet.

Issue #222, opened 2020-02-11
Documentation project: JNI Workshop


johnwalker:
I'm working on materials for a workshop that teaches JNI programming to Rust programmers, which uses jni-rs. Would you be interested in a project like that as a contribution to jni-rs?

dmitry-timofeev:
Hi, @johnwalker , 
I think it's an awesome initiative. Including it in the project will extend the audience to all (prospective) users; and will make it more likely to stay up-to-date in case the APIs evolve.

When I studied the JNI, I found the [tutorial on IBM Developer Works](https://developer.ibm.com/tutorials/j-jni/) to be quite well-written. Although it is rather dated and does not cover additions post 1.2, it might work as a source of ideas on what to cover.

johnwalker:
Awesome! Yeah, the one you referenced is nice. JNI is hard, so I'd love to get this stuff reviewed and representative of the best practices.

I have the start of some sections written, and can probably start raising PRs soon. Is there a place you think makes sense to add them? It's all mdbook material.

dmitry-timofeev:
Awesome!

I'd certainly keep the tutorial code in the main repo in a separate module, so that in case of any changes to the API the tutorial code is also updated. Regarding the tutorial pages — what's the best practice for that? Would it make sense to create a separate repository for doc pages or keep them here?

--- 

On native resource management (_Cleaning up resource leaks, using Cleaner / Try-with-resources_), we've done some research when started our project, and discovered that GC-based approaches (i.e., `finalize`, `PhantomReference` and `Cleaner`) are quite tricky to get completely right. I've summarized the main nuances in [this post](https://stackoverflow.com/a/44799935), with a link to a more detailed presentation. It may be useful if you plan to cover GC-based approaches and show a complete and cross-platform solution.

johnwalker:
I see a nice example here, which uses a single repo: https://github.com/camshaft/bolero

Let's do something similar to that! I like your idea of keeping the examples building together with the docs.

I've been very lucky to rely on https://github.com/corretto/amazon-corretto-crypto-provider/blob/develop/src/com/amazon/corretto/crypto/provider/Janitor.java for cleaning up native objects. It sounds like you've come up with a similar approach, and we should definitely spend the time to get that section right.

dmitry-timofeev:
> Let's do something similar to that!

Agree!

> It sounds like you've come up with a similar approach

[...]

Issue #84, opened 2018-03-01
Consider a registry of native peers


dmitry-timofeev:
## Overview
This is a _feature request_ for an additional safety net for the use-case when Java passes pointers to native objects to the native code.

Consider the following code:
```java
// An example of Java proxy of a native object: `vec` proxy
class VecProxy {
    // A handle to the corresponding native object (aka native peer).
    private final long nativeHandle;

    // Constructors, public methods using the native methods below:
    
    // Creates a new `vec` and returns a pointer to it.
    private static native long nativeCreate(int initialSize);

    // Adds the given element to the vector.
    private static native void nativeAdd(long nativeHandle, int newElement);

    // Destroys the corresponding native peer.
    private static native void nativeDestroy(long nativeHandle);
}
```

The native code has to cast `jlong`s to the corresponding objects. If there is a bug in Java code (use-after-free, double-free, unknown-pointer, wrong-pointer-type), the app is likely to crash.

## Possible solution
A registry of all native objects that can be used at the given moment by a pointer. Such an object checks that a native handle:
  - is known,
  - has a correct type.

Such checks may be optional (e.g., enabled in debug builds only), and may be exposed through such interface:
  - Create a handle to a native object, optionally, registering it in the registry: `to_handle`
  - Cast a handle to a native object of the given type (with _optional_ checks above): `cast_handle`
  - Drop handle, destroying the native object and, optionally, removing the handle from the registry: `drop_handle`

In our code base, we currently have a synchronized HashMap-based implementation.

## Questions
- [ ] Q0: Is it useful to all users? Shall it be:
    - In the core library?
    - In a separate satellite crate, e.g., `jni-java-proxies`?
    - Outside the lib?

dmitry-timofeev:
> Did you consider, as a possible solution, the usage of generation-based (also known as versioned) pointers? This seems like a possible (and possibly efficient) solution to the issue.

No, have not had a change to use them. Are they supposed to be significantly more efficient than, say, a proper concurrent hash map, allowing both concurrent reads *and* writes?

Our [present code](https://github.com/exonum/exonum-java-binding/blob/v0.3/exonum-java-binding-core/rust/src/utils/resource_manager/imp.rs#L22) uses a simple mutex-protected HashMap, but we consider using a concurrent hash map. However, this feature is mostly oriented at developers working on the project to catch bugs during development early, so performance is not an issue.

koutheir:
[...]

Issue #81, opened 2018-02-27
Auto-generate native Java-interfacing files


dmitry-timofeev:
## Overview
This is a _feature request_ and an another take on #80 . 

Consider adding a tool that takes a Java source file with native methods or a native C header of such file, produced by `javac`, and turns it into a Rust-template. Basically:
```
X.java --> (X.rs, X-human-readable.rs)
```

Such a tool may produce a single template file, similar to what one currently writes by hand, or a couple of files:
* An _intermediate_, always re-generated, low-level .rs file, that has the methods with signatures as required by C-JNI. Such file might have some boilerplate that is required to guarantee safe operations (catches panics, converts errors into Java exceptions, etc., see the primary issue #80)
* Optionally, a _user-facing_ template file. This one is generated once, has human-readable identifiers, possibly, exactly matching the names of Java methods (e.g., `nativeCreate` instead of `Java_com_example_company_NativePeerProxy_nativeCreate`). Such methods might also enjoy some conversions that are performed by the code in _intermediate_ file (e.g., `jint -> i32`, again, see #80).

If there is a _user-facing_ file, _intermediate_ one calls its functions.

## Requirements
1. Standard-compliance or reasonable restrictions. I.e., such a tool must be able to handle whatever `javac -h` produces. Restrictions may apply on method names (e.g., method names that contain only ASCII symbols).
2. Ease of use: using such a tool must be simpler than copy-pasting `javac -h` output.
3. Good UX:
    - _Reasonable defaults_, works out-of-the box. 
    - _Descriptive_ error messages when things go wrong at FFI boundary, since FFI is hard.
    - _User-facing_ template that is IDE-friendly (macros hidden in _intermediate_ file).
    - _Refactoring-friendly_: must work reasonably when native methods are added/renamed/have their signatures modified.
    - _Build integration_: whatever build-system you are using, the tool must integrate nicely (e.g., Gradle/Maven/Ant/Bazel/Cargo).
4. Low or moderate overhead: ideally, you get extra safety for no extra performance or complexity cost.
5. Works on popular OSes.

## Criteria
1. Low or moderate complexity: no knowledge of :rocket: science required to implement and maintain.
2. Maintainable: when this library updates, it's easy and well-known how to modify the tool.
3. Permissive license.

## Possible approaches
1. GNU-fu: Awk & friends, taking a `javac -h` output.
    - Don't have to reinvent the wheel to process text files.
    - Need to un-mangle method names to produce good user-facing file.
[...]

PR #251, opened 2020-09-11
Add draft tutorial for jni programming

johnwalker:
Hi there, this is a followup draft from #222 . I'd like to work with you on the content here. A few others contributed to these materials as well. :)

## Overview

Add a tutorial that introduces JNI programming with jni-rs. Build with mdbook to get a preview. One of the goals of the book is to have the reader do stuff and think about JNI/FFI in exercises. At the moment, it introduces JNI with division, debugging, and a native Counter implementation. You will see TODOs as you reach the end of the tutorial, especially on Cleaners.

### Definition of Done

- [ ] Everyone is happy enough with writing style and content
- [x] The material is tested on Windows
- [x] The material is tested on Linux
- [x] The material is tested on OSX
- [ ] The authors tag is filled in


book/src/linking_and_dividing.md
```diff
@@ -0,0 +1,337 @@
...
+code that accompanies this book. It uses `System.loadLibrary` to locate the
+`jnibookrs` lib, and specifies native methods using the `native` keyword. The
+native method called `verify_link` was used earlier to ensure that the project
+was setup correctly, with environment variables.
+
+For style reasons, we always add native methods to `NativeAPI`. It doesn't need
+to be that way, it's just a convention that we follow here.
+
+```java
+package com.github.jni_rs.jnibook;
```
dmitry-timofeev, line 19: Is it possible to include code from files?

book/src/linking_and_dividing.md
```diff
@@ -0,0 +1,337 @@
...
+   private static final Throwable INIT_ERROR;
+
+   // The static block will be executed the first time the NativeAPI
+   // class is used.
+   static {
+       Throwable error = null;
+       try {
+           System.loadLibrary("jnibookrs");
+       } catch (Throwable t) {
+           error = t;
```
dmitry-timofeev, line 32: (here and on the full file)

Why don't we throw an Exception, with, potentially, a clearer error message,
on the first use? It seems to be simpler, fails immediately, and there is no need to repeatedly checkAvailability which never changes after the static initializer has executed: the library was either loaded or not.

For better error reporting (non tutorial purposes), it might make sense to extract that code, common across several classes with native methods, in a separate class. E.g., that's what Exonum has (it has to be clear on what's wrong as a framework): https://github.com/exonum/exonum-java-binding/blob/ed61d3454469530575c264074ed9ed06c0300039/exonum-java-binding/core/src/main/java/com/exonum/binding/core/util/LibraryLoader.java#L115

book/src/linking_and_dividing.md
```diff
@@ -0,0 +1,337 @@
...
+## Function Linking Rules
+
+JNI identifies the functions it should link to using the package path, class
+name, and method name by following [these
+rules](https://docs.oracle.com/en/java/javase/11/docs/specs/jni/design.html).
+Since `.` isn’t allowed in C function names, the Java function's package path's
+`.` is mapped to `_`. However, since `_` could appear in a function name, it
+needs a different representation. So `_` gets replaced by `_1` before replacing
+`.` with `_`. And finally, the function name has to be prefixed with `Java_`, to
+avoid conflicts with other languages that support similar naming schemes.
```
dmitry-timofeev, line 55: Wouldn't it be better to suggest that users run `javac -h <path-to-headers>`
and copy the function declarations from the generated header file(s)?

E.g.,
```
javac -h . NativeAPI.java
cat NativeAPI.h

# Or with Gradle
./gradlew build
cat target/<our-path>/NativeAPI.h
```

It doesn't have to be configured separately from usual javac invocation,
see for example, https://github.com/exonum/exonum-java-binding/blob/master/exonum-java-binding/core/pom.xml#L184). It can be put into our Gradle template, so that in `target/native-headers`
there are up-to-date headers with correct function names.

javac reliably produces correct method names, that a user can copy and paste.
Even though the transformations below are described really well, and the examples
illustrate them well, if someone tries to follow them IRL, it may be tedious:
  - Copy/type the package name
  - Separately copy/type method name
  - Mangle as specified.

I think the rules distract from the more interesting parts; and may be harmful:
it is possible to mis-understand them, or make a typo or an error during a manual
conversion, or miss "Java_", and end up debugging "LinkageError: symbol not found".

Surely, it would be lovely to not have to do that at all: https://github.com/jni-rs/jni-rs/issues/81

book/src/linking_and_dividing.md
```diff
@@ -0,0 +1,337 @@
...
+
+```rust
+fn Java_com_github_imaginarypackagename_LookingGlass_test_1call() {
+
+}
+```
+
+Now the native function is named correctly for JNI.
+
+## Parameters and Return Types
```
dmitry-timofeev, line 110: Would it be clearer if first a 1st argument (`JNIEnv`) is desribed; 
then — the 2nd argument:
- for native static methods (a reference to the defining class);
- for native instance methods (a reference to `this`)?

book/src/linking_and_dividing.md
```diff
@@ -0,0 +1,337 @@
...
+Mangling is last, and much less work.
+
+## Specifying the ABI and Disabling Mangling
+ABIs (Application Binary Interfaces) standardize low-level details, so that
+object code built using different compilers may still rely upon one another. In
+Rust, functions default to using the "Rust" ABI, which is usually desirable, but
+not for shared libraries.
+
+Therefore, it's necessary to explicitly set the ABI using `pub extern "system"`.
+For further information, see https://doc.rust-lang.org/std/keyword.extern.html
```
dmitry-timofeev, line 210: My mdbook renders them as plain text :confused:

book/src/linking_and_dividing.md
```diff
@@ -0,0 +1,337 @@
...
+pub fn Java_com_github_jni_1rs_jnibook_NativeAPI_divide(
+    _env: JNIEnv,
+    _class: JClass,
+    jint a, 
+    jint b) -> jint {
+    a/b
+}
+```
+
+Finally, set the ABI using `pub extern “system”`.
```
dmitry-timofeev, line 314: Fancy quotes in code

book/src/linking_and_dividing.md
```diff
@@ -0,0 +1,337 @@
...
+```
+
+Now that you're probably done, write a unit test on the Java side to verify that
+division works as expected.
+
+# Tip: JNIEnv Lifetimes
+
+`JNIEnv` instances share the lifetime of the calling thread, which means you
+should avoid holding onto `JNIEnv` instances. Once the calling thread is
+deallocated, it the `JNIEnv` instance will no longer be valid. [JNI00]
```
dmitry-timofeev, line 335: probably, redundant "it"

dmitry-timofeev (commented):
I've reviewed only the first two chapters, but brilliant job! Some comments below.

book/src/introduction.md
```diff
@@ -0,0 +1,169 @@
+# Introduction
+This book is about building shared libraries in Rust that work with Java via JNI
+(Java Native Interface). JNI is a Java API for using shared libraries from Java.
```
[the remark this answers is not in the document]
johnwalker: Thanks, I like that description much more. :)

book/src/linking_and_dividing.md
johnwalker: Yes it would. The writing there is kind of odd, in that it talks about argument 2, then 1, then 2. Will fix.

book/src/linking_and_dividing.md
johnwalker: Ah, got it. I think I fixed similar problems through the rest of the tutorial.

book/src/introduction.md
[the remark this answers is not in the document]
johnwalker: I borrowed the wording you suggested and rewrote the intro to talk about applications and libraries

book/src/introduction.md
```diff
@@ -0,0 +1,169 @@
...
+using JNI in certain ways. For example, Java Strings and Buffers sometimes have
+to be copied when they're used in the shared library. Due to the drawbacks of
+JNI, it's usually more appropriate to build using pure Java.
+
+There are plenty of reasons why you should consider Rust if you need to build a
+shared library for Java. Most notably, you can separate JNI code into a specific
+layer, so that all benefits you get from Rust in normal application or library
+development apply to the core. Within the JNI layer itself, similar benefits
+still apply with the caveat that JNI API misuse is still highly possible, and
+unsafe Rust is often necessary. Expect the JNI code for Rust to look similar to
```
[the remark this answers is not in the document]
johnwalker: I decided to remove discussion of this, since often depends entirely on what you're trying to do and may confuse people.

book/src/linking_and_dividing.md
johnwalker: Decided to just use javac for now

book/src/linking_and_dividing.md
johnwalker: Fixed all instances

book/src/linking_and_dividing.md
johnwalker: Fixed

book/src/linking_and_dividing.md
johnwalker: Updated this part

book/src/introduction.md
```diff
@@ -0,0 +1,169 @@
...
+from gcc, some systems will never have Rust debuggers, and it's possible that
+you need wider platform support than what Rust supports today. If these problems
+outweigh the benefits for you, then C may be better option, unless you are
+interested in improving the toolchain.
+
+## Prerequisites
+For Java, you need experience with defining classes. For Rust, you should have
+experience with `Box`, and defining `struct`s and `function`s.
+
+Your system needs Java 11+, recent versions of Cargo and Rust, and one Java
```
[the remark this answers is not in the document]
johnwalker: I rewrote this part to explicitly mention that jni-rs is compatible with older versions, and that it's strictly a tutorial dependency.

johnwalker:
@dmitry-timofeev thoughts so far? I'm thinking of fleshing out the memory section and maybe adding an example for using hyper over jni.

dmitry-timofeev:
@johnwalker , sorry for the massive delay! I'll try to review the updates and the next chapters in the coming weeks. I've recently switched jobs, and no longer work for a company-active jni-rs user, so can only contribute in my personal capacity. 

If your project or company starts using jni-rs extensively, please consider subscribing for updates and getting involved in reviews/design discussions. This project was originally created by @jrobsonchase for Prevoty, and then was led by Exonum Java Binding team, with many contributions from engineers from other projects. Having feedback from direct usage helps a lot.

book/src/introduction.md
```diff
@@ -0,0 +1,169 @@
...
+wanted.
+
+JNI, regardless of the language used to build the shared library, comes with a
+lot of sharp edges. The shared library needs to be built for each platform you
+support, you have to figure out a distribution strategy for them (e.g., have the
+consumer unzip the jar and put in on their load path), and JNI misuse can lead
+to resource leaks or JVM crashes. There are also performance implications of
+using JNI in certain ways. For example, Java Strings and Buffers sometimes have
+to be copied when they're used in the shared library. Due to the drawbacks of
+JNI, it's usually more appropriate to build using pure Java.
```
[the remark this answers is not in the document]
dmitry-timofeev: I think JNR allows to use native libraries without any glue code (you define only a Java interface and annotate it). There is also JNA. 

The difference between them, if I remember correctly, is that JNR generates the glue code in object form (so it may be as efficient as hand-written JNI); whilst JNA uses a native library with one method that performs dynamic invocation by symbol name.

book/src/linking_and_dividing.md
dmitry-timofeev: Mostly, simplicity, and also the fact that you get an error as soon as it occurs. Or are you saying that the benefit of this approach is that any subsequent usages (e.g., in tests) will include the original message (without this, you will have to find the _first_ error message).

book/src/cleaning.md
```diff
@@ -0,0 +1,62 @@
+# TODO Cleaning up Resource Leaks
+
+Since Java doesn’t automatically free the memory on Rust’s heap when the
+`Counter` garbage collected, there could be a memory leak. Many native resources
+need similar treatment, and so we are going to automate it using the [Cleaner
```
dmitry-timofeev, line 5: Even if we don't cover the approach with explicit memory management, I think it shall be mentioned with a todo to expand on this topic, with some considerations on choosing the right approach.

Cleaner, as any GC-based approach, is not universally applicable:
- Cleaning actions are performed in a separate thread, hence the Java class has to be thread-safe
- If the app creates Java proxies of dependent native objects, they likely have to be destroyed in the reverse order of their creation — something that is possible with PhantomReferenceQueues, but will add quite some complexity
- Any finalization-based approach is not suitable for high-volume de-allocation rate. I no longer have the benchmarks, as they are in private Exonum repository, but about three years ago we also considered which approach to take o simplify the memory management, and prototyped:
    - Explicit (it may be eventually hidden in a single place if all native peers are created in a certain bounded context — that's how EJB works)
    - GC-based approaches:
      - `finalize`
      - `PhantomReferenceQueue`, no explicit clean action
      - `PhantomReferenceQueue`, with explicit clean action

Any GC-based approaches were at least an order of magnitude slower. The GC-based approaches compared in the following way, from fastest to slowest:
1. `PhantomReferenceQueue`, with explicit clean action
2. `finalize`
3. `PhantomReferenceQueue`, no explicit clean action
      
See also [my answer](https://stackoverflow.com/a/44799935) from the time we researched this topic in Exonum.

book/src/counter.md
```diff
@@ -0,0 +1,331 @@
...
+class Counter {
+    public Counter() {}
+
+    public int get() {
+    }
+
+    public int increment() {
+    }
+
+    public void close() {
```
dmitry-timofeev, line 22: Shall we clarify here or in the previous paragraph what this method is supposed to do (destroy the native peer)?

dmitry-timofeev, line 22: I'd consider implementing AutoCloseable so that this class can be used in try-with-resources statement.

book/src/counter.md
```diff
@@ -0,0 +1,331 @@
...
+The counter needs to store a pointer back to native memory, so that the native
+methods can find the memory it owns. Instantiating a new `Counter` object will
+allocate the native Counter, and store a pointer to it in `long ptr` within the
+Java object. `ptr` will then be passed down to each of the native methods, for
+fetching the counter, incrementing it, and closing it.
+
+```java
+class Counter {
+    // The pointer to the counter in the native (Rust's) heap
+    private long ptr;
```
dmitry-timofeev, line 41: final?

book/src/counter.md
```diff
@@ -0,0 +1,331 @@
...
+```
+
+Now, we have a few goals for the implementation.
+
+1. `counter_1new` should allocate a counter on the heap, and return a valid
+   pointer to it.
+2. `counter_1new`, `counter_1increment`, and `counter_1get` should not deallocate the counter.
+3. `counter_1destroy` should deallocate the counter.
+
+We'll need to use [`Box`](https://doc.rust-lang.org/std/boxed/index.html) to
```
dmitry-timofeev, line 162: I wish we already had https://github.com/jni-rs/jni-rs/issues/84 moved from EJB into a satellite library to not have to discuss this :)

book/src/counter.md
```diff
@@ -0,0 +1,331 @@
...
+meet these goals. `Box` is used for storing objects on the heap. Boxed objects
+can't outlive their boxes, except through through `Box::into_raw` and
```
dmitry-timofeev, line 164:
```suggestion
can't outlive their boxes, except through `Box::into_raw` and
```

book/src/counter.md
```diff
@@ -0,0 +1,331 @@
...
+JNI function, except `counter_1destroy`.
+
+We'll also need to be able to translate `ptr` back into a `Box` via
+[`Box::from_raw`](https://doc.rust-lang.org/std/boxed/struct.Box.html#method.from_raw),
+in each of `counter_1increment`, `counter_1get`, and `counter_1destroy`.
+`Box::into_raw` and `Box::from_raw` are frequently necessary at language
+boundaries, since they enable the other language to play a role in memory
+management. By virtue of letting multiple languages play together with memory,
+it's easier to make memory errors at this boundary than elsewhere - either in
+pure Java, or in Pure Rust.
```
dmitry-timofeev, line 177:
```suggestion
pure Java, or in pure Rust.
```

book/src/counter.md
```diff
@@ -0,0 +1,331 @@
...
+) -> jlong {
+    let boxed_counter = Box::new(Counter::new());
+    Box::into_raw(boxed_counter) as jlong
+}
+```
+
+## Implement Increment and Get
+
+To implement increment, we'll need to convert `ptr` back into a `Box<Counter>`,
+and then call increment, drop the box through `into_raw` to about deallocating
```
dmitry-timofeev, line 199: avoid deallocating?

book/src/counter.md
```diff
@@ -0,0 +1,331 @@
...
+In this section, we've covered how to let a heap-allocated Rust object outlive
+its stack for the purpose of cross-language memory management. We will return to
+this API in the [Cleaning](./cleaning.md) section, where we will see how we can
+make it harder to run into memory leaks, and in [Java Objects in
+Rust](./java_objects_in_rust.md), where we will discuss using arrays of `Counter`.
+
+## Exercises
+
+1. What happens if destroy is never called, or gets called multiple times?
+2. Does the memory created by Rust count towards heap size limits set through
```
dmitry-timofeev, line 329: The ultimate answer from async-profiler author, Andrei Pangin: https://www.youtube.com/watch?v=c755fFv1Rnk :-)

book/src/debugging.md
```diff
@@ -0,0 +1,63 @@
...
+
+Run a test with debug and add a breakpoint here:
+
+```
+●   static native int divide(int a, int b);
+```
+
+You'll need to find the PID of your test. If you're using Java 11, then you can
+fetch it using the `ProcessHandle` API, as shown below. Otherwise, you'll need
+to find the PID through other means, such as through `ps -x` on Unix-like
```
dmitry-timofeev, line 29: There is [`jps`](https://docs.oracle.com/javase/9/tools/jps.htm#JSWOR733) coming with any JDK, possibly, it is also available on Windows.

The docs say though "experimental and unsupported" :)

book/src/error_handling.md
```diff
@@ -0,0 +1,57 @@
...
+```
+
+## Exercises
+
+1. Instead of waiting for the operation to complete, identify an illegal divisor
+   ahead of time and throw an `IllegalArgumentException` instead.
+2. The function
+   [`catch_unwind`](https://doc.rust-lang.org/std/panic/fn.catch_unwind.html)
+   will allow you to execute code and recover from a call to `panic!`. This is
+   discouraged since it won’t catch all panics (in some cases, a panic may just
```
dmitry-timofeev, line 49: Is it _generally_ discouraged, or is it also discouraged in this use-case of catching panics at the FFI boundary? My understanding is that it is fine to use at FFI boundary to prevent UB. In Exonum we did catch panics, and then converted them into Exceptions: https://github.com/exonum/exonum-java-binding/blob/master/exonum-java-binding/core/rust/src/storage/list_index.rs#L54
    
If it's fine (and even, recommended) to use here, I'd clarify that.

book/src/java_objects_in_rust.md
```diff
@@ -0,0 +1,75 @@
...
+       [`delete_local_ref`](https://docs.rs/jni/0.15.0/jni/struct.JNIEnv.html#method.delete_local_ref),
+       as you might in C++.
+    2. Give each iteration its own local reference frame using
+       [`with_local_frame`](https://docs.rs/jni/0.15.0/jni/struct.JNIEnv.html#method.with_local_frame),
+       so that all references within that frame are released by the next
+       iteration.
+    3. Use the
+       [`AutoLocal`](https://docs.rs/jni/0.15.0/jni/struct.JNIEnv.html#method.auto_local)
+       helper implemented within the JNI crate, so you don’t have to manually
+       track the usage of that reference at all.
```
dmitry-timofeev, line 54: I think a rule of thumb is that if you create more than a couple of AutoLocals,
       with_local_frame will be more efficient (I think we have benchmarks for that in `//benches/`),
       therefore it's not like AutoLocals are always better. If docs are unclear on that, it might be
       better to update them instead of repeating them here.

book/src/java_objects_in_rust.md
```diff
@@ -0,0 +1,75 @@
...
+call `delete_local_ref` when the object goes out of scope. By relying on the
+compiler’s ownership and borrowing checks, AutoLocal safely tracks the local
+usage of your objects and prevents you from releasing a reference too early. How
+does it do that? `AutoLocal` takes ownership of the reference when you
+initialize it, so you can’t have any other references to one (without using
+unsafe code). The method `AutoLocal.as_obj()` will hand out `JObject` references
+*only as long as* those references are shorter-lived than the `AutoLocal`
+itself.
+
+This is a contrived example (we would probably convert to `long` on the Java
```
dmitry-timofeev, line 70: My concern is the same: we show an example that doesn't seem to make sense
(Java loop over NativeCounters will be faster, because in that case no "Rust -> Java"
calls will happen; passing an array of longs may or may not be faster than a Java loop, but certainly will be faster than the example).

Maybe passing an array of Points to print/render them? Again, Points as Java objects might not be the most efficient solution in such a case :)

book/src/introduction.md
```diff
@@ -0,0 +1,168 @@
...
+crashes. There are also performance implications of using JNI in certain ways.
+For example, Java Strings and Buffers sometimes have to be copied when they're
+used on the native side. Due to the drawbacks of JNI, it's often worth
+considering alternatives (like reimplementation in one language or IPC/RPC).
+Eventually, [Project Panama](https://openjdk.java.net/projects/panama/) may also
+offer a simpler or more efficient API.
+
+There are plenty of reasons why you should consider Rust if you need to write
+native code that works with Java. Most notably, you can separate JNI code into a
+specific layer, so that all benefits you get from Rust in normal application or
```
dmitry-timofeev, line 22: We may also say that unlike C, jni-rs offers some safer to use abstractions.

For example, there are Executor that ensures native threads get detached when they are destroyed; GlobalRef/AutoLocal/AutoByteArray that will always delete the ref; with_local_frame that won't let you miss "Pop", etc. Most of them could be implemented equally well in C++, some safety features, however, are unique to Rust (like not being able to share JNIEnv/JObject across threads).

For C++, there is https://github.com/mapbox/jni.hpp, though we didn't use that, because we had to integrate a _Rust_ framework with Java :)

book/src/linking_and_dividing.md
```diff
@@ -0,0 +1,241 @@
...
+
+1. Add a `native` function to a Java class. Remember that the package path,
+   class name, and function name are all encoded into the C function name, so
+   make sure those are as you want them. 
+2. Run `javac -h . NativeAPI.java` to produce C headers. (We assume that the
+   class resides in `NativeAPI.java`, but it doesn't have to.)
+3. Copy the C function name out of the header file it produced.
+
+Upon completing these steps, you will see a file called
+`com_github_jni_rs_jnibook_NativeAPI.h` that contains the following:
```
dmitry-timofeev, line 67: Nit: We probably shall not use com.github namespace. For a tutorial Java code that isn't uploaded on Central, we may use a shorter name, like jni_rs.jnibook or just jnibook.

book/src/linking_and_dividing.md
```diff
@@ -0,0 +1,241 @@
...
+
+// Added no_mangle
+#[no_mangle]
+pub extern "system" fn Java_com_github_jni_1rs_jnibook_NativeAPI_verify_1link(
+    _env: JNIEnv,
+    _class: JClass
+) {}
+```
+
+With that, we're done. You'll find thats exactly what the Rust starter code
```
dmitry-timofeev, line 156:
```suggestion
With that, we're done. You'll find that’s exactly what the Rust starter code
```

dmitry-timofeev (commented):
Awesome job! I've reviewed all md files, and most other files.

book/src/cleaning.md
dmitry-timofeev: Also, we can consider finishing this chapter in subsequent PRs, so that it doesn't grow too large.

johnwalker:
@dmitry-timofeev Thanks for your thoughtful response :) I'll work through your feedback today and tomorrow.

johnwalker:
Also, I'd certainly like to help out more. I'll subscribe to PRs/issues.

dmitry-timofeev:
Thanks a lot! 

Do you need a board to keep track of any smaller tasks on this tutorial (like this one: https://github.com/jni-rs/jni-rs/projects/1)? It might be more convenient than a checklist or full-blown issues.

johnwalker:
That could be extremely helpful. Would you make one? I can add some tasks in case folks want to contribute.

book/src/counter.md
johnwalker: fixed

book/src/counter.md
johnwalker: Yeah, it's just a question of when best to do it. At the moment, Counter gets AutoCloseable in the cleaning chapter. I think we could probably introduce that here, though.

johnwalker: Yes, let's do that.

book/src/java_objects_in_rust.md
johnwalker: Part of the motivation is to familiarize the reader with arrays of objects, and I agree with you that the solution doesn't make much sense for the problem considering all the extra Rust->Java calls. On the other hand, I like that it introduces field and array access.

We could probably replace NativeCounter with a different class (I'm open to it) or perform a different operation. I'm partly tempted by your suggestion to pass an array of Points or longs, since we could maybe move it to construction time (potentially not the critical path) and talk a little bit better about the cost of the JNI upcalls.

book/src/linking_and_dividing.md
johnwalker: Sounds good to me. Maybe jnirsbook?

dmitry-timofeev:
Sure: https://github.com/jni-rs/jni-rs/projects/2

book/src/linking_and_dividing.md
dmitry-timofeev: 👍 , or "jni_rs_book" so that it's slightly more readable.

book/src/java_objects_in_rust.md
dmitry-timofeev: > Again, Points as Java objects might not be the most efficient solution in such a case :)

Possibly, some objects with behaviour (not plain data holders, as with Points readers could see more efficient approaches), e.g., Java passing to native an array of `Spaceship`s, and native `Spaceship#launch`ing them.

johnwalker: For an outline of this chapter, how do you feel about:

- Problem description: Describe as an exercise for updating many X (possibly Counters)
- Solution 1: Java loop. Describe as simplest solution.
- Solution 2: Rust loop with ptrs. Describe as more complex, but potentially warranted if performance tests demonstrated an improvement.
- Solution 3: Rust loop with Java objects. Describe as most complex, with the highest number of upcalls. Mention that in certain cases, JNI performance considerations may be negligible (As in , `Spaceship#launch`).
- Analysis: Summarize tradeoffs for each solution. (Number of upcalls/downcalls in each, complexity, additional code, flexibility for future changes). Conclude with an author's preference for 1 based off simplicity. Note that which of Solution 1 and 2 is faster depends on the JNI implementation, and list some considerations (for example, array copying).

As a result, the reader should get exposed to a few different patterns for managing arrays across Rust and Java, and they would be more inclined to reach for Solution 1, or try Solution 2 after testing performance themselves. However, they'd also be able to get data out of a Java object and LongArrays if they needed to.

book/src/java_objects_in_rust.md
johnwalker: The `AutoLocal` docs (https://docs.rs/jni/0.18.0/jni/objects/struct.AutoLocal.html) don't currently mention `with_local_frame`. I'd also expect `with_local_frame` to be more efficient, and think that's worth mentioning in the rustdocs.

Is there any way I can add this as a task to the project board? I don't think I have permissions to manipulate it.

book/src/cleaning.md
johnwalker: Yeah, we'll need to expand on this over time. I'll edit it to state that the chapter isn't a complete guide to memory management of native resources, and describe the upsides/downsides of the approach, then mention the alternatives you've described. Then, we can expand on those over time - potentially in chapters that revisit the same topic, or by additions to this particular chapter.

book/src/counter.md
johnwalker: Yeah that does look helpful. Fortunately, it's not a one-way door. Once #84 is resolved, we can decide how to introduce it here.

book/src/java_objects_in_rust.md
dmitry-timofeev: > Is there any way I can add this as a task to the project board? I don't think I have permissions to manipulate it.

I've sent a collaborator invitation earlier, but see it is still pending. Likely, it must be somewhere in the mail. I think when you accept you must be able to modify the board.

johnwalker: Aha, found it. Thank you :)

book/src/java_objects_in_rust.md
dmitry-timofeev: Depends on what this chapter attempts to cover. Is the focus on using Java objects in Rust, or accessing Java arrays?

For instance, if the focus is on passing ~~Java objects~~ data, I think we can make an example with a single argument, and consider two "types" of objects the native might need:
- Plain data carriers. In this case, accessing Java objects may or may not be the best idea (some of the considerations are the speed of development and the runtime efficiency). There are at least three viable approaches:
  - Pass Java objects as-is. Native -> Java method invocations are rather expensive; don't have data on field accesses. However, sometimes it might be simpler than other approaches in some cases.
  - Unroll a Java object into a collection of primitive fields (e.g., instead of passing a single `3dPoint`, pass `x, y, z`).
  - Pass the Java object in a serialized form as a byte array, e.g., a protocol buffer. Depending on the number and types of fields and serde performance, might or might not be better.
- Objects with behaviour that Rust _needs_ to call (e.g., Java object implements a Rust trait). In this case, no much choice here: you have to pass a `Spaceship` if the `MissionControlCenter` is in Rust.

If we limit tje discussion to objects (and do not discuss passing data), there will be only one solution; if extend to data: at least three.

I'd consider keeping arrays as a separate sub-section, if we discuss passing data.

johnwalker: Based off this feedback, I've introduced arrays in a separate exercise (together with string passing). (Actually, the whole thing is kind of rewritten, to make it less wordy and more about completing exercises.) I think the discussion in the chart exercise should also be updated to cover object unrolling, and plan on introducing native->Java performance separately (maybe as a profiler lab, so that people can try it themselves).

book/src/introduction.md
johnwalker: Thanks, I've updated the tone and discussion here, so that it talks more about jni-rs abstractions. The later exercises/solutions should also include some discussion of the drop guarantees, since they are really useful.

book/src/counter_solution.md
```diff
@@ -0,0 +1,382 @@
...
+for Java. The Java side is purposefully oversimplified, so that we may focus on
+Rust (which would not change, with a more correct Java implementation).
+
+## Java Counter API
+First, we need to add a `close()` to the counter, so that we can free native
+resources.
+
+```java
+// Autocloseable is used to allow try-with-resources
+class Counter implements Autocloseable {
```
dmitry-timofeev, line 13:
```suggestion
// AutoCloseable is used to allow try-with-resources
class Counter implements AutoCloseable {
```

dmitry-timofeev: And below

book/src/counter_solution.md
```diff
@@ -0,0 +1,382 @@
...
+        return NativeAPI.counter_increment(ptr);
+    }
+
+    public void close() {
+        // Note the following problems with this implementation of close:
+        //
+        // 1. Close is not idempotent, as encouraged in the Autocloseable docs
+        // 2. The counter may be double freed.
+        // 3. Close may race with calls to get or increment
+        // 4. It's not guaranteed that a caller will call close.
```
dmitry-timofeev, line 65: I wouldn't say that both are necessarily problems:
(3) — not an issue unless the class is specified as thread-safe (i.e., it is a misuse to call increment and close from different threads).
(4) — depending on how the class is used, it might be a valid choice to make it `AutoCloseable` and document as one that must be closed by its users (e.g., using try-with-resources). In some cases, it will get unwieldly (many objects; or long-living objects), and other solutions would be better (`Cleaner`s or context-based approaches).

book/src/counter_solution.md
```diff
@@ -0,0 +1,382 @@
...
+}
+```
+
+## Summary
+In this exercise, we examined a few ways to wrap a Rust Counter for use from
+Java. You should be familiar with how panic can interact with `drop`, and some
+approaches for preventing resources from being dropped prematurely. There are
+still these problems with the counter:
+
+1. `increment` and `get` may use the counter after its been freed.
```
dmitry-timofeev, line 376:
```suggestion
1. `increment` and `get` may use the counter after it's been freed.
```

book/src/counter_solution.md
```diff
@@ -0,0 +1,382 @@
...
+2. it can be double freed ( via multiple `close` calls).
+3. or perhaps never freed (never call `close`).
+4. `close()` is not idempotent, as encouraged in the [`Autocloseable`
+   docs](https://docs.oracle.com/javase/8/docs/api/java/lang/AutoCloseable.html#close--).
+5. it's not threadsafe; it's possible to call `get` during `destroy`, for
```
dmitry-timofeev, line 381:
```suggestion
5. it's not thread-safe; it's possible to call `get` during `destroy`, for
```

book/src/error_handling.md
```diff
@@ -31,27 +49,116 @@ pub extern "system" fn Java_com_github_jni_1rs_jnibook_NativeAPI_divide(
...
+    match f() {
+        Ok(s) => s,
+        Err(e) => {
+            if let Some(e) = e.downcast_ref::<jni::errors::Error>() {
+                match e {
+                    // Since JavaException implies an Exception has already been thrown, 
+                    // don't throw another one.
+                    jni::errors::Error::JavaException => {},
+                    _ => {
+                        let _ = env.throw_new("java/lang/Exception", e.to_string());
```
dmitry-timofeev, line 84: It is not enough to prevent exception masking: you might get some error type that is not `jni::errors::Error::JavaException` _and_ have a pending exception in Java. The only reliable way is to check if there is an exception pending already:

https://github.com/exonum/exonum-java-binding/blob/a6627aa00d10bc8175d68392e9156fbf99f48bfe/exonum-java-binding/core/rust/src/utils/errors.rs#L162

book/src/error_handling.md
```diff
@@ -31,27 +49,116 @@ pub extern "system" fn Java_com_github_jni_1rs_jnibook_NativeAPI_divide(
...
+                    }
+                }
+            } else {
+            // This branch handles all other error types, such as those returned by
+            // all non-JNI code
+                let _ = env.throw_new("java/lang/Exception", e.to_string());
```
dmitry-timofeev, line 90: Since it appears as a library method, using _checked_ "java/lang/Exception" is not generally applicable (as it is valid to have native method declarations without checked exceptions). RuntimeException seems more appropriate.

dmitry-timofeev: I wonder if there might be client code that returns non-jni error, but has a pending Java exception (e.g., by ignoring a jni error that they got; and getting an error from somewhere else).

book/src/error_handling.md
dmitry-timofeev: Possibly, even before downcast (see the comment below)

book/src/introduction.md
```diff
@@ -2,167 +2,107 @@
...
-RUST_BACKTRACE=1 LD_LIBRARY_PATH=`readlink -m ../jnibookrs/target/debug` ./gradlew test --info
+    at java.base/java.lang.ClassLoader.loadLibrary(ClassLoader.java:2660)
+    at java.base/java.lang.Runtime.loadLibrary0(Runtime.java:829)
+    at java.base/java.lang.System.loadLibrary(System.java:1867)
+    at com.github.jni_rs.jnibook.NativeAPI.<clinit>(NativeAPI.java:10)
 ```
 
-Assuming that the test succeeded, you're done! Consider setting up the same
-configuration for your Java editor.
+Then verify that your path is listed Exception message. If it's not, then you
```
dmitry-timofeev, line 103:
```suggestion
Then verify that your path is listed in the Exception message. If it's not, then you
```

book/src/chart_exercise.md
```diff
@@ -0,0 +1,102 @@
...
+
+## Docs to get you started
+These are some selections from the `jni-rs` docs that may help you.
+
+### Reading Java Array Lengths
+[`JNIEnv::get_array_length(&self, array: jarray) ->
+Result<jsize>`](https://docs.rs/jni/0.18.0/jni/struct.JNIEnv.html#method.get_array_length)
+can be used to read the length of Java Arrays. Note that it can be better to
+pass the array length together with the array, down to the native layer, when
+that is an option, since it saves an upcall.
```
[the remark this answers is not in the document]
johnwalker: That's a good point, it's very likely that some JNI calls are less expensive than others, and that getting length is common enough that some VMs will handle it relatively well. I'll remove that suggestion, because it depends on the implementation.

book/src/error_handling.md
johnwalker: Thanks :)

Right, `RuntimeException` is much better. A colleague also pointed out that we'd need to decorate the native functions with `throws Exception`, which wasn't done in the examples. I'll update this to `RuntimeException`

I agree, this doesn't handling pending exceptions properly. Any unchecked (doesn't check for an exception, when an exception is possible) JNI method can lead to problems. I'll review exonum and see if we can get a better pattern here.

dmitry-timofeev: > Any unchecked (doesn't check for an exception, when an exception is possible) JNI method can lead to problems

I think all methods that are required by the spec do check for exceptions (if not — that's a bug). I mostly though of some strange client code that, for example, throws an exceptions/invokes a method that throws an exception, and does not return that error straight away. Hopefully, you can just pull this check a little higher to cover any such cases. On top of that, if an actual error does not match JniError:Exception when there is a pending excpetion in Java, we can log such a discrepancy.

book/src/chart_exercise.md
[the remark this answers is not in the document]
dmitry-timofeev: FYI, there are some benches that could be extended. Mostly they cover native -> Java method invocation using approaches on the various level of abstraction (from "pass class and method names as strings" to "look up the class ref and method ID and reuse between repeated invocations").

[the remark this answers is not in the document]
dmitry-timofeev: https://github.com/jni-rs/jni-rs/blob/master/benches/api_calls.rs

book/src/error_handling.md
johnwalker: Sounds good. I think the strange client code is likely common, so hopefully we can figure something out here. I still need to experiment with the jni-rs APIs to understand a little bit better.

One case I'm thinking about is dependent on jni-rs APIs. If call_method calls some Java method that throws an Exception, would the result be `Err` or `Ok`? Intuitively, I'd expect it to be `Ok`, and so some additional checks are necessary that the writing doesn't advertise.

book/src/chart_exercise.md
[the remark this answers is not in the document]
johnwalker: I'm tempted to include a reference to the benches in the materials (perhaps in future topics). What do you think?

johnwalker:
Also, I think I don't have permissions to add cards to this board: https://github.com/jni-rs/jni-rs/projects/2

book/src/error_handling.md
johnwalker: Perhaps this is an example?

```
try_java(|| {
  env.call_method(obj, "throwAnException", "()V"); // ignore exception.
  env.call_method(obj, "doSomethingElse", "()V"); // exception still pending
});
```

johnwalker: Maybe something like this instead?
```rust

fn try_java<F, T>(env: JNIEnv, error_value: T, f: F) -> T
    where
        F: FnOnce() -> Result<T, anyhow::Error>,
{
    match f() {
        Ok(s) => s,
        Err(e) => {
            // Only throw an exception is one isn't already pending.
            if !env.exception_check().unwrap() {
                env.throw_new("java/lang/RuntimeException", e.to_string())
            }
            error_value
        }
    }
}
```

book/src/error_handling.md
johnwalker: I think I see your point - if we have some nested calls that translate a `JavaException` into some other value or type, this will fail.

book/src/error_handling.md
dmitry-timofeev: > So far, I haven't found a JNI function call that returns Ok when it finishes with an exception pending

There might be some examples that may return Error, but of a different type (e.g., if they check an int error code first). But generally, methods that wrap JNI APIs that can throw exceptions use a macro that checks them for you (so that the client code rarely, if ever, have to). For example:
https://github.com/jni-rs/jni-rs/blob/master/src/wrapper/macros.rs#L21

> I think this is probably OK to start with.

+1; exception_check must not be particularly expensive: https://github.com/jni-rs/jni-rs/blob/master/benches/api_calls.rs#L206-L217

book/src/chart_exercise.md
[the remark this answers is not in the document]
dmitry-timofeev: Agree, the set of tests already gives insights into the price of some features or coding patterns; it could be extended with more tests. It's good for users to understand the trade-offs of various options we provide, on their JVM version.

book/src/error_handling.md
dmitry-timofeev: Possibly; we can expect anything from BL code :-) 

Also, just as with the module that controls the longs flying between Java and Rust to be valid pointers to objects of correct types (aka resource-manager); this method might be needed to many projects. If we can come up with a reasonably generic and performant one, we can include it in the library, as a separate PR.

book/src/future_topics.md
```diff
@@ -9,9 +9,10 @@ jni-rs has additional abstractions that haven't been discussed, like executors.
 JNI also has a few features that need their own sections (critical APIs, and
 local and global references).
 
-There are also performance optimizations that haven't been covered. It is more
-efficient to issue calls from Java to native code, than it is to go from native
-code to Java, and it's possible to cache class/methodIds/fieldIds.
+There are also performance optimizations that haven't been covered. For example,
+it is more efficient to issue calls from Java to native code than from native
+code to Java, and caching class/methodIds/fieldIds can reduce the number of
```
dmitry-timofeev, line 14: Possibly, this section could be useful to link: https://docs.rs/jni/0.18.0/jni/struct.JNIEnv.html#checked-and-unchecked-methods

dmitry-timofeev:
Weird; I guess the "Triage" role is not enough; please try with "Write". I've shared the feedback to GH team.

book/src/error_handling.md
```diff
@@ -0,0 +1,173 @@
...
+                    _ => {
+                        let _ = env.throw_new("java/lang/RuntimeException", e.to_string());
+                    }
+                }
+            } else {
+            // This branch handles all other error types, such as those returned by
+            // all non-JNI code
+                let _ = env.throw_new("java/lang/RuntimeException", e.to_string());
+            }
+            default_value
```
Progdrasil, line 94: Found this MR while looking for information on error handling with this crate.

Found this mistake where you call a value `default_value` whilst the parameter is actually `dummy_value` so either

```suggestion
            dummy_value
```

book/src/error_handling.md
```diff
@@ -0,0 +1,173 @@
...
+It can be helpful to wrap common error handling into one function, so that there
+is less repeated JNI code. For ease of implementation, we'll use
+`anyhow::Error`, and make the caller pass the `dummy_value`, which is returned
+only on error paths. The `JNIEnv` APIs all throw `Result<T,E>`, and the
+`JavaException` variant marks that an exception has already been thrown - so for
+that variant alone, we will skip throwing an Exception. You may come up with
+your own Error hierarchies, which could have a different logic.
+
+```rust
+fn try_java<F, T>(env: JNIEnv, dummy_value: T, f: F) -> T
```
Progdrasil, line 73: or this,
```suggestion
fn try_java<F, T>(env: JNIEnv, default_value: T, f: F) -> T
```

book/src/error_handling.md
johnwalker: Thanks, there are a few typos in the book. I'm going to update this tomorrow with a set of changes that includes the fix. (I switched the code samples to use Anchor tags, so that we can catch these syntactic errors more easily in the future).

Thank you for commenting that you found this though, I'd appreciate you examining the diff once it's out in case I made other mistakes.

book/src/error_handling.md
johnwalker: This is the one it uses after the correction :)

johnwalker:
Alright, apologies for the big diff. I probably should have put the structural changes for getting more help from the compiler and the edits in separate commits.

1. Addressed previous reviewer comments, updated all links to `0.19.0`, and updated the jni-rs dependency.
2. Added feature flags to a completed project, the compiler will test more for us. There are still a few sections that haven't moved there yet, but they tend to be small.
3. Added `<details>` tags to the Counter discussion section, and 1 question about using Box.
4. Added `catch_unwind` to the error_handling section. Before we just said it was possible, now we actually tell how to do it.
5. Use `expect` instead of `unwrap` in `try_java` for `throw_exception`. This is useful in case someone writes the code out manually, and uses an exception class that does not exist.
6. Fix a bug in `checkAvailability`, where `Error` is wrapped with `RuntimeException`.
7. Make the `ChartRenderer` method static. The constructor was distracting.
8. Remove the Counter solution section for now. The set of changes is big enough. I'll bring this into a separate PR.
9. Replace Exception with exception with lowercase to reduce possible confusion with checked exceptions, where appropriate.
10. Suggest that the reader use cargo watch together with gradle, to rerun tests and builds. This helps users without Java IDEs.
11. Various editing improvements.

book/src/introduction.md
```diff
@@ -8,18 +8,25 @@ intended for people that are new to JNI, and would like to learn through
...
-write JNI code for some applications. If you need to write JNI code, hopefully
-you can have fun doing it in Rust.
+JNI has a lot of sharp edges. Native code needs to be written for each supported
+platform so that Java can can interact with native functions, and misuse can
+lead to resource leaks or JVM crashes. Due to the drawbacks of JNI, it is often
+worth considering alternatives (like reimplementation in one language or
+IPC/RPC). Eventually, [Project
+Panama](https://openjdk.java.net/projects/panama/) will make it [unnecessary to
+write JNI-like
+bindings](https://github.com/openjdk/panama-foreign/blob/foreign-jextract/doc/panama_jextract.md)
```
dmitry-timofeev, line 18: The Panama features include:
- Linking to a native library — it lands in 16 as an incubator feature and already provides an alternative to JNI. Instead of implementing native methods, users will be able to assemble some MethodHandles: 
  - https://openjdk.java.net/jeps/389
  - https://github.com/openjdk/panama-foreign/blob/foreign-jextract/doc/panama_ffi.md#appendix-full-source-code
- jextract — will assemble these MethodHandles for you based on the library header.

book/src/counter_discussion.md
```diff
@@ -5,21 +5,36 @@ Can they pass it to Java, like this? Why or why not? If not, does the error
...
+someone introduces new code that panics between `from_raw` and `into_raw`, and
+introduces a panic handler. In this case, the following occurs during `get`:
+
+1. Rust makes a `Box<Counter>`, then panics panics.
+2. `panic!` leads to the `Counter` being dropped.
+3. The program recovers from the panic.
+
+At this point, the Counter has already been unintentionally freed, and a
+use-after-free or double free is likely the next time any function on the
+Counter is called.
```
dmitry-timofeev, line 38: It could be unclear to the reader what to use :) 

The corresponding code in EJB returns a reference to avoid this situation: https://github.com/exonum/exonum-java-binding/blob/a6627aa00d10bc8175d68392e9156fbf99f48bfe/exonum-java-binding/core/rust/src/handle/mod.rs#L39-L55 

Possibly, it might make sense to first add this library code; and then just use it in the solution.

book/projects/starter/jnibookrs/src/division.rs
```diff
@@ -0,0 +1,72 @@
+// ANCHOR: imports
```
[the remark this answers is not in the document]
johnwalker: Ah, I'm crazy and didn't notice. Fixing

book/src/future_topics.md
```diff
@@ -1,18 +1,18 @@
...
 There are also performance optimizations that haven't been covered. For example,
 it is more efficient to issue calls from Java to native code than from native
-code to Java, and caching class/methodIds/fieldIds can reduce the number of
-upcalls that are necessary.
+code to Java, and [caching class/methodIds/fieldIds can reduce the number of
+upcalls that are
+necessary.](https://docs.rs/jni/0.19.0/jni/struct.JNIEnv.html#checked-and-unchecked-methods)
 
 There is open work that could simplify portions of the book. For example, [a
 native peer registry](https://github.com/jni-rs/jni-rs/issues/84) can help
```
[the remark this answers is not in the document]
johnwalker: Fixed, thanks

book/src/introduction.md
johnwalker: Thanks :) I've updated it to include these, let me know what you think.

book/src/counter_discussion.md
johnwalker: I agree, it would be much better if it offered a solution here. I've backed off from providing one because I haven't found a reference for the invariants the programmer must ensure hold when they perform the cast. The unsafe code guidelines/rustonomicon/rust book didn't have anything that clicked here.

There could be one other alternative, which is dereferencing a raw pointer instead.

book/projects/starter/jnibookrs/src/division.rs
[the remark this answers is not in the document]
johnwalker: Fixed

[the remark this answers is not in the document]
johnwalker: (It was also just that one with this property)

rib:
Hi @johnwalker 

Sorry that it's literally been years since you first started working on this and then things went quiet before there was a final resolution for this work.

Especially in the last year or so the project hasn't had lots of active development and I've only quite recently taken over maintaining the project (though still rather ad-hoc, not with a lot of dedicated time) and have been working through the backlog of issues and PRs to figure out how to address them.

I think @dmitry-timofeev was most familiar with the contents of your work and was doing a good job at providing feedback but it looks like that review process lost steam around February last year.

It looks like there are some good gems of information in the book even if it wasn't entirely complete and it would be good if those things could be shared somehow.

Unfortunately it seems like adding a book to this repo probably just ends up raising the bar for publishing something because all the content needs to have buy in + review from the project maintainers and we probably need to be a more considerate of how we would plan to maintain it over time (e.g. considering evolving API changes).

On the other hand if the same content were published independently in some form you could still seek review/feedback but there would be no other constraints on how / when you publish the content. E.g. Maybe some of the content could be adapted into blog posts.

We could also sign post from this project e.g. to blog posts or a personal repo hosting this book in a way that still means people could discover it via this project.

I have to admit that I don't think I'm going to be in a position to pick up an active review process for this much content and so I don't currently see that there will be much progress here unless another maintainer comes along with more time to dedicate.

If you have a moment, it'd be good to hear your thoughts.

rib:
Okey, I'm going to close this issue for now. Sorry this work didn't land in the end. I'm still subscribed to this issue so should see if there are any follow up comments while it's closed.

johnwalker:
Thanks for following up, it likely needs a bit of a refresh. It sounds like you're OK with this being published separately, perhaps with some refresh. And maybe in the future we can re-evaluate whether to transition ownership. What do you think?

PR #255, opened 2020-09-17
Make JNIEnv `Copy`

jack-signal:
## Overview

Currently JNIEnv is `Clone` but not `Copy`, however there seems no difficulty in having it also be `Copy`.

Possibly I've missed or misunderstood something such that this is a bad idea, in which case I look forward to learning the reason. Thank you.

If this is acceptable but needs some additional test or documentation work just let me know.

### Definition of Done

- [x] There are no TODOs left in the code
- [ ] Change is covered by automated [tests](https://github.com/jni-rs/jni-rs/blob/master/CONTRIBUTING.md#tests)
- [x] The [coding guidelines](https://github.com/jni-rs/jni-rs/blob/master/CONTRIBUTING.md#the-code-style) are followed
- [ ] Public API has documentation
- [x] This change is not breaking **or** mentioned in the Changelog
- [ ] The [continuous integration build](https://www.travis-ci.org/jni-rs/jni-rs) passes

PR #259, opened 2020-09-23
Add repr(transparent) attribute to JavaVM struct

aleksuss:
## Overview

Adding of the attribute makes rust 1.46 compiler happy about issue: `extern` fn uses type `jni::JavaVM`, which is not FFI-safe`
 
```
warning: `extern` fn uses type `jni::JavaVM`, which is not FFI-safe
  --> src/utils/jni_cache.rs:67:39
   |
67 | pub extern "system" fn JNI_OnLoad(vm: JavaVM, _: *mut c_void) -> jint {
   |                                       ^^^^^^ not FFI-safe
   |
   = note: `#[warn(improper_ctypes_definitions)]` on by default
   = help: consider adding a `#[repr(C)]` or `#[repr(transparent)]` attribute to this struct
   = note: this struct has unspecified layout
``` 

### Definition of Done

### DIFF
--- a/src/wrapper/java_vm/vm.rs
+++ b/src/wrapper/java_vm/vm.rs
@@ -130,6 +130,7 @@ use crate::InitArgs;
 /// [actp]: struct.JavaVM.html#method.attach_current_thread_permanently
 /// [actd]: struct.JavaVM.html#method.attach_current_thread_as_daemon
 /// [spec-references]: https://docs.oracle.com/en/java/javase/12/docs/specs/jni/design.html#referencing-java-objects
+#[repr(transparent)]
 pub struct JavaVM(*mut sys::JavaVM);
 
 unsafe impl Send for JavaVM {}
--- a/tests/threads_nested_attach_daemon.rs
+++ b/tests/threads_nested_attach_daemon.rs
@@ -45,7 +45,5 @@ pub fn nested_attaches_should_not_detach_daemon_thread() {
         assert_eq!(jvm().threads_attached(), 1);
     }
     assert_eq!(jvm().threads_attached(), 1);
-
-    drop(env);
     assert_eq!(jvm().threads_attached(), 1);
 }
--- a/tests/threads_nested_attach_daemon.rs
+++ b/tests/threads_nested_attach_daemon.rs
@@ -45,5 +45,4 @@ pub fn nested_attaches_should_not_detach_daemon_thread() {
         assert_eq!(jvm().threads_attached(), 1);
     }
     assert_eq!(jvm().threads_attached(), 1);
-    assert_eq!(jvm().threads_attached(), 1);
 }
--- a/tests/threads_nested_attach_permanently.rs
+++ b/tests/threads_nested_attach_permanently.rs
@@ -47,7 +47,6 @@ pub fn nested_attaches_should_not_detach_permanent_thread() {
     assert_eq!(jvm().threads_attached(), 1);
 
     // Dropping JNIEnv obtained from a permanent attach has no effect on the thread status.
-    drop(env);
     assert_eq!(jvm().threads_attached(), 1);
     assert!(jvm().get_env().is_ok());
 }
--- a/tests/threads_nested_attach_permanently.rs
+++ b/tests/threads_nested_attach_permanently.rs
@@ -45,8 +45,5 @@ pub fn nested_attaches_should_not_detach_permanent_thread() {
         assert_eq!(jvm().threads_attached(), 1);
     }
     assert_eq!(jvm().threads_attached(), 1);
-
-    // Dropping JNIEnv obtained from a permanent attach has no effect on the thread status.
-    assert_eq!(jvm().threads_attached(), 1);
     assert!(jvm().get_env().is_ok());
 }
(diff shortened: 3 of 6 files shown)


tests/threads_nested_attach_daemon.rs
```diff
@@ -45,7 +45,5 @@ pub fn nested_attaches_should_not_detach_daemon_thread() {
         assert_eq!(jvm().threads_attached(), 1);
     }
     assert_eq!(jvm().threads_attached(), 1);
-
-    drop(env);
```
maurolacy, line 49: Why are you removing this?

tests/threads_nested_attach_permanently.rs
```diff
@@ -47,7 +47,6 @@ pub fn nested_attaches_should_not_detach_permanent_thread() {
     assert_eq!(jvm().threads_attached(), 1);
 
     // Dropping JNIEnv obtained from a permanent attach has no effect on the thread status.
-    drop(env);
```
maurolacy, line 50: Same here.

src/wrapper/java_vm/vm.rs
```diff
@@ -130,6 +130,7 @@ use crate::InitArgs;
 /// [actp]: struct.JavaVM.html#method.attach_current_thread_permanently
 /// [actd]: struct.JavaVM.html#method.attach_current_thread_as_daemon
 /// [spec-references]: https://docs.oracle.com/en/java/javase/12/docs/specs/jni/design.html#referencing-java-objects
+#[repr(transparent)]
```
maurolacy, line 133: Makes sense. Doc for reviewers: https://doc.rust-lang.org/1.26.2/unstable-book/language-features/repr-transparent.html

tests/threads_nested_attach_daemon.rs
aleksuss: Because of https://travis-ci.org/github/jni-rs/jni-rs/jobs/729379085#L258

aleksuss: By the way. What's a point to call drop here ? As I can see `JNIEnv` doesn't implement `Drop` trait.

maurolacy: I see. Thanks for the clarification.

I'm thinking this is probably related to a recent change (#255), where `JNIEnv` now implements `Copy`.

tests/threads_nested_attach_permanently.rs
stanislav-tkach: It looks like a part of the test, but `drop` for `JNIEnv` doesn't do anything. We have several similar tests, but only one of them (`nested_attaches_should_not_detach_guarded_thread`) uses `AttachGuard`.

stanislav-tkach (approved):

tests/threads_nested_attach_permanently.rs
dmitry-timofeev: Doesn't it make sense to leave it? It seems to verify that whatever `Drop` implementation JNIEnv has, it does not detach a permanently attached thread. At the moment it has no implementation, but if it ever changes, this test will verify that `attach_current_thread_permanently` contract is preserved: i.e., "permanently" means "until explicitly detached, and no through other means".

tests/threads_nested_attach_daemon.rs
dmitry-timofeev: I'd consider clarifying the intent, as in the other method, e.g.,

```
    // Dropping JNIEnv obtained from a daemon attach has no effect on the thread status.
```

dmitry-timofeev (changes requested):
Thanks! The attribute and the clean ups make sense, but I think the tests shall keep these verifications, possibly, with some clarifications and SA suppressions, if needed.

The JNIEnv already has the attribute, are there any other types that need it?

tests/threads_nested_attach_permanently.rs
stanislav-tkach: The type cannot implement both `Copy` and `Clone`, so we cannot implement `Drop` without breaking changes (removing `Copy` implementation). We can silence this warning in order to preserve drop in the test, but I'm not sure if it makes sense. I guess extending the comment will be enough.

dmitry-timofeev: I see, thanks for the explanation. Then I suggest to remove the redundant comments and assertions. @aleksuss could you do that please, or would you prefer to postpone these changes till a separate PR?

dmitry-timofeev:
@maurolacy , big thanks for getting involved in the reviews and issue discussions, we really appreciate that! Neither Stanislav nor I currently work for companies that are jni-rs users, hence we cannot spend any company time on this project. We also no longer get any internal feedback on this project — that in Exonum was one of the biggest sources of improvements, with the Exonum team contributing numerous improvements and fixes over the years. It is very useful to have an active user involved in discussions around the project.

maurolacy:
@dmitry-timofeev my pleasure. Contributing here was and still is a great learning opportunity for me.

dmitry-timofeev (approved):
LGTM, thanks!

PR #261, opened 2020-09-24
Add conversion JvmError into Error

aleksuss:
## Overview

It would be convenient to make such conversions in user's code.

### Definition of Done

### DIFF
--- a/src/wrapper/errors.rs
+++ b/src/wrapper/errors.rs
@@ -37,6 +37,8 @@ pub enum Error {
     ThrowFailed(i32),
     #[error("Parse failed for input: {1}")]
     ParseFailed(#[source] combine::error::StringStreamError, String),
+    #[error("Error occurred in JavaVM: {0}")]
+    JvmError(String),
     #[error("JNI call failed")]
     JniCall(#[source] JniError),
 }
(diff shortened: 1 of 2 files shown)


stanislav-tkach:
I suppose it can be useful if some code encapsulates `InitArgsBuilder::build` and other JNI calls, but I'm not sure if that is generally convenient because looking at the `Error` definition one would expect possible `JvmError` from _every_ JNI function.

dmitry-timofeev:
Are there any issues with two _separate_ hierarchies of errors? Is it a good practice in Rust to keep errors coming from two different types (JNIEnv and JavaVM) but the same library under the same hierarchy? What are the consideration?

stanislav-tkach:
I think that having a single error type for all (including unrelated) operations makes api less clear. Applications that don't care about the specific error type and only want to propagate it can use libraries like `anyhow` or even `Box<dyn Error>`.

rib:
It's a little bit random that there's this special case `JvmError` that's just used by `InitArgsBuilder::build()` and I do wonder if that could perhaps at least be renamed to make it a little clearer that it's not a very general purpose error.

More generally though I also don't really see that it needs to be possible to map the `JvmError`s into `Error`s while they are used for different things.

Please re-open if more specific context could clarify why this would be helpful.

PR #264, opened 2020-09-30
Implement `TryFrom<JValue>` for JValue variants

### LINKED ISSUE #262: Implement `TryInto<T>` for JValue with types that implement `Into<JValue>`
Hi, I'm working on a library that interfaces with `jni` and it would make my life *a lot* easier if types that implement `Into<JValue>` (e.g. `bool`, `f32`, `JObject`, etc...) implemented `TryInto`, e.g. by simply calling the already defined methods `z()`, `f()`, etc. I don't see any technical limitations to this.

The end result would be something like:

```rust
let raw: JValue = /* some JValue */;
let value: Result<i32> = raw.try_into();
/* do something with `value` */
```

If you are ok with this I could even implement the necessary code and follow through with a PR. :smiley:

giovanniberti:
Closes #262

### DIFF
--- a/src/wrapper/objects/jvalue.rs
+++ b/src/wrapper/objects/jvalue.rs
@@ -1,3 +1,4 @@
+use std::convert::TryFrom;
 use std::mem::transmute;
 
 use log::trace;
@@ -185,10 +186,32 @@ impl<'a> From<jboolean> for JValue<'a> {
     }
 }
 
+impl<'a> TryFrom<JValue<'a>> for jboolean {
+    type Error = Error;
+
+    fn try_from(value: JValue<'a>) -> Result<Self> {
+        match value {
+            JValue::Bool(b) => Ok(b),
+            _ => Err(Error::WrongJValueType("bool", value.type_name())),
+        }
+    }
+}
+
 // jchar
 impl<'a> From<jchar> for JValue<'a> {
     fn from(other: jchar) -> Self {
         JValue::Char(other)
+    }
+}
+
+impl<'a> TryFrom<JValue<'a>> for jchar {
+    type Error = Error;
+
+    fn try_from(value: JValue<'a>) -> Result<Self> {
+        match value {
+            JValue::Char(c) => Ok(c),
+            _ => Err(Error::WrongJValueType("char", value.type_name())),
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/wrapper/objects/jvalue.rs
```diff
@@ -185,58 +186,157 @@ impl<'a> From<jboolean> for JValue<'a> {
...
+}
+
 // jvoid
 impl<'a> From<()> for JValue<'a> {
     fn from(_: ()) -> Self {
         JValue::Void
     }
 }
+
+impl<'a> TryFrom<JValue<'a>> for () {
```
dmitry-timofeev, line 333: Is `TryFrom` implementation needed for `()`? Are there any cases when a user _expects_ Void and needs to do try_from to get void — essentially, nothing?

dmitry-timofeev (changes requested):
Thanks, the code lgtm. Would you please update the changelog?

We usually require automated tests for new functionality when possible/sensible. Do you think it is "too simple to break case" (@stanislav-tkach , WDYT?)? Please do not remove PR templates as it helps both the author and the reviewers.

giovanniberti:
Whoops, my bad, sorry about the template!

I'll update the changelog 👍

dmitry-timofeev:
No problem at all! Thanks!

src/wrapper/objects/jvalue.rs
giovanniberti: It's not really needed, it's just for... symmetry?
The only way I think it could be useful is to check for Java exceptions when calling a method in a generic way, even for void-returning methods, but I understand it may not be that useful.

stanislav-tkach (approved):
I think that in this case the types prevent all possible errors except for error messages, so we can omit tests, but personally I would still add them.

src/wrapper/objects/jvalue.rs
giovanniberti: To be more concrete: I could easily write something like (not actual code)
```rust
struct AJavaGenericFunction; // closure-like struct but for a static function, so no fields

impl JavaCall {
    fn<T, U, 'a> call_generic_fn(&self, arg: T) -> Result<U> where T: Into<JValue<'a>>,  U: Into<JValue<'a>> + SomeOtherTrait {
        let env = self.get_env();
        env.call_static_method_unchecked(self.into(), self.get_fn_method_id(), U::get_java_type(), &[arg.into()]).map(TryFrom::try_from)
    }
}

// some lines below
let arg: jint = 3;
let v: jint = java_call.call_generic_fn(arg)?;
```


Which would make managing types much more hassle free by having a universal type-conversion utility function.

Actually it's [something I'm looking into automating](https://github.com/giovanniberti/robusta) via a derive macro (not in the form outlined above, it's just to offer some plausible use-case for the `Void` impl)

dmitry-timofeev: I see, thanks for the explanation. I do not object to adding it to enable more concise client-side code.

dmitry-timofeev (approved):
LGTM, thanks!

PR #266, opened 2020-10-01
Introduce WeakGlobalRef

### LINKED ISSUE #167: Support for weak globals refs
This library is awesome! Using it to add Android support for https://github.com/vislyhq/stretch :) One thing that I found missing was the ability to create weak references (NewWeakGlobalRef & DeleteWeakGlobalRef JNI functions, see https://docs.oracle.com/javase/7/docs/technotes/guides/jni/spec/functions.html)

Is there any reason I should know about as to why these have not been implemented? I'm happy to give a go at implementing this. I would probably try to start by basically copy-pasting all the GlobalRef code and prefixing it with 'Weak', and ensuring to call the correct underlying JNI code. Does that sounds about right?

maxv-rust:
## Overview
Introduce WeakGlobalRef. Closes #167

* WeakGlobalRef allows user to keep weak reference on the underlying Java object.
* The WeakGlobalRef may be created via JNIEnv::new_weak_global_ref.
* JNIEnv::upgrade_weak_global_ref may be used to obtain GlobalRef and access the underlying object.

### Definition of Done

### DIFF
--- a/src/wrapper/jnienv.rs
+++ b/src/wrapper/jnienv.rs
@@ -375,7 +375,7 @@ impl<'a> JNIEnv<'a> {
         )
         .into();
 
-        let weak_global = unsafe { GlobalRef::from_raw(self.get_java_vm()?, new_ref.into_inner()) };
+        let global = unsafe { GlobalRef::from_raw(self.get_java_vm()?, new_ref.into_inner()) };
         Ok(weak_global)
     }
 
--- a/src/wrapper/jnienv.rs
+++ b/src/wrapper/jnienv.rs
@@ -14,7 +14,7 @@ use crate::{
     objects::{
         AutoByteArray, AutoLocal, AutoPrimitiveArray, GlobalRef, JByteBuffer, JClass, JFieldID,
         JList, JMap, JMethodID, JObject, JStaticFieldID, JStaticMethodID, JString, JThrowable,
-        JValue, ReleaseMode,
+        JValue, JWeak, ReleaseMode, WeakGlobalRef,
     },
     signature::{JavaType, Primitive, TypeSignature},
     strings::{JNIString, JavaStr},
@@ -345,6 +345,38 @@ impl<'a> JNIEnv<'a> {
             jni_unchecked!(self.internal, NewGlobalRef, obj.into().into_inner()).into();
         let global = unsafe { GlobalRef::from_raw(self.get_java_vm()?, new_ref.into_inner()) };
         Ok(global)
+    }
+
+    /// Turns an object into a weak global ref.
+    ///
+    /// `WeakGlobalRef` doesn't prevent the underlying object from been garbage
+    /// collected. It's not safe to access the underlying object via `WeakGlobalRef`.
+    /// Instead, consider using [upgrade_weak_global_ref](#method.upgrade_weak_global_ref) for
+    /// create `GlobalRef` and check the `GlobalRef` is not null.
+    pub fn new_weak_global_ref<O>(&self, obj: O) -> Result<WeakGlobalRef>
+    where
+        O: Into<JObject<'a>>,
+    {
+        let new_ref: JWeak =
+            jni_unchecked!(self.internal, NewWeakGlobalRef, obj.into().into_inner()).into();
+        let weak_global =
+            unsafe { WeakGlobalRef::from_raw(self.get_java_vm()?, new_ref.into_inner()) };
+        Ok(weak_global)
+    }
+
+    /// Try to upgrade a weak global ref into a global ref.
+    ///
+    /// Returned `GlobalRef` may be null if the underlying object was garbage collected.
+    pub fn upgrade_weak_global_ref(&self, weak_global_ref: &WeakGlobalRef) -> Result<GlobalRef> {
+        let new_ref: JObject = jni_unchecked!(
+            self.internal,
+            NewGlobalRef,
+            weak_global_ref.as_weak().into_inner()
+        )
+        .into();
+
+        let weak_global = unsafe { GlobalRef::from_raw(self.get_java_vm()?, new_ref.into_inner()) };
+        Ok(weak_global)
     }
 
     /// Create a new local ref to an object.
--- a/src/wrapper/jnienv.rs
+++ b/src/wrapper/jnienv.rs
@@ -376,7 +376,7 @@ impl<'a> JNIEnv<'a> {
         .into();
 
         let global = unsafe { GlobalRef::from_raw(self.get_java_vm()?, new_ref.into_inner()) };
-        Ok(weak_global)
+        Ok(global)
     }
 
     /// Create a new local ref to an object.
--- a/src/wrapper/objects/jweak.rs
+++ b/src/wrapper/objects/jweak.rs
@@ -1,0 +1,45 @@
+use std::marker::PhantomData;
+
+use crate::sys::jweak;
+
+/// Wrapper around `sys::jweak` that adds a lifetime. This prevents it from
+/// outliving the context in which it was acquired. It matches C's representation
+/// of the raw pointer, so it can be used in any of the extern function argument
+/// positions that would take a `jweak`.
+///
+/// See also: [WeakGlobalRef](./struct.WeakGlobalRef.html)
+#[repr(transparent)]
+#[derive(Clone, Copy, Debug)]
+pub struct JWeak<'a> {
+    internal: jweak,
+    lifetime: PhantomData<&'a ()>,
+}
+
+impl<'a> From<jweak> for JWeak<'a> {
+    fn from(other: jweak) -> Self {
+        Self {
+            internal: other,
+            lifetime: PhantomData,
+        }
+    }
+}
+
+impl<'a> ::std::ops::Deref for JWeak<'a> {
+    type Target = jweak;
+
+    fn deref(&self) -> &Self::Target {
+        &self.internal
+    }
+}
+
+impl<'a> JWeak<'a> {
+    /// Unwrap to the internal jni type.
+    pub fn into_inner(self) -> jweak {
+        self.internal
+    }
+
+    /// Creates a new null object
+    pub fn null() -> Self {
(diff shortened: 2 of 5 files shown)


stanislav-tkach:
Thanks for the pull request and sorry for the late replay. I will try to review it soon.

src/wrapper/jnienv.rs
```diff
@@ -347,6 +347,38 @@ impl<'a> JNIEnv<'a> {
...
+            jni_unchecked!(self.internal, NewWeakGlobalRef, obj.into().into_inner()).into();
+        let weak_global =
+            unsafe { WeakGlobalRef::from_raw(self.get_java_vm()?, new_ref.into_inner()) };
+        Ok(weak_global)
+    }
+
+    /// Try to upgrade a weak global ref into a global ref.
+    ///
+    /// Returned `GlobalRef` may be null if the underlying object was garbage collected.
+    pub fn upgrade_weak_global_ref(&self, weak_global_ref: &WeakGlobalRef) -> Result<GlobalRef> {
```
stanislav-tkach, line 370: What do you think about returning `Result<Option<GlobalRef>>` instead of `Result<GlobalRef>`? This way `GlobalRef` will always be non-null and the api becomes a little more obvious.

src/wrapper/jnienv.rs
```diff
@@ -347,6 +347,38 @@ impl<'a> JNIEnv<'a> {
...
+        let new_ref: JObject = jni_unchecked!(
+            self.internal,
+            NewGlobalRef,
+            weak_global_ref.as_weak().into_inner()
+        )
+        .into();
+
+        let weak_global = unsafe { GlobalRef::from_raw(self.get_java_vm()?, new_ref.into_inner()) };
```
stanislav-tkach, line 378: Nitpick: at this point it is no longer weak, right?

src/wrapper/objects/jweak.rs
```diff
@@ -0,0 +1,45 @@
...
+}
+
+impl<'a> JWeak<'a> {
+    /// Unwrap to the internal jni type.
+    pub fn into_inner(self) -> jweak {
+        self.internal
+    }
+
+    /// Creates a new null object
+    pub fn null() -> Self {
```
stanislav-tkach, line 42: I'm not sure about this. Personally I prefer to use `Option<JWeak>` instead of a "nullable" object if I need a potentially absent value. In that case it is immediately clear if the value needs to be checked or not.

maxv-rust: I'm trying to mimic [JObject::null](https://docs.rs/jni/0.17.0/jni/objects/struct.JObject.html#method.null) there. 

Should I remove `JWeak::null` and change `JWeak::from` to return `Option<JWeak>`? Then at `JWeak::from` we can check if `other: jweak` is `null` or not. Then we can guarantee that `JWeak` is not `null` for all the cases.

src/wrapper/jnienv.rs
maxv-rust: I agree we should avoid null-checks. From the other side Result<Option<GlobalRef>> looks a little bit too verbose. 

Do you think it's possible to use one of existing [ErrorKind](https://docs.rs/jni/0.17.0/jni/errors/enum.ErrorKind.html)s to provide meaningful error as a part of `Result<GlobalRef>`. It'll allow us to keep the method signature `pub fn upgrade_weak_global_ref(&self, weak_global_ref: &WeakGlobalRef) -> Result<GlobalRef>`. Or we can create a new one to cover that case.

What do you think?

tests/jni_weak_refs.rs
```diff
@@ -0,0 +1,84 @@
...
+use jni::{
+    objects::{AutoLocal, JValue},
+    sys::jint,
+};
+
+mod util;
+use util::{attach_current_thread, unwrap};
+
+#[test]
+pub fn global_ref_works_in_other_threads() {
```
dmitry-timofeev, line 17: I'd suggest a couple of simpler tests to verify the basic functionality — this one is rather complex:

1. Create a WeakGlobalRef (and a GlobalRef, to keep the object around), detach the current thread, re-attach it, and verify it can be upgraded to strong local and global references.
2. Create a WeakGlobalRef from `JObject::null`.
  - :warning: A design question: shall JNIEnv create a WeakGlobalRef that can never be upgraded, or return Option::none?
3. (A more advanced test, similar to the ones we use for GlobalRefs) Create a VM with small memory limit (e.g., 32MB), create a WeakGlobalRef to a large-ish object (e.g., 16MB), release any strong refs to it, create a different large-ish object, so that the first is GCed (e.g., 16MB will certainly do the job), verify the WeakGlobalRef cannot be upgraded.
4. (May keep this one) Test it is accessible from different threads.

src/wrapper/jnienv.rs
dmitry-timofeev: I can't comment on the ergonomics in Rust, but Option<GlobalRef> may relieve the users from comparing the returned reference to `null`. The thing is, `null` GlobalRef/AutoLocal is usually a possible thing, but WeakGlobalRef are not really supposed to be `null`. We may consider distinguishing between "WGR was _originally_ null" (but maybe, we shall rather make such case _non-representable_ by returning an Option<WGR> from `new_weak_global_ref`?) and "WGR wasn't null, but the object was GCed". Alternatively, we may keep things as in JNI, where WGR could be `null`, and you can get `null` in `NewLocalRef`/`NewGlobalRef` which could mean two things (WGR originally null or the object got GCed).

dmitry-timofeev (changes requested):
Thanks, I think we shall figure out a couple of design questions:
- How to handle `null` passed to `new_wgr`
- The best API to upgrade WGR to local or global refs 
  - Method location
  - How to handle GCed WGRs in terms of the return type
- Do any users need JWeak?

dmitry-timofeev:
> Thanks, I think we shall figure out a couple of design questions:

Do I understand correctly that most usages of wgr come in the following form:

```rust
// Initialize wtf

weakFoo = env.new_weak_global_ref(initial)

// Use, in other place
foo = weakFoo.get_as_local()
if (isNull(foo)) {
   // foo was GCed, recompute it
   foo = recompute(env)
   // Update the wtr
   weakFoo = env.new_weak_global_ref(foo)
}
use(foo)
```

If so, would it make sense to provide a type that does that for you? Something like 

```rust
// Initialize smart wtf
// supplier is ~ Function<JNIEnv, JObject>
weakFoo = env.new_smart_weak_global_ref(supplier)
  /* Optional inital value, if user already has one */ 
  .initial_value(initial)
  .create();

// Use — it will get the value from supplier if it got GCed automatically
foo = weakFoo.get_as_local();
use(foo)
```

I do not suggest to do that as part of this PR, but if it is the most common (or the only) pattern — it might make sense to add such a type.

maxv-rust:
@dmitry-timofeev , @stanislav-tkach 

Thanks for the review. Sorry for the late response.

Unfortunately I don't have enough time to jump in into the conversation about the right architecture for the feature.

I'll be happy to update my review whenever you made final decision how the interface should looks like.

My original toughs was to follow GlobalRef architecture as much as possible. Unfortunately it may be not the best solution for the WeakGlobalRef.

src/wrapper/objects/weak_global_ref.rs
```diff
@@ -0,0 +1,97 @@
...
+
+/// A weak global JVM reference. **It doesn't protect underlaying object from been
+/// garbage collected.** Still, `WeakGlobalRef` is allowed to outlive the `JNIEnv` that
+/// it came from and can be used in other threads.
+///
+/// `WeakGloablRef` may be created via
+/// [JNIEnv::new_weak_global_ref](../struct.JNIEnv.html#method.new_weak_global_ref).
+///
+/// `WeakGloablRef` doesn't allow access to the underlying object but may be
+/// upgraded into _potentially null_ [GlobalRef](struct.GlobalRef.html) via
```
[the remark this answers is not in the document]
maxv-rust: I don't believe it's a good idea to use JavaVM because the thread may not be attached to the JavaVM. 

We're forcing thread attaching for Drop because we don't have any other option. For all other purposes it's User's responsibility to provide correct JNIEnv.

I have no objections as per creation the methods for WeakGlobalRef instead of JNIEnv. But the methods should require JNIEnv to be provided as a parameter.

What do you think?

src/wrapper/objects/jweak.rs
```diff
@@ -0,0 +1,45 @@
+use std::marker::PhantomData;
+
+use crate::sys::jweak;
+
+/// Wrapper around `sys::jweak` that adds a lifetime. This prevents it from
+/// outliving the context in which it was acquired. It matches C's representation
+/// of the raw pointer, so it can be used in any of the extern function argument
+/// positions that would take a `jweak`.
```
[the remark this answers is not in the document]
maxv-rust: Theoretically it may be useful to interact with others JNI code/libraries which are not based on jni-rs (or even Rust).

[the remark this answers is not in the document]
dmitry-timofeev: Possibly, in that case, would `jweak` returned from WeakGlobalRef be enough (see also https://github.com/jni-rs/jni-rs/pull/266/files/028d6c1386eb9f78413cd10c92c9f0b61fa5a448#r504867022)?

src/wrapper/objects/weak_global_ref.rs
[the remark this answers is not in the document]
dmitry-timofeev: For locals, having the current thread attached is the only way to do it; for globals that is a sensible way, because thread attachment is expensive. 

My thought was to get the env of an already attached thread via JavaVM#get_env, and failing in case it is not.

I think the benefit of passing JNIEnv is that it is more explicit that it is needed; the benefit of taking it under the hood is that the methods require less arguments. But there is a cost with the 2nd approach: there are more failure modes, and it can be used incorrectly (without JNIEnv). I lean towards passing an argument.

rib:
Okey, closing this now that `jni-rs` now has support for weak references since merging #304

PR #272, opened 2020-10-14
Add support for accessing long array elements

s1ck:
## Overview

When using the crate, I couldn't find a way of accessing long array elements in a convenient away. I had a look at `JNIEnv#get_byte_array_elements` and figured that we can add a similar approach for accessing long arrays.

### Definition of Done


s1ck:
@dmitry-timofeev I'd appreciate a `hacktoberfest-accepted` label for the PR. I need a new t-shirt :)

dmitry-timofeev:
Thanks @s1ck ! 

There are [two main ways](https://docs.oracle.com/en/java/javase/11/docs/specs/jni/design.html#accessing-primitive-arrays) to access an array in JNI: copying & "pinning". Copying is an order of magnitude easier to implement for more JNI types (just a single function, see [get_byte_array_region](https://docs.rs/jni/0.17.0/jni/struct.JNIEnv.html#method.get_byte_array_region), compared to three functions + `AutoXArray`). There is already a function for long arrays: `get_long_array_region`, it is possible to add `convert_long_array` for convenience, just like `convert_byte_array`. I am reluctant to add pinned access to other array types, unless there is explicit request for them.

If there _is_ a demand for pinned access, I'd first consider:
- Do we need to keep "dumb" methods public? I see that for bytes we have both "dumb" JNI-like methods and AutoByteArray. There is a discussion in the PR, but I do not see an explicit decision/considerations on why we need the dumb methods when there is `AutoByteArray`: https://github.com/jni-rs/jni-rs/pull/237#discussion_r442284097 
@stanislav-tkach , @maurolacy am I missing any considerations? Do you think it makes sense to keep the "dumb" methods? Maybe, to provide explicit control over the time the array is released, so that the native code does not hold it for longer than needed?
- Can we make a single generic AutoArray type? I see @maurolacy have attempted that, but it didn't work at the time: https://github.com/jni-rs/jni-rs/pull/237#discussion_r442284097
- Is it sensible to make a single pair of commit/release methods compatible with all array types?

maurolacy:
> Can we make a single generic AutoArray type? I see @maurolacy have attempted that, but it didn't work at the time: #237 (comment)

I don't think it's possible, because of what I mention in [#237](https://github.com/jni-rs/jni-rs/pull/237#discussion_r442284097).

In any case, you can use `AutoPrimitiveArray` for multiple types, as the related JNI method return / receive `void *`; they are therefore generic.

The problem is with the `Auto<Type>Array` types. It tried to write a generic container for them, but at some point you need to call the proper JNI API, which includes the type in its name :-(

So, maybe there's a way using macros, or some trick to map from types to functions, but at this point I don't know if and how it could be done.

maurolacy:
So, at @s1ck, you can try using the `get/release_primitive_array` methods, or the `AutoPrimitiveArray` container. It should work for `longs`; though I have not tried it.

maurolacy:
> Do we need to keep "dumb" methods public?

Are you referring to the `get_<type>_array_elements()` methods? In my opinion they should be kept, yes. They should also be completed / extended as required, like in this (@s1ck's) contribution.

It would be great to find a way to write a generic (`get_array_elements<T>()`) version of them, though.

s1ck:
@dmitry-timofeev @maurolacy Thanks for the input. For my use case, "pinning" is the desired access method. I'll try the suggested `get_auto_primitive_array_critical`. 
However, I still think it makes sense to have the "dumb" type-specific methods for convenience and - depending on the GC - there is a chance that the arrays are not copied. https://shipilev.net/jvm/anatomy-quarks/9-jni-critical-gclocker/ is also insightful on this topic.

For the PR: Do you want me to add the `AutoLongArray` pendant for completeness? We could play with a generic solution using the Byte and Long versions in a follow-up PR.

maurolacy:
Hi @s1ck,

> For the PR: Do you want me to add the `AutoLongArray` pendant for completeness? We could play with a generic solution using the Byte and Long versions in a follow-up PR.

My opinion is that, yes, you should go ahead and add the `AutoLongArray`.

And that, yes, we can play with a generic solution later, and eventually remove the redundant methods / structs. In fact I, was playing last night :-) but couldn't make it work yet. It's like trying to join two really loose and distant ends, by the way: generic types on one side, and raw pointers on the other.

maurolacy:
> I'll try the suggested `get_auto_primitive_array_critical`.

Please let us know how that goes.

dmitry-timofeev:
> For my use case, "pinning" is the desired access method

Could you tell us please what is the reason? Is it faster on your target VM vendor/VM version/GC combination (it doesn't copy and yields perf benefits)? @maurolacy did some measurements, though I am not sure on which VM — and this method is specified as VM-implementation specific, and there were no difference from the existing copying method. Or is it easier to use in your use-cases (e.g., updating an array)?

> It would be great to find a way to write a generic (get_array_elements<T>()) version, though.

I wonder if it could return a correct array type? Or would it require pattern-matching, like JValue? Would it be easier to have separate public methods to `get_array_elements` that reuse some internal generic methods/macros? 

> Are you referring to the get_<type>_array_elements() methods? In my opinion they should be kept, yes. They should also be completed / extended as required; like in this contribution.

What is the reasoning behind keeping _both_ "dumb" methods and AutoXArrays? I could only think of higher level of control — if someone needs to release an array earlier than it will leave its scope. That is valid in case of "critical" methods, which must be released ASAP, but doesn't seem so for normal "get_x_elements".

> However, I still think it makes sense to have the "dumb" type-specific methods for convenience and - depending on the GC - there is a chance that the arrays are not copied.

The question is if we should have both "dumb" methods and AutoXArrays — the latter are supposed to provide the "no-copy" behaviour of the former.

maurolacy:
> @maurolacy did some measurements, though I am not sure on which VM — and this method is specified as VM-implementation specific, and there were no difference from the existing copying method.

I benchmarked using OpenJDK 8, and you're right, the `get_X_array_elements()` didn't provide an advantage over methods that do simple copying (`get_X_array_region` and such). I guess basically because they were always copying the array under the hood (i. e. `is_copy` always returned `true`).

> > It would be great to find a way to write a generic (get_array_elements()) version, though.
> 
> I wonder if it could return a correct array type? Or would it require pattern-matching, like JValue? Would it be easier to have separate public methods to `get_array_elements` that reuse some internal generic methods/macros?

It could return the right type, by using the generic as return type. This will require some type anotations from the user, so type inference can work.

 I'll update this with the version of the code I wrote, and we can discuss it, and see if there's a path forward.
 
> The question is if we should have both "dumb" methods and AutoXArrays — the latter are supposed to provide the "no-copy" behaviour of the former.

Your are right. I think that it only makes sense from an usability / flexibility point of view. I would leave them, just to give more options to the users. But I'm OK with removing them too, if you think they should be removed.

maurolacy:
Here's my current `get_array_elements()` best version so far:

```
    pub fn get_array_elements<T>(&self, array: jarray) -> Result<(*mut T, bool)> {
        non_null!(array, "get_array_elements array argument");
        let ptr = jni_non_void_call!(self.internal, GetByteArrayElements, array, &mut is_copy);
        Ok((ptr as *mut T, is_copy == sys::JNI_TRUE))
    }
```

This compiles, but as you can see, it´s still calling `GetByteArrayElements`, so it's not complete.

I think this requires another macro (` jni_type_call!` or so), which takes the type (or the `type_name`, which is a `&str`) as a parameter, and build the `GetXArrayElements`. Don't know much about Rust macros yet, but it looks doable.

s1ck:
@dmitry-timofeev In my current use-case, I'm handling multiple primitive long arrays which can grow into the size of multiple GB. I'd like to avoid copying those. I'm also accessing them read only, which would also make copying and syncing obsolete. However, with openjdk 11, I see the same behavior that @maurolacy reported, which is having `is_copy` always set to true. As mentioned, I'll try the `_critical` methods for that particular use case.

I understand that `Auto*Array` is just another convenience for avoiding manual release, that's why i didn't add them in the first place. 

Are you suggesting to keep the PR as-is? I'm a bit confused reading both of your suggestions :)

maurolacy:
> 
> Are you suggesting to keep the PR as-is? I'm a bit confused reading both of your suggestions :)

@dmitry-timofeev, it's your call. As I said, I'm OK with both, keeping or removing the "dumb" methods.

dmitry-timofeev:
> This compiles, but as you can see, it´s still calling GetByteArrayElements. So, it's not complete.

Would accepting a type name to translate into the JNI method name work (like "Byte" in this case — I guess it requires turning the whole method into a macro)? It certainly won't work for user-facing methods, but if we stay with "separate method per type" option, I think this method/macro can still simplify their implementation to a one-liner. E.g., `get_array_elements(array, "Byte")`.

If a single method for all types is easier to use, I think we can have an internal piece of code performing pattern matching. However, if it would require extra arguments from users, then likely it won't be easier to use, and we will have to go with separate methods per array type.

> Your are right. I think that it only makes sense from an usability / flexibility point of view. I would leave them, just to give more options to the users. But I'm OK with removing them too, if you think they are redundant.

I'd ponder on that a little. AutoByteArray seems to provide all required functionality in a safer way (commit & release). I wonder in which contexts and scenarious separate methods may be useful. In a large-ish scope it is probably quite hard to ensure that they get called on all return paths (e.g., on every `?`), making them error-prone to use. In a small scope, there is probably no point in extra control (= explicit release) — it happens immediately at the end of the scope. These, however, are guesses — ideally, I'd like to hear more input on their usages IRL. What do you already use, or would like to use?

Also, are AutoByteArrays as easy to use as what you get from a dumb method (i.e., we will not inconvenience users by hiding "dumb" methods)? 

> Are you suggesting to keep the PR as-is? I'm a bit confused reading both of your suggestions :)

I'd consider clarifying the questions above before making any big changes. If we decide to expose the methods, we may skip AutoLongArray for now and keep the patch as-is — unless you are already convinced you need it :) If we decide to expose only AutoLongArray, then the methods may be redundant.

> I'm handling multiple primitive long arrays which can grow into the size of multiple GB.

Those are quite big, beware the "critical" requirements and limitations. Are direct ByteBuffers/LongBuffers an option in your case? These can certainly be accessed in native without copying, but have their own issues (e.g., they are GCed through finalization, which does not work well for a large number of objects).

It also seems we will soon have a better API to access native memory from Java (you can try it already on JDK 15, but it is a _preview_ feature, hence, it may change): https://openjdk.java.net/jeps/383

maurolacy:
> Would accepting a type name to translate into the JNI method name work (like "Byte" in this case — I guess it requires turning the whole method into a macro)? It certainly won't work for user-facing methods, but if we stay with "separate method per type" option, I think this method/macro can still simplify their implementation to a one-liner. E.g., `get_array_elements(array, "Byte")`.

It would require turning the whole method into a macro, yes. I like this approach. It can also be easily extended to other similar use cases.

> Also, are AutoByteArrays as easy to use as what you get from a dumb method (i.e., we will not inconvenience users by hiding "dumb" methods)?

I think the "dumb" methods are somewhat easier to use; particularly for new users.

dmitry-timofeev:
>  I think the "dumb" methods are somewhat easier to use; particularly for new users.

I am looking at usages and our tests, and don't see it (look at release):      

https://github.com/jni-rs/jni-rs/blob/1d51a37d5c1591e13fcb8612a412e3ac6297807b/tests/jni_api.rs#L331-L352

vs 

```rust
    // Use a scope to test Drop
    {
        // Get byte array elements auto wrapper
        let auto_ptr = env
            .get_auto_byte_array_elements(java_array, ReleaseMode::CopyBack)
            .unwrap();

        // Check pointer access
        let ptr = auto_ptr.as_ptr();
        assert_eq!(unsafe { *ptr.offset(0) }, 1);
        assert_eq!(unsafe { *ptr.offset(1) }, 2);
        assert_eq!(unsafe { *ptr.offset(2) }, 3);

        // Modify
        unsafe {
            *ptr.offset(0) += 1;
            *ptr.offset(1) += 1;
            *ptr.offset(2) += 1;
        }

        // Commit/Release would be necessary here, if there were no closure
        //auto_ptr.commit();
    }
```

The only missing functionality that I see is that you can't change your decision on release mode after you've created an AutoByteArray (e.g., if my updates are bad — don't copy back). Probably, such gap can be closed if we add a method to release the array discarding any changes (Mode::NoCopyBack).

maurolacy:
I think you are right. I was talking more about a "cultural" issue, or factor: When I start with something, I usually look for a "getter". And then, if I find that it comes with an associated "release", I just handle it / deal with it.

But, I guess you are right: if we stop supporting those get/release idioms, and only provide the auto variants, people will just use them. And yes, they are simpler to use in the end, in that they do part of the work for you. Plus, they are also safer.

s1ck:
@dmitry-timofeev Looking at the two examples, I agree that the `Auto*Array` version looks nicer. 

However, when I first tried to find a way to work with long arrays, I checked the [JNI docs](https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/functions.html) and saw `GetLongArrayElements`. Next step was to find it in `rust-jni` which brought me to this PR :)

With my current knowledge, I'd use `Auto*Array`.

> It also seems we will soon have a better API to access native memory from Java (you can try it already on JDK 15, but it is a preview feature, hence, it may change): https://openjdk.java.net/jeps/383

I'm certainly looking forward to that :)

s1ck:
@dmitry-timofeev @maurolacy should I go ahead and replace the two methods with the `get_auto_long_array_elements`?

dmitry-timofeev:
Thanks @maurolacy , @s1ck  for sharing your feedback!

As `AutoByteArray` appears to be easier and safer to use, I suggest to add any missing pieces to it and deprecate "dumb" methods: #275 

> I checked the JNI docs and saw GetLongArrayElements

Would `get_auto_long_array_elements` be sufficiently discoverable? If we plan to deprecate and remove `get_byte_array_elements`, we can instead name all "auto" methods as "get_<type>_array_elements", with no "auto".

Regarding this PR, if you can make `get_long_array_elements` return `AutoLongArray` that would be wonderful!

maurolacy:
> Would `get_auto_long_array_elements` be sufficiently discoverable? If we plan to deprecate and remove `get_byte_array_elements`, we can instead name all "auto" methods as `get_<type>_array_elements`, with no "auto".

+1. I can work on the related issue  (#272) as soon as possible.

s1ck:
> Would get_auto_long_array_elements be sufficiently discoverable? If we plan to deprecate and remove get_byte_array_elements, we can instead name all "auto" methods as get_<type>_array_elements, with no "auto".

I think that solves the issue of discoverability. Nice!

s1ck:
@dmitry-timofeev @maurolacy I added the `AutoLongArray`, but I'm running into some issues in terms of visibility. My first idea was to move `[get,release|commit]_long_array` as private methods into the `AutoLongArray`. This doesn't work since they require access to the `internal` field in `JNIEnv`. My second approach was to leave the methods on `JNIEnv`, but limiting their visibility to `pub(crate)`. This works, but they are not visible for the API tests, which makes sense, since they aren't public API. Since we don't wanna loose the tests for those methods, I thought about moving them into a test module within `JNIEnv`. Wdyt?

I also had a look at #273 which looks really nice. Could we implement the `release` and `commit` methods in a similar way and figure out a way to generate the `Auto*Array` types?

maurolacy:
> My first idea was to move `[get,release|commit]_long_array` as private methods into the `AutoLongArray`. This doesn't work since they require access to the `internal` field in `JNIEnv`.

Take a look at  `get_native_interface()`, to access to the internal field from the outside.
  
> My second approach was to leave the methods on `JNIEnv`, but limiting their visibility to `pub(crate)`. This works, but they are not visible for the API tests, which makes sense, since they aren't public API. Since we don't wanna loose the tests for those methods, I thought about moving them into a test module within `JNIEnv`.

I think you have two options here:

1. Implement `*_long_array` in exactly the same way `*_byte_array` is currenty being implemented. That is, with public manual methods and all. This is the simplest option, as the access rules will just work.
2. Go a little further, and implement only `get_auto_long_array`, and the `AutoLongArray` struct. That will require putting the functionality inside `AutoLongArray` (you can use `self.env.get_native_interface()` when needed). This is the logical thing to do, as we're planning to deprecate the manual methods.
In fact, I'm working on that deprecation for `*_byte_array` manual methods so, if you wait a little you can take a look at my PR, and borrow from there.

> I also had a look at #273 which looks really nice. Could we implement the `release` and `commit` methods in a similar way and figure out a way to generate the `Auto*Array` types?

I'm still also waiting for feedback on the latest iteration I wrote for this. My take is, that you can go ahead an implement  the specific version for `long`. We'll handle the generic version later (if ever). If we do it, i will no be a minor task, as it will be good to put the functionality into a macro, handle error messages well, etc.

maurolacy:
@s1ck, take a look at #276.

s1ck:
> Take a look at get_native_interface(), to access to the internal field from the outside.

Thanks!

dmitry-timofeev:
Agree with @maurolacy , let's remove the manual methods and make `AutoLongArray` consistent with `AutoByteArray` in #276. 

`pub (crate)` methods could also work, I think it is totally fine to test `AutoLongArray` only through public API (e.g., modify the existing tests to use `AutoLongArray`).

src/wrapper/jnienv.rs
```diff
@@ -2062,6 +2063,95 @@ impl<'a> JNIEnv<'a> {
...
+    }
+
+    /// Return an AutoLongArray of the given Java long array.
+    ///
+    /// The result is valid until the AutoLongArray object goes out of scope, when the
+    /// release happens automatically according to the mode parameter.
+    ///
+    /// Since the returned array may be a copy of the Java array, changes made to the
+    /// returned array will not necessarily be reflected in the original array until
+    /// release_array_elements() is called.
```
dmitry-timofeev, line 2141: Since the user does not see this method at all, would it make sense to reword to "until the array is dropped" or "until the array goes out of scope", whichever sounds more natural to Rust users? 

If yes, could you please check the wording in other AutoXArrays?

src/wrapper/jnienv.rs
```diff
@@ -2062,6 +2063,95 @@ impl<'a> JNIEnv<'a> {
...
+            self.internal,
+            ReleaseLongArrayElements,
+            array,
+            elems,
+            sys::JNI_COMMIT
+        );
+        Ok(())
+    }
+
+    /// Return an AutoLongArray of the given Java long array.
```
dmitry-timofeev, line 2134: Could you add please that if an array is `null`, an Error::NullPtr is returned (here and in other auto-arrays)?

src/wrapper/objects/auto_long_array.rs
```diff
@@ -0,0 +1,85 @@
...
+    env: &'b JNIEnv<'a>,
+}
+
+impl<'a, 'b> AutoLongArray<'a, 'b> {
+    /// Creates a new auto-release wrapper for a pointer-based long array
+    ///
+    /// Once this wrapper goes out of scope, `release_long_array_elements` will be
+    /// called on the object. While wrapped, the object can be accessed via
+    /// the `From` impl.
+    pub fn new(
```
dmitry-timofeev, line 29: `pub (crate)` (here and in other AutoXArrays)?

src/wrapper/objects/auto_long_array.rs
```diff
@@ -0,0 +1,85 @@
...
+    is_copy: bool,
+    env: &'b JNIEnv<'a>,
+}
+
+impl<'a, 'b> AutoLongArray<'a, 'b> {
+    /// Creates a new auto-release wrapper for a pointer-based long array
+    ///
+    /// Once this wrapper goes out of scope, `release_long_array_elements` will be
+    /// called on the object. While wrapped, the object can be accessed via
+    /// the `From` impl.
```
dmitry-timofeev, line 28: Given clients don't use this method directly, would it make sense to merge this doc into the struct level docs, that are visible?

src/wrapper/objects/auto_long_array.rs
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    /// Commits the result of the array, if it is a copy
+    pub fn commit(&mut self) {
+        let res = self
+            .env
+            .commit_long_array_elements(*self.obj, unsafe { self.ptr.as_mut() });
+        match res {
+            Ok(()) => {}
+            Err(e) => debug!("error committing long array: {:#?}", e),
```
dmitry-timofeev, line 57: Given it is an unexpected result, I'd use 'error' so that the clients are more likely to see and investigate. `debug` logs are usually disabled for libraries, don't they?

src/wrapper/objects/auto_long_array.rs
```diff
@@ -0,0 +1,85 @@
...
+impl<'a, 'b> Drop for AutoLongArray<'a, 'b> {
+    fn drop(&mut self) {
+        let res = self.env.release_long_array_elements(
+            *self.obj,
+            unsafe { self.ptr.as_mut() },
+            self.mode,
+        );
+        match res {
+            Ok(()) => {}
+            Err(e) => debug!("error releasing long array: {:#?}", e),
```
dmitry-timofeev, line 76: error (as above)?

src/wrapper/objects/auto_long_array.rs
```diff
@@ -0,0 +1,85 @@
...
+        match res {
+            Ok(()) => {}
+            Err(e) => debug!("error committing long array: {:#?}", e),
+        }
+    }
+
+    /// Indicates if the array is a copy or not
+    pub fn is_copy(&self) -> bool {
+        self.is_copy
+    }
```
dmitry-timofeev, line 64: Could you please add `discard` as in #276 ?

tests/jni_api.rs
```diff
@@ -413,6 +413,109 @@ pub fn get_byte_array_elements_auto() {
...
+    assert_eq!(res[0], 2);
+    assert_eq!(res[1], 3);
+    assert_eq!(res[2], 4);
+}
+
+#[test]
+pub fn get_long_array_elements_auto() {
+    let env = attach_current_thread();
+
+    // Create original Java array
```
dmitry-timofeev, line 464: Duplicate

tests/jni_api.rs
```diff
@@ -413,6 +413,109 @@ pub fn get_byte_array_elements_auto() {
...
+    let java_array = env
+        .new_long_array(3)
+        .expect("JNIEnv#new_long_array must create a java array with given size");
+
+    // Insert array elements
+    let _ = env.set_long_array_region(java_array, 0, buf);
+
+    // Use a scope to test Drop
+    {
+        // Get byte array elements auto wrapper
```
dmitry-timofeev, line 476:
```suggestion
        // Get long array elements auto wrapper
```

tests/jni_api.rs
```diff
@@ -413,6 +413,109 @@ pub fn get_byte_array_elements_auto() {
...
+    // Confirm modification of original Java array
+    let mut res: [i64; 3] = [0; 3];
+    env.get_long_array_region(java_array, 0, &mut res).unwrap();
+    assert_eq!(res[0], 2);
+    assert_eq!(res[1], 3);
+    assert_eq!(res[2], 4);
+}
+
+#[test]
+pub fn get_long_array_elements_auto() {
```
dmitry-timofeev, line 461: Can we also test explicit `commit` (with an assumption that an array is a copy)?

dmitry-timofeev (commented):
Great job, thank you! A couple of suggestions below.

s1ck:
@dmitry-timofeev @maurolacy  Thanks for the review and comments!

tests/jni_api.rs
```diff
@@ -410,6 +410,108 @@ pub fn get_byte_array_elements_auto() {
...
+        .expect("JNIEnv#new_long_array must create a java array with given size");
+
+    // Insert array elements
+    let _ = env.set_long_array_region(java_array, 0, buf);
+
+    // Get long array elements auto wrapper
+    let mut auto_ptr = env
+        .get_auto_long_array_elements(java_array, ReleaseMode::CopyBack)
+        .unwrap();
+
```
dmitry-timofeev, line 487: As the array may be a direct pointer (according to the spec, we don't know of VM vendor/version/GC combinations where it happens), I think we shall check if it is a copy, and if not — return early (or, if Rust has an [`Assumptions`](https://junit.org/junit5/docs/current/api/org.junit.jupiter.api/org/junit/jupiter/api/Assumptions.html) alternative, use them).

dmitry-timofeev (approved):
Great job! LGTM!

tests/jni_api.rs
s1ck: Makes sense. A quick search didn't show any `assume`, I added a simple condition and return early if it's not a copy.

s1ck:
@dmitry-timofeev ready for merge, I guess :)

dmitry-timofeev (approved):

dmitry-timofeev:
@s1ck Thank you very much for contributing! 

@stanislav-tkach merged as it's very similar to AutoByteArray, please review when you have time.

PR #273, opened 2020-10-15
Add generic get_array_elements()

maurolacy:
## Overview

This is proof-of-concept code for implementing a "generic" `get_array_elements()` method. This is the result of a series of comments / talks, in #272 and #237.

If you think this has merit, it's interesting, or useful, I can follow up with implementing the rest of the methods, i.e. `release/ commit_array_elements`, and the `auto` versions.

I don't particularly like the approach used here, but pulling this anyway, because:
- There can be better ways to do it, and we can maybe realise them here.
- This can still be useful, in the sense that it can be converted into a macro, and used in other parts of the code base, to simplify not only these, but also similar methods.

Please let me know what you think.

### Definition of Done


src/wrapper/jnienv.rs
```diff
@@ -1973,6 +1974,43 @@ impl<'a> JNIEnv<'a> {
...
+    /// Return a tuple with a pointer to elements of the given Java array as first element.
+    /// The tuple's second element indicates if the pointed-to array is a copy or not.
+    ///
+    /// The result is valid until the corresponding release_array_elements() function is
+    /// called. Since the returned array may be a copy of the Java array, changes made to the
+    /// returned array will not necessarily be reflected in the original array until
+    /// release_array_elements() is called.
+    ///
+    /// See also [`release_array_elements`](struct.JNIEnv.html#method.release_array_elements)
+    pub fn get_array_elements<T>(&self, array: jarray) -> Result<(*mut T, bool)> {
```
dmitry-timofeev, line 1986: Is such method easier to use in client code than a specialized? Do you have to specify the type as 
```
foo: (*mut u8, bool) = get_array_elements(array)?;
```
?

dmitry-timofeev (commented):
Thanks a lot, that's a good experiment!

src/wrapper/jnienv.rs
maurolacy: You can still do something like
```
let (ptr, is_copy): (*mut jbyte, _) = env.get_array_elements(java_array)?;
```
for some "clarity" and robustness. But yes, not the best user experience here.

maurolacy: **Update**:
An clearer alternative, using explicit type annotation:
```
let (ptr, is_copy) = env.get_array_elements::<jbyte>(java_array)?;
```

dmitry-timofeev: Thanks, yeah, both options seem more verbose than a specialized method `get_byte_array_elements`. One possible benefit is that it can cast JNI types to Rust types (e.g., jbyte -> u8), though I don't know if we shall support that.

maurolacy: > it can cast JNI types to Rust types (e.g., jbyte -> u8), though I don't know if we shall support that.

That always happens. "JNI types" are just "types aliases" of Rust types (see [jni-sys](https://github.com/sfackler/rust-jni-sys/blob/550c6da1d1e6c86761bc9ef8e9fe5ec1ceaa9e78/src/lib.rs#L10)). Even with the specialized functions, if you take a look at the result's type, it's always a Rust type.
And as you can see, nothing prevents you from using the "JNI types".

My opinion is that this second syntax in particular, based on turbofish, is good / clear enough. Another benefit is implicit type resolution, like in [this usage in jnienv.rs](https://github.com/maurolacy/jni-rs/blob/6cb87706730388c5b64b77b90f6b2237e621126e/src/wrapper/jnienv.rs#L2088).

The issue here, in my opinion, is more one of low-level implementation... we can save some code repetition and methods but, do we want to do a match on type name as part of these functions?

And moreover: Is there a better way to do it (using traits by example), that we are missing? I for one am not happy with converting the type to a string, and doing a match on that.

maurolacy: Regarding performance, I realize now that despite its complexity, **all of this will be resolved at compile time**; because types are known / resolved at compile time! (this is obvious, I know; it just took me some time to realize here).

I've implemented a variant using traits but, it's not worth it. I now. also have a clunky impl based on `TypeId::of::<T>()`; but I think using `type_name()` with a `match` is better / clearer.

So, that's it! I don't think there is a better way to do it than the current one. It is for you to decide its worth, and if you want those constructs in the lib.

src/wrapper/jnienv.rs
```diff
@@ -1973,20 +1986,46 @@ impl<'a> JNIEnv<'a> {
...
-    /// See also [`release_byte_array_elements`](struct.JNIEnv.html#method.release_byte_array_elements)
-    pub fn get_byte_array_elements(&self, array: jbyteArray) -> Result<(*mut jbyte, bool)> {
-        non_null!(array, "get_byte_array_elements array argument");
+    /// See also [`release_array_elements`](struct.JNIEnv.html#method.release_array_elements)
+    pub fn get_array_elements<T: JType>(&self, array: jarray) -> Result<(*mut T, bool)> {
+        non_null!(array, "get_array_elements array argument");
         let mut is_copy: jboolean = 0xff;
-        let ptr = jni_non_void_call!(self.internal, GetByteArrayElements, array, &mut is_copy);
-        Ok((ptr, is_copy == sys::JNI_TRUE))
+        let type_name = type_name::<T>();
```
stanislav-tkach, line 2006: According to the [documentation](https://doc.rust-lang.org/std/any/fn.type_name.html), type name representation isn't specified, so we shouldn't rely on the specific values.

stanislav-tkach (commented):
At first I really liked the idea of having this generic implementation that can be used by specialized methods, but now I think that this approach doesn't really help to reduce the code-base and even adds some trait-related clutter.

src/wrapper/jnienv.rs
maurolacy: I agree. There's also `TypeId::of::<T>()`, which is more robust (types are compared to types). But, it cannot be used in a `match`... we're left with a chain of `if`s, which is anything but elegant.

The only options I can think of are:
- Put the version with the chain of `if`s into a macro, and try to use it in different parts of the code base. I'll push that version, so you can review it.
- Don't do anything, and resort to writing explicit code on demand (!).

maurolacy:
> At first I really liked the idea of having this generic implementation that can be used by specialized methods, but now I think that this approach doesn't really help to reduce the code-base and even adds some trait-related clutter.

That trait (`JType`) is optional, really. The default in the `match` is enough for safety in my opinion. I'm just exploring the alternatives.

src/wrapper/jnienv.rs
maurolacy: OK, I've removed the trait, and switched to `TypeId::of`. Please take a look at [the last commit](https://github.com/jni-rs/jni-rs/pull/273/commits/d119785ab29d02fcceec9fd06d87d980fd65d557).

maurolacy: OK, I've implemented `AutoArray<T>`, and used it to implement `get_auto_byte_array()`. Please take a look when you can.

dmitry-timofeev:
Thanks @maurolacy for the updates! 

I currently have two questions that I think we shall address before moving forward:
1. Is the present approach to supporting AutoArrays fine (a conditional statement put inside generic AutoArray)? I don't have any concerns, but possibly @stanislav-tkach has, or can suggest better approaches. In terms of alternative approaches, I can only think of making AutoArray a macro, or a generic struct that is parameterized with a single lambda releasing the array.
2. How we expose this to users:
    - As a single method `get_array_elements`. 
    - As type-specific methods `get_byte_array_elements` & friends.

Regarding the second question, do I understand correctly that a single method, in its current implementation, allows such usages: `get_elements<u8>` (from the reader's perspective, it is unclear if the array is expected to be `boolean` or `byte`)? Can we forbid such usages? Or simply recommend/support only JNI types: `get_elements<jboolean>` and `get_elements<jbyte>`?

maurolacy:
>     2. How we expose this to users:
>        
>        * As a single method `get_array_elements`.
>        * As type-specific methods `get_byte_array_elements` & friends.
> 
> 
> Regarding the second question, do I understand correctly that a single method, in its current implementation, allows such usages: `get_elements<u8>` (from the reader's perspective, it is unclear if the array is expected to be `boolean` or `byte`)?

Yes, that's correct.

> Can we forbid such usages?

No, not that I know of. These are just type aliases, so I don't think you can restrict their usage.

> Or simply recommend/support only JNI types: `get_elements<jboolean>` and `get_elements<jbyte>`?

That's something we should document, yes. JNI types should be used in examples and unit tests, etc.
Also, the only effect using the Rust types has, is potential confusion. What I mean is that calling these functions with the Rust type wouldn't cause any trouble; because these types are in fact already being used. If you take a look at the return value of a `get_byte_array_elements()` by example, you'll see that it's already `u8`, not `jbyte`.

Regarding writing a single generic method, vs. type specific methods, I think writing the type specific methods would be better, both in terms of clarity (less confusion with the types), and simplicity of use (no need to use turbofish, or type annotations).

The big advantage of an `AutoArray<T>`, as I see it, is that we don't have to replicate these structs to support different types. That is, we don't have to write the same struct many times, just changing a couple of lines here and there every time.

maurolacy:
>        * As a single method `get_array_elements`.
>        * As type-specific methods `get_byte_array_elements` & friends.

We can also provide **both** variants, the generic and the specific. And, use the one to implement the others.

src/wrapper/objects/auto_array.rs
```diff
@@ -0,0 +1,193 @@
+use log::debug;
+
+use crate::{errors::*, objects::JObject, sys, JNIEnv};
+use combine::lib::any::TypeId;
```
stanislav-tkach, line 4:
```suggestion
use std::any::TypeId;
```
`conbine` only reexports `TypeId` from `std, right?

stanislav-tkach:
I like the current api state. :+1: 

I'm not sure about all those `TypeId` related implementation details, though. Now I think that a helper trait would be better. Something like that:
```rust
// Perhaps some better name should be used.
trait AutoArrayImpl {
    fn get() -> NonNull<Self>;
    fn release(ptr: NonNull<Self>, mode: ReleaseMode);
}

impl AutoArrayImpl for jlong {
    fn get(env: &JNIEnv<'a>) -> NonNull<Self> {
        jni_non_void_call!(env, GetLongArrayElements, *obj, &mut is_copy) as *mut T
    }

    fn release(env: &JNIEnv<'a>, ptr: NonNull<Self>, mode: ReleaseMode) {
        jni_void_call!(
            env,
            ReleaseLongArrayElements,
            *self.obj,
            ptr as *mut i64,
            mode
        );
    }
}

impl<'a, 'b, T: AutoArrayImpl> AutoArray<'a, 'b, T> {
    pub fn new(env: &'b JNIEnv<'a>, obj: JObject<'a>, mode: ReleaseMode) -> Result<Self> {
        Ok(AutoArray {
            obj,
            ptr: T::get(env),
            mode,
            is_copy: is_copy == sys::JNI_TRUE,
            env,
            type_id: TypeId::of::<T>(),
        })
    }
```
This approach will not be more concise, but it will not require to store `TypeId`. Additionally, the boilerplate can be reduced with macros.

maurolacy:
> I like the current api state. +1
> 
> I'm not sure about all those `TypeId` related implementation details, though. Now I think that a helper trait would be better. Something like that:
> 
> ```rust
> // Perhaps some better name should be used.
> trait AutoArrayImpl {
>     fn get() -> NonNull<Self>;
>     fn release(ptr: NonNull<Self>, mode: ReleaseMode);
> }
> 
> impl AutoArrayImpl for jlong {
>     fn get(env: &JNIEnv<'a>) -> NonNull<Self> {
>         jni_non_void_call!(env, GetLongArrayElements, *obj, &mut is_copy) as *mut T
>     }
> 
>     fn release(env: &JNIEnv<'a>, ptr: NonNull<Self>, mode: ReleaseMode) {
>         jni_void_call!(
>             env,
>             ReleaseLongArrayElements,
>             *self.obj,
>             ptr as *mut i64,
>             mode
>         );
>     }
> }
> 
> impl<'a, 'b, T: AutoArrayImpl> AutoArray<'a, 'b, T> {
>     pub fn new(env: &'b JNIEnv<'a>, obj: JObject<'a>, mode: ReleaseMode) -> Result<Self> {
>         let mut is_copy: jboolean = 0xff;
>         Ok(AutoArray {
>             obj,
>             ptr: T::get(env),
>             mode,
>             is_copy: is_copy == sys::JNI_TRUE,
>             env,
>             type_id: TypeId::of::<T>(),
>         })
>     }
> ```
> 
> This approach will not be more concise, but it will not require to store `TypeId`. Additionally, the boilerplate can be reduced with macros.

Yes, that's the alternative. OK, I'll do the impl and push it, and we can then decide if we want to "macrify" and merge it.

maurolacy:
Rebased from master (already pushed).

And, experimented with implementing the suggested changes, above.

Which is not pushed yet. I think some comments are in order first: Traits over primitive types really don't look like a good idea. We're basically "super classing" or so the primitive types. With the "side" effect that our primitive types aren't anymore: they are now our super-classed types.
Plus, they need to be public (we're returning arrays of them from our functions). That is, the **Trait** needs to be public. Which exposes its interface; which, in the end, breaks encapsulation. We could technically construct one of these "traited" types, and call their methods directly. Or, call them on one of our returned array elements, which is worse.

On the other hand, it doesn't work. At least, not fully... it works with `jbyte`, but now with `jlong`... which is weird beyond comprehension (at least for me). Byte array tests work, but long array tests are failing with: "FATAL ERROR in native method: Array element type mismatch in JNI".
And that, with "-Xcheck:jni" enabled... if I disable it, I don't get a JNI error, but the long array elements are not updated, and then the test fails with an assertion.

All right, I can publish this thing in a different branch or so, in case you want to take a look. But I really think traits (or at least, traits over primitive types), are not the way to solve this.

maurolacy:
In any case, here's the branch with the impl for reference:
[get_array_elements_type_trait](https://github.com/maurolacy/jni-rs/tree/get_array_elements_type_trait)

maurolacy:
I'm now working on a Traits based impl which uses composition over the original `AutoArray`.

But, it isn't looking good so far... Rust doesn't currently allow accessing fields in a trait (In case you're interested, see https://github.com/nikomatsakis/fields-in-traits-rfc and https://github.com/rust-lang/rfcs/pull/2393 for potential solutions to this).
So, I have to compose the common parts into the type-specific impls... and wrap all the common methods in order  to expose them... where's my plain old inheritance when I need it? :-)

It seems writing a big ugly macro for this will end up being the only acceptable solution.

stanislav-tkach:
As far as I can see, the `release_array_elements` method implementation in the `get_array_elements_type_trait` branch is incorrect: it uses the `ReleaseByteArrayElements` function, but `T::release` should be used instead. Perhaps this is the reason for the "FATAL ERROR in native method: Array element type mismatch in JNI" error.

I don't think that making the `TypeArray` trait public is an issue. There are some workarounds for making a trait "sealed" (forbidding downstream implementations), though I'm not sure it is worth the hustle in our case.

maurolacy:
> As far as I can see, the `release_array_elements` method implementation in the `get_array_elements_type_trait` branch is incorrect: it uses the `ReleaseByteArrayElements` function, but `T::release` should be used instead. Perhaps this is the reason for the "FATAL ERROR in native method: Array element type mismatch in JNI" error.

Of course! Thanks for this. After I saw it was working for byte, I totally forgot to implement `release_array_elements` properly. I've now oushed a fixed version.

> 
> I don't think that making the `TypeArray` trait public is an issue. There are some workarounds for making a trait "sealed" (forbidding downstream implementations), though I'm not sure it is worth the hustle in our case.

I don't like this impl much, but if you like it and think that traits over primitive types are OK, I'm in favour of merging it. It certainly solves the mutiple types issue in a more or less minimal way (in terms of code multiplication); and that without depending on `TypeId`.

There is still some stuff I have to polish here, like extending to other types, better naming, proper docs / comments, some ugly clippy warnings, and the like.

If you're OK with whis, I'll work on that, and create a new PR for this. What do you think?

stanislav-tkach:
Perhaps I feel the same way about `TypeId` as you about the trait. 🙃 I suppose I don't have really strong arguments, except that with the `TypeId` approach adding a new type can result in a runtime error if one extends the `new` function, but forgets about `release_array_elements`. It would result in the compile time error if the trait were used, though it is highly unlikely that JNI api would be extended with new basic types.

maurolacy:
Fair enough. Use of `TypeId` is discouraged. It basically implies you're not being able to use traits / the type system efficiently ;-)

I'll close this, and create a new shiny new PR for the "type traits" branch.

maurolacy:
Closing in favour of #287.

maurolacy:
> So, I have to compose the common parts into the type-specific impls... and wrap all the common methods, in order to expose them... where's my plain old inheritance when I need it most? :-)

Turns out I was wrong! Today I learned there's something akin to inheritance in Rust, by abusing the `Deref` trait. See [Deref polymorphism](https://github.com/rust-unofficial/patterns/blob/master/anti_patterns/deref.md).

Issue #38, opened 2017-09-01
`JMethodID`/`JStaticMethodID` lifetime


stanislav-tkach:
Both `JMethodID` and `JStaticMethodID` have lifetime associated with `JNIEnv`, but it isn't entirely correct:

> A field or method ID does not prevent the VM from unloading the class from which the ID has been derived. After the class is unloaded, the method or field ID becomes invalid. The native code, therefore, must make sure to:
> - keep a live reference to the underlying class, or
> - recompute the method or field ID

This way method id should depend on class and not on `JNIEnv`. As far as I can see, there is no easy way to express that in the Rust code. In my code I use "hack" to workaround current restriction:
```rust
struct CachedConstructor {
    pub element_class: GlobalRef,
    // `static` lifetime is used here.
    pub constructor_id: JMethodID<'static>,
}

impl<InnerIter> CachedConstructor {
    pub fn new(env: &JNIEnv, class_name: &str) -> Result<Self> {
        let class = env.find_class(class_name)?;
        let element_class = env.new_global_ref(class.into())?;
        let id = env.get_method_id(class_name, "<init>", "([B[B)V")?;
        Ok(CachedConstructor {
            element_class,
            // Transmute & unsafe:
            constructor_id: unsafe { mem::transmute(id) },
        })
    }
}
```

Any advices on how to handle it better?

jechas01:
This is only a problem when you want to make the method/field id outlive the method it's obtained from, correct? It shouldn't be able to escape the extern method that where it first gets looked up, since the JNIEnv can't either. We might need something like you've got above in order to give out "Global" Method/FieldIDs.

koutheir:
> This is only a problem when you want to make the method/field id outlive the method it's obtained from, correct?

Correct.

> It shouldn't be able to escape the extern method that where it first gets looked up, since the JNIEnv can't either.

Actually, it can. The `JNIEnv` object obtained by a JNI function can be kept in thread-local storage, and it can be used as long as that thread is attached to the JVM. That thread can also cache identifiers of methods and fields.

cherryleafroad:
Is there a better way to cache these rather than having to use `std::mem::transmute()`? At the very least classes are cacheable without an issue by using `new_global_ref`, but I kinda feel hacky having to `std::mem::transmute()` on these ones

PR #274, opened 2020-10-15
Add Send and Sync implementations for JMethodID

stanislav-tkach:
### Definition of Done

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   specifying its name. The name is inferred from the class data. (#246)
 - `SetStatic<type>Field`. (#248)
 - `TryFrom<JValue>` for types inside JValue variants
-- `Send` and `Sync` implementations for `JMethodID` and `JFieldID`.
+- `Send` and `Sync` implementations for `JMethodID`, `JStaticMethodID`, `JFieldID` and
+  `JStaticFieldID`. (#274)
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   specifying its name. The name is inferred from the class data. (#246)
 - `SetStatic<type>Field`. (#248)
 - `TryFrom<JValue>` for types inside JValue variants
-- `Send` and `Sync` implementations for `JMethodID`.
+- `Send` and `Sync` implementations for `JMethodID` and `JFieldID`.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   specifying its name. The name is inferred from the class data. (#246)
 - `SetStatic<type>Field`. (#248)
 - `TryFrom<JValue>` for types inside JValue variants
+- `Send` and `Sync` implementations for `JMethodID`.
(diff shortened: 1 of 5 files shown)


dmitry-timofeev (commented):
1. There was also a consideration on too restrictive lifetimes (also in #38)
2. There are also JFieldIDs, which are the same.

dmitry-timofeev (commented):
There are also JStatic*IDs :)

dmitry-timofeev:
> There was also a consideration on too restrictive lifetimes (also in #38) 

Also, doesn't it make sense to extend the lifetimes? My understanding is that with such lifetimes it's not really easy to share them between threads.

stanislav-tkach:
> There was also a consideration on too restrictive lifetimes (also in #38)

I'm not sure how to correctly handle the lifetimes and preserve the ergonomics. As far as I can see, the lifetime should be bound to a class, so the the following definition is incorrect:
```rust
pub fn get_field_id<'c, T, U, V>(&self, class: T, name: U, sig: V) -> Result<JFieldID<'a>>
    where
        T: Desc<'a, JClass<'c>>,
        U: Into<JNIString>,
        V: Into<JNIString>,
    {
        // Now we have a class, but its lifetime is limited to this function, 
        // so `JFieldID` cannot outlive it either.
        let class = class.lookup(self)?;

        ...
```

The signature can be changed to the following:
```rust
pub fn get_field_id<'c, T, U, V>(&self, class: JClass<'c>, name: U, sig: V) -> Result<JFieldID<'c>>
    where
        U: Into<JNIString>,
        V: Into<JNIString>,
```
But in this case getting a field id would be more verbose:
```rust
// Current api:
let field_id = env.get_field_id("com/my/Class", "intField", "I");

// New api:
let class = "com/my/Class".lookup()?;
let field_id = env.get_field_id(class, "intField", "I");
```
Additionally, I'm not sure that this solves all lifetime issues:
```rust
let global_class;
let field_id;
{
    let class = "com/my/Class".lookup()?;
    field_id = env.get_field_id(class, "intField", "I");

    // Keep the class alive with a GlobalRef:
    global_class = new_global_ref(class)?;
}
// Here `class` goes out of scope, so Rust will complain about the `field_id` lifetime, 
// because it knows nothing about the class to global_class relation.
```
I don't see a clear solution here. Am I missing something?

dmitry-timofeev:
> As far as I can see, the lifetime should be bound to a class

Yes, the lifetime should be bound to a Java class. However, for the avoidance of doubt, it does not necessarily have to be bound to the class _native reference_ used to look up the method ID. The method ID won't go away just because a class native reference has been deleted; it will only go away if the class gets unloaded. Keeping any _native reference_ to the class object (`Class` instance) or to any instance of that class is enough to guarantee the method/field ID remains valid.

To prevent any possible window when the class can be unloaded, probably it's best to keep the same class _native reference_ that was used to create an ID.

Possibly, one solution is to marry IDs with class references, so that you can't keep an ID without a class reference. Would it require different types for global and local class refs?

The questions seem to be:
- How much safety are we willing to provide?
- If it is right to restrict sharing of these IDs across threads (allowed by the spec _if_ the class is kept alive — but that's the developer's responsibility) just because we can't guarantee safety?

Answers to that might help choose between:
- Keep as is (no Sync, no Send). Users have to fallback to unsafe to share across threads.
- Allow sharing, do not put safety measures. Keeping the class alive is on the user.
- Implement (class_ref + id) pairs, making unsafe usage impossible. Might need x2 types, to accomodate local and global class refs (local for cases when you keep IDs in a single JNIEnv context).

stanislav-tkach:
I need to think a little bit more about this, but right now I'm more inclined to the third approach (save class_ref + id pairs).

PR #276, opened 2020-10-24
Deprecate get byte array elements

### LINKED ISSUE #275: Deprecate get_byte_array_elements
Deprecate `get_byte_array_elements` and related methods in JNIEnv in favour of `get_auto_byte_array_elements`.

As have been agreed in #272, `Auto*Array`s are easier and safer to use. The only known missing piece of functionality is _conditionally_ discrarding changes in a copied array during release.

- Support discarding changes to _copied_ `Auto*Array`s instead of releasing with the mode passed to its constructor. For non-copied arrays such operation does not make sense.
- Deprecate `get_byte_array_elements` for removal.

maurolacy:
## Overview

Closes #275.

Will modify / extend unit tests soon.

**Update**: I think unit tests are fine as they are. The new `discard()` functionality is straightforward. And, we're not currently explicitly testing a call to `commit()` either (which is also straightforward).
In any case, If you want me to add explicit tests for these, I'll do it.
Also, not sure what to do with the tests of the deprecated methods... I guess we keep them, until removing these methods in a future release.

### Definition of Done

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Switch from `error-chain` to `thiserror`, making all errors `Send`. Also, support all JNI errors
   in the `jni_error_code_to_result` function and add more information to the `InvalidArgList`
   error. ([#242](https://github.com/jni-rs/jni-rs/pull/242))
+- `AutoByte/PrimitiveArray.commit()` now returns `Result`. (#275)
 
 ## [0.17.0] — 2020-06-30
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - `TryFrom<JValue>` for types inside JValue variants (#264).
 - Implemented Copy for JNIEnv (#255).
 - `repr(transparent)` attribute to JavaVM struct (#259)
+- `discard()` method to `AutoByteArray` and `AutoPrimitiveArray`. (#275)
+- deprecation notice for get/release/commit_byte/primitive_array_{elements|critical}. (#275)
 
 ### Changed
 - Switch from `error-chain` to `thiserror`, making all errors `Send`. Also, support all JNI errors
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
+- `discard()` method to `AutoByteArray` and `AutoPrimitiveArray`. (#275)
+- deprecation notice for get/release/commit_byte/primitive_array_{elements|critical}. (#275)
+
+### Changed
+- `AutoByte/PrimitiveArray.commit()` now returns `Result`. (#275)
+
 ## [0.18.0] — 2020-09-23
 
 ### Added
@@ -24,14 +31,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - `TryFrom<JValue>` for types inside JValue variants (#264).
 - Implemented Copy for JNIEnv (#255).
 - `repr(transparent)` attribute to JavaVM struct (#259)
-- `discard()` method to `AutoByteArray` and `AutoPrimitiveArray`. (#275)
-- deprecation notice for get/release/commit_byte/primitive_array_{elements|critical}. (#275)
 
 ### Changed
 - Switch from `error-chain` to `thiserror`, making all errors `Send`. Also, support all JNI errors
   in the `jni_error_code_to_result` function and add more information to the `InvalidArgList`
   error. ([#242](https://github.com/jni-rs/jni-rs/pull/242))
-- `AutoByte/PrimitiveArray.commit()` now returns `Result`. (#275)
 
 ## [0.17.0] — 2020-06-30
 
--- a/src/wrapper/objects/auto_byte_array.rs
+++ b/src/wrapper/objects/auto_byte_array.rs
@@ -58,7 +58,7 @@ impl<'a, 'b> AutoByteArray<'a, 'b> {
         self.ptr.as_ptr()
     }
 
-    /// Commits the result of the array, if it is a copy
+    /// Commits the changes to the array, if it is a copy
     pub fn commit(&mut self) {
         let res = self
             .env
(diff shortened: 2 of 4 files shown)


src/wrapper/objects/auto_byte_array.rs
```diff
@@ -58,17 +58,45 @@ impl<'a, 'b> AutoByteArray<'a, 'b> {
         self.ptr.as_ptr()
     }
 
-    /// Commits the result of the array, if it is a copy
+    fn commit_byte_array_elements(&mut self) -> Result<()> {
```
dmitry-timofeev, line 61: Nit: I'd place these utility methods after their callers, so that it is easier to see how they are used.

src/wrapper/objects/auto_byte_array.rs
```diff
@@ -58,17 +58,45 @@ impl<'a, 'b> AutoByteArray<'a, 'b> {
...
+
+    /// Commits the changes to the array, if it is a copy
     pub fn commit(&mut self) {
-        let res = self
-            .env
-            .commit_byte_array_elements(*self.obj, unsafe { self.ptr.as_mut() });
+        let res = self.commit_byte_array_elements();
         match res {
             Ok(()) => {}
             Err(e) => debug!("error committing byte array: {:#?}", e),
```
dmitry-timofeev, line 88: error, so that users are more likely to see and investigate it?

src/wrapper/objects/auto_byte_array.rs
```diff
@@ -58,17 +58,45 @@ impl<'a, 'b> AutoByteArray<'a, 'b> {
...
         }
     }
 
+    /// Don't commit the changes to the array on release (if it is a copy).
+    /// This has no effect if the array is not a copy.
```
dmitry-timofeev, line 93: I wonder if it shall be a precondition that a user shall check? Whether an array is a copy is out of a user's control. Therefore, they can only rely on `discard` if they ensure that an array is a copy.

What would be the best approach here? A requirement that a user checks if an array is a copy, and an error here? Or panic? Or keep as is?

src/wrapper/objects/auto_byte_array.rs
```diff
@@ -58,17 +58,45 @@ impl<'a, 'b> AutoByteArray<'a, 'b> {
...
+            self.env.get_native_interface(),
+            ReleaseByteArrayElements,
+            *self.obj,
+            self.ptr.as_mut(),
+            sys::JNI_COMMIT
+        );
+        Ok(())
+    }
+
+    fn release_byte_array_elements(&mut self) -> Result<()> {
```
dmitry-timofeev, line 72: Here and in the critical flavour: would it make sense to add an utility method 
```
fn release_byte_array_elements(&mut self, i32)
```

and then reusing it in both of these methods:
```
commit: release_byte_array_elements(sys::JNI_COMMIT)

release: release_byte_array_elements(mode as i32)
```

dmitry-timofeev (commented):
Great job, thanks for working on it!

src/wrapper/objects/auto_byte_array.rs
maurolacy: Sure. I wanted to keep the API as similar as possible, but I think this makes sense.

src/wrapper/objects/auto_byte_array.rs
maurolacy: I think it's the user's responsibility; user should have checked the array is a copy or not long before calling this method, and there's a public method for doing that.

The most appropriate thing would be a warning, but I don't want to potentially pollute the output with it. Besides, the documentation is clear; I would leave it as it is.

src/wrapper/objects/auto_byte_array.rs
maurolacy: This could be an option; particularly now that these methods are private. The problem is with the naming, that comes from the JNI overloading of release and commit... I would leave it as it is, just for clarity.

maurolacy: On second though, I think you're right. The new version is more compact, and the naming just follows the original JNI impl.

dmitry-timofeev: Not sure I get the naming issue: if a new method is named `release_array_elements` and accepts a release mode, it will be the same as the JNI method, but with some of its arguments already applied (= coming from the struct fields). We can then inline the other two private methods (I see commit is already inlined), or rename them to "release"/"commit".

maurolacy: Yep. See the latest version.

CHANGELOG.md
```diff
@@ -31,6 +31,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Switch from `error-chain` to `thiserror`, making all errors `Send`. Also, support all JNI errors
   in the `jni_error_code_to_result` function and add more information to the `InvalidArgList`
   error. ([#242](https://github.com/jni-rs/jni-rs/pull/242))
+- `AutoByte/PrimitiveArray.commit()` now returns `Result`. (#275)
```
dmitry-timofeev, line 34: I've released 0.18, could you please move the entries for this PR to "unreleased" section?

src/wrapper/objects/auto_primitive_array.rs
```diff
@@ -46,7 +46,8 @@ impl<'a, 'b> AutoPrimitiveArray<'a, 'b> {
         self.ptr.as_ptr()
     }
 
-    fn commit_primitive_array_critical(&mut self) -> Result<()> {
+    /// Commits the changes to the array, if it is a copy
+    pub fn commit(&mut self) -> Result<()> {
         jni_void_call!(
```
dmitry-timofeev, line 51: Wouldn't similar changes to deduplicate the methods make sense here?

dmitry-timofeev (approved):
LGTM, thanks!

Regarding tests — I'd consider adding tests for commit, but it can be done separately from this PR.

CHANGELOG.md
maurolacy: Sure. I'll change the version in the #deprecation notices.

src/wrapper/objects/auto_primitive_array.rs
maurolacy: IMO, no. Exacly because of the naming conventions. We shouldn't carry or reflect a mistake or poor design decision in the underlying API to our public API. This is opinionated, I know; and that's my opinion. :-)

maurolacy: Ah, you're thinking about doing the same as before but for primitive array... Sure! good catch, I thought I already did that.

dmitry-timofeev (approved):

dmitry-timofeev:
@stanislav-tkach merged this PR to prevent any conflicts with #272 , please review when you have time.

PR #280, opened 2020-10-30
Allow `deprecated` warning in CI build

s1ck:
Manual methods for handling byte arrays are being deprecated in https://github.com/jni-rs/jni-rs/pull/276 but are used in tests, which is considered a warning by clippy and leads to a failing CI build.

### DIFF
--- a/.travis/run_travis_job.sh
+++ b/.travis/run_travis_job.sh
@@ -34,7 +34,7 @@ echo 'Performing checks over the rust code'
 cargo fmt --all -- --check
 
 # Run clippy static analysis.
-cargo clippy --all --tests --all-features -- -D warnings
+cargo clippy --all --tests --all-features -- -D warnings -A deprecated
 
 # Run tests with default features (stable-only)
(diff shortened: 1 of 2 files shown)


maurolacy (commented):
LGTM. Deprecation warnings can be re-enabled after removing deprecated methods.

maurolacy:
I think doing it all together is a good idea. You can count on me to do it.

maurolacy:
See #281.

.travis/run_travis_job.sh
```diff
@@ -34,7 +34,7 @@ echo 'Performing checks over the rust code'
 cargo fmt --all -- --check
 
 # Run clippy static analysis.
-cargo clippy --all --tests --all-features -- -D warnings
+cargo clippy --all --tests --all-features -- -D warnings -A deprecated
```
stanislav-tkach, line 37: Can this be allowed at the test/function level? I think we don't want to allow app deprecated methods.

stanislav-tkach:
I suppose this can be closed because #282 is merged.

PR #284, opened 2020-11-03
Fix: ReleaseMode struct / methods (#283)

### LINKED ISSUE #283: Get array elements error
Get array elements test are [failing](https://github.com/jni-rs/jni-rs/pull/282/checks?check_run_id=1343885451):

The error is:
```
ReleasePrimitiveArrayCritical: release array failed bounds check, incorrect pointer returned ? array: 0x00007f4de007bc40 carray: 0x00007f4dd8001260
GuardedMemory(0x00007f4de445c420) base_addr=0x00007f4dd8001240 tag=0x00007f4dd8001720 user_size=24 user_data=0x00007f4dd8001260
  Header guard @0x00007f4dd8001240 is BROKEN
  Trailer guard @0x00007f4dd8001278 is OK
  User data appears to have been freed
FATAL ERROR in native method: ReleasePrimitiveArrayCritical: failed bounds check
```

maurolacy:
## Overview

Fixes #283. Sorry for the confusion regarding release modes. I think it's OK now; please review.

### Definition of Done

### DIFF
--- a/tests/jni_api.rs
+++ b/tests/jni_api.rs
@@ -15,6 +15,7 @@ use jni::{
 
 mod util;
 use jni::objects::ReleaseMode;
+use jni::objects::ReleaseMode::CopyBackNoFree;
 use util::{attach_current_thread, unwrap};
 
 static ARRAYLIST_CLASS: &str = "java/util/ArrayList";
@@ -356,9 +357,6 @@ pub fn get_byte_array_elements() {
             *ptr.offset(1) += 1;
             *ptr.offset(2) += 1;
         }
-
-        // Commit would be necessary here, if there were no closure
-        //auto_ptr.commit().unwrap();
     }
 
     // Confirm modification of original Java array
@@ -413,9 +411,6 @@ pub fn get_long_array_elements() {
             *ptr.offset(1) += 1;
             *ptr.offset(2) += 1;
         }
-
-        // Commit would be necessary here, if there were no closure
-        //auto_ptr.commit().unwrap();
     }
 
     // Confirm modification of original Java array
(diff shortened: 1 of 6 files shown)


tests/jni_api.rs
```diff
@@ -508,13 +503,19 @@ pub fn get_auto_long_array_elements_commit() {
...
+    // Call drop explicitly
+    std::mem::drop(auto_ptr);
 
     // Confirm modification of original Java array
     env.get_long_array_region(java_array, 0, &mut res).unwrap();
     assert_eq!(res[0], 2);
     assert_eq!(res[1], 3);
     assert_eq!(res[2], 4);
+
+    // Native buffer automatically released here anyway
```
dmitry-timofeev, line 518: I might be missing something, but how exactly is it freed? It doesn't seem to get freed on AutoArrays drop, because it is simply stored in a NonNull. If I understand correctly, at this point it is "owned" by the client code.

maurolacy: I understand the same. It will be freed when the `AutoArray` object goes out of scope. Not because of the call to `drop()`, but just because of automatic memory management.

maurolacy: OK, I've added an extra check to the original `commit` test, and it shows that the original pointer contents change after `drop`, and are therefore no longer valid. 

So, my take here is that what you say is correct: `drop` is releasing the native data associated to the pointer. Not because of the call to Release, but as part of normal (Rust's) memory / borrowing management.

I've also tried explicitly calling `drop` again, on `ptr` and `*ptr`, to see what happens, and couldn't observe any effect.

Finally, I've also tried with the original release mode (`CopyBack`) and couldn't detect any difference in behaviour. This makes me doubt the usefulness of the whole `JNI_COMMIT` thing, in Rust at least.

I'm now thinking that we can probably remove `JNI_COMMIT` enum from `ReleaseMode`. At least, until a clear use case (or mode of use) arises.

dmitry-timofeev: Interesting, does Rust insert `free` for random pointers inside NonNull 🤔? If it does, why don't we get double-free because of Release(CopyBack) and subsequent "rust magic"?

maurolacy: I don't know. It may be `NonNull`, but it also may be the implementation of `ReleaseArrayElements` we're hitting. I guess the only way to know what's happening for sure, would be to get a debugger, and follow the assembler.

maurolacy: Another interesting tidbit: I see now that my assertion here in the CI failed, whereas in my local machine, it passed (invalid values).

@dmitry-timofeev, if you guys give me some hints, I'm willing to debug / continue tracking this. Probably will require to compile libjni for debug / etc.

stanislav-tkach: Rust definitely shouldn't free anything inside `NonNull`: it doesn't even have a `Drop` implementation. It would be nice to figure out what issue we are hitting and why, but right now I have no idea, either.

maurolacy:
Closing this in favor of a follow-up on #283.

PR #286, opened 2020-12-09
Add size() helper for Auto*Array

### LINKED ISSUE #278: Consider adding AutoArray.size
As it already contains the Java array reference and JNIEnv, it could also support size, so that the users don't have to request it separately with `get_array_length`.

maurolacy:
## Overview

Closes #278. 

### Definition of Done

### DIFF
--- a/tests/jni_api.rs
+++ b/tests/jni_api.rs
@@ -433,6 +433,7 @@ pub fn get_long_array_elements() {
 }
 
 #[test]
+#[ignore]
 pub fn get_long_array_elements_commit() {
     let env = attach_current_thread();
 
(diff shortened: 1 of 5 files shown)


maurolacy:
Tests are failing because of the commit() bug. Maybe we should disable that test in the meantime.

tests/jni_api.rs
```diff
@@ -426,7 +432,7 @@ pub fn get_long_array_elements() {
     assert_eq!(res[2], 4);
 }
 
-#[test]
+//#[test]
```
stanislav-tkach, line 435: I suppose marking test as ignored (`#[ignore]`) would be better. Ideally an explaining comment should be added too.

stanislav-tkach (approved):

tests/jni_api.rs
maurolacy: Sure. I didn't knew about `#[ignore]`, thanks.

dmitry-timofeev (approved):
👍

PR #287, opened 2020-12-12
Get array elements type trait

maurolacy:
## Overview

Draws on top of #273, using a `TypeArray` trait to abstract the type-dependent parts of the impl that cannot be generalized using generic data types.
Then uses a `type_array!` macro to implement these traits for the required primitive types.

### Definition of Done


stanislav-tkach (approved):
👍

maurolacy:
Great. I'll check de documentation / comments once more, as soon as possible, and we can merge this!

maurolacy:
> I'll check de documentation / comments once more, as soon as possible, and we can merge this!

LGTM.

src/wrapper/jnienv.rs
```diff
@@ -1973,52 +1973,105 @@ impl<'a> JNIEnv<'a> {
...
-    /// AutoByteArray has a commit() method, to force a copy of the array if needed (and without
+    /// the corresponding Release*ArrayElements JNI method is called.
+    /// AutoArray has a commit() method, to force a copy of the array if needed (and without
     /// releasing it).
-    ///
-    /// If the given array is `null`, an `Error::NullPtr` is returned.
-    ///
-    /// See also [`get_primitive_array_critical`](struct.JNIEnv.html#method.get_primitive_array_critical)
-    pub fn get_byte_array_elements(
+    /// See also the convenience wrappers:
```
dmitry-timofeev, line 1986:
```suggestion

    /// Prefer to use the convenience wrappers:
```

src/wrapper/objects/auto_array.rs
```diff
@@ -0,0 +1,135 @@
...
+    env: &'b JNIEnv<'a>,
+}
+
+impl<'a, 'b, T: TypeArray> AutoArray<'a, 'b, T> {
+    /// Creates a new auto-release wrapper for a pointer-based generic array
+    ///
+    /// Once this wrapper goes out of scope, `Release<Type>ArrayElements` will be
+    /// called on the object. While wrapped, the object can be accessed via
+    /// the `From` impl.
+    pub fn new(env: &'b JNIEnv<'a>, obj: JObject<'a>, mode: ReleaseMode) -> Result<Self> {
```
dmitry-timofeev, line 74: Shan't it be crate-local (pub crate)?

src/wrapper/objects/auto_array.rs
```diff
@@ -0,0 +1,135 @@
...
+        })
+    }
+
+    /// Get a reference to the wrapped pointer
+    pub fn as_ptr(&self) -> *mut T {
+        self.ptr.as_ptr()
+    }
+
+    /// Commits the changes to the array, if it is a copy
+    pub fn commit(&mut self) -> Result<()> {
```
dmitry-timofeev, line 94: Does it have to be `&mut`? No modifications to the array/its copy are performed in a release.

src/wrapper/objects/auto_array.rs
```diff
@@ -0,0 +1,135 @@
...
+        self.env.get_array_length(*self.obj)
+    }
+}
+
+impl<'a, 'b, T: TypeArray> Drop for AutoArray<'a, 'b, T> {
+    fn drop(&mut self) {
+        let res = self.release_array_elements(self.mode as i32);
+        match res {
+            Ok(()) => {}
+            Err(e) => debug!("error releasing array: {:#?}", e),
```
dmitry-timofeev, line 126:
```suggestion
            Err(e) => error!("error releasing array: {:#?}", e),
```

tests/jni_api.rs
```diff
@@ -315,122 +315,140 @@ pub fn java_byte_array_from_slice() {
...
+    set_int_array_region
+);
+
+test_get_array_elements!(
+    get_int_array_elements,
+    jint,
+    new_int_array,
+    get_int_array_region,
+    set_int_array_region
+);
```
dmitry-timofeev, line 395: Are these two any different?

tests/jni_api.rs
```diff
@@ -315,122 +315,140 @@ pub fn java_byte_array_from_slice() {
...
+    set_byte_array_region
+);
+
+test_get_array_elements!(
+    get_boolean_array_elements,
+    jboolean,
+    new_boolean_array,
+    get_boolean_array_region,
+    set_boolean_array_region
+);
```
dmitry-timofeev, line 419: Are "1", "2", "3" values for boolean well defined? What does their increment mean?

tests/jni_api.rs
```diff
@@ -315,122 +315,140 @@ pub fn java_byte_array_from_slice() {
...
+                // Commit would be necessary here, if there were no closure
+                //auto_ptr.commit().unwrap();
+            }
+
+            // Confirm modification of original Java array
+            let mut res: [$jni_type; 3] = [0 as $jni_type; 3];
+            env.$get_array(java_array, 0, &mut res).unwrap();
+            assert_eq!(res[0] as i32, 2);
+            assert_eq!(res[1] as i32, 3);
+            assert_eq!(res[2] as i32, 4);
```
dmitry-timofeev, line 376: Isn't that problematic for floats?

src/wrapper/objects/auto_array.rs
maurolacy: These are left-overs from the original version. Good catch.

tests/jni_api.rs
maurolacy: No, not really. It's there just to show that you can call the generic / parametric method if you want.

I can remove it.

tests/jni_api.rs
maurolacy: Good point. I haven't tought about it. I'll separate that test case for clarity.

tests/jni_api.rs
maurolacy: That's there because float equalities are unreliable, and require a precision parameter. It's that, or introducing a precision parameter for those equalities.

This is just a way to "overload" the test case to work for floats. I can also separate float tests cases for clarity and  "correctness", but I don't think it's really necessary. As you wish.

tests/jni_api.rs
maurolacy: Come to think of it, in Java boolean is well defined, as you have true and false. The same in Rust.

But in JNI land, these "booleans" are just `u8`s under the hood. So, these "1", "2", "3" values are all a redudant way to say "true".

I would leave it as it is, not to break the generality of the test macro. I can also modify the test macro to start with a zero and do one increment to 1, so that it most clearly mirrors the boolean types.

What do you think?

tests/jni_api.rs
dmitry-timofeev: I see, I didn't notice. Let's add a comment? 
// Test a generic get_array_elements 
// Test type-specific array accessors

tests/jni_api.rs
dmitry-timofeev: I'd use a separate test for boolean, because the spec says the values are either 0 or 1: 
https://docs.oracle.com/en/java/javase/11/docs/specs/jni/types.html#primitive-types

Writing anything else is probably UB: JVM may keep the provided values; or normalize them; or fail. If the test passes, then it probably keeps them as is.

Generally, writing anything other than 0 or 1 may be problematic if some other JNI code reading these values compares them to JNI_FALSE/JNI_TRUE.

tests/jni_api.rs
dmitry-timofeev: OK, let's keep it as is, if these casts and comparisons work.

tests/jni_api.rs
maurolacy: Take a look at the latest version, and if you don't like it, I'll just write a separate test.

dmitry-timofeev (approved):
Let's probably keep it like that, thank you for updating!

PR #288, opened 2020-12-13
Add (minimal) test of primitive array critical commit()

maurolacy:
## Overview

Adding a (minimal) test to confirm that `GetPrimitiveArrayCritical` also errors with a double free when invoked twice (first time with `JNI_COMMIT`) and `-Xcheck:jni` is enabled.

Related to #283.

Looking at the output, both implementations (`ArrayElements` and `ArrayCritical`) seem to be based on the same code. At least when `-Xcheck:jni` is enabled.
The only difference in the error output (besides the pointer addresses) is that the `Critical` version adds a warning:
" Warning: Calling other JNI functions in the scope of Get/ReleasePrimitiveArrayCritical or Get/ReleaseStringCritical"
Interestingly, this warning is emitted even when no other JNI functions seem to be called (the test is minimal to confirm that).

Perhaps we can merge this to master, after disabling the test with `#ignore`, like in the other `commit()` test.

Finally, this bug raises a couple of questions / issues:
- Maybe both settings (`-Xcheck:jni` enabled / present, and `-Xcheck:jni` not present) should be tested.
- What about other JDKs / non HotSpot VMs ? Would it make sense to test against Graal, OpenJ9, etc. for compatibility?

### Definition of Done

### DIFF
--- a/src/wrapper/jnienv.rs
+++ b/src/wrapper/jnienv.rs
@@ -2050,7 +2050,7 @@ impl<'a> JNIEnv<'a> {
     ) -> Result<AutoPrimitiveArray> {
         non_null!(array, "get_primitive_array_critical array argument");
         let mut is_copy: jboolean = 0xff;
-        let ptr = jni_non_void_call!(
+        let ptr = jni_unchecked!(
             self.internal,
             GetPrimitiveArrayCritical,
             array,
--- a/src/wrapper/jnienv.rs
+++ b/src/wrapper/jnienv.rs
@@ -1994,7 +1994,7 @@ impl<'a> JNIEnv<'a> {
     ) -> Result<AutoByteArray> {
         non_null!(array, "get_byte_array_elements array argument");
         let mut is_copy: jboolean = 0xff;
-        let ptr = jni_non_void_call!(self.internal, GetByteArrayElements, array, &mut is_copy);
+        let ptr = jni_unchecked!(self.internal, GetByteArrayElements, array, &mut is_copy);
         AutoByteArray::new(self, array.into(), ptr, mode, is_copy == sys::JNI_TRUE)
     }
 
@@ -2017,7 +2017,7 @@ impl<'a> JNIEnv<'a> {
     ) -> Result<AutoLongArray> {
         non_null!(array, "get_long_array_elements array argument");
         let mut is_copy: jboolean = 0xff;
-        let ptr = jni_non_void_call!(self.internal, GetLongArrayElements, array, &mut is_copy);
+        let ptr = jni_unchecked!(self.internal, GetLongArrayElements, array, &mut is_copy);
         AutoLongArray::new(self, array.into(), ptr, mode, is_copy == sys::JNI_TRUE)
     }
 
--- a/src/wrapper/jnienv.rs
+++ b/src/wrapper/jnienv.rs
@@ -2056,7 +2056,6 @@ impl<'a> JNIEnv<'a> {
             array,
             &mut is_copy
         );
-        non_null!(ptr, "get_primitive_array_critical return value");
         AutoPrimitiveArray::new(self, array.into(), ptr, mode, is_copy == sys::JNI_TRUE)
     }
 }
--- a/tests/jni_api.rs
+++ b/tests/jni_api.rs
@@ -433,7 +433,7 @@ pub fn get_long_array_elements() {
 }
 
 #[test]
-#[ignore]
+#[ignore] // Disabled until issue #283 is resolved
 pub fn get_long_array_elements_commit() {
     let env = attach_current_thread();
 
@@ -535,6 +535,7 @@ pub fn get_primitive_array_critical() {
 }
 
 #[test]
+#[ignore] // Disabled until issue #283 is resolved
 pub fn get_primitive_array_critical_commit() {
     let env = attach_current_thread();
 
--- a/tests/jni_api.rs
+++ b/tests/jni_api.rs
@@ -535,6 +535,24 @@ pub fn get_primitive_array_critical() {
 }
 
 #[test]
+pub fn get_primitive_array_critical_commit() {
+    let env = attach_current_thread();
+
+    // Create original Java array
+    let java_array = env
+        .new_long_array(3)
+        .expect("JNIEnv#new_long_array must create a java array with given size");
+
+    // Get primitive array elements auto wrapper
+    let mut auto_ptr = env
+        .get_primitive_array_critical(java_array, ReleaseMode::CopyBack)
+        .unwrap();
+
+    // Call commit
+    auto_ptr.commit().unwrap();
+}
+
+#[test]
 pub fn get_object_class() {
     let env = attach_current_thread();
     let string = env.new_string("test").unwrap();


tests/jni_api.rs
```diff
@@ -534,6 +534,24 @@ pub fn get_primitive_array_critical() {
...
+        .new_long_array(3)
+        .expect("JNIEnv#new_long_array must create a java array with given size");
+
+    // Get primitive array elements auto wrapper
+    let mut auto_ptr = env
+        .get_primitive_array_critical(java_array, ReleaseMode::CopyBack)
+        .unwrap();
+
+    // Call commit
+    auto_ptr.commit().unwrap();
```
dmitry-timofeev, line 552: Shan't we have any assertions here (e.g., modify the first element, check that it's there)? However, the JNI spec does not allow calling other JNI methods whilst holding a critical array, so it might not be possible 🤔

dmitry-timofeev:
Thanks for putting this test together!

> " Warning: Calling other JNI functions in the scope of Get/ReleasePrimitiveArrayCritical or Get/ReleaseStringCritical"

Good catch! I see get_primitive_array_critical uses a macro that checks exceptions with is a JNI function: `jni_non_void_call`. If we use a macro that doesn't do that (`jni_unchecked`), but check the result of this function to be [non-null](https://docs.oracle.com/en/java/javase/11/docs/specs/jni/functions.html#getprimitivearraycritical-releaseprimitivearraycritical), the warning will be gone:
https://github.com/jni-rs/jni-rs/blob/1d51a37d5c1591e13fcb8612a412e3ac6297807b/src/wrapper/macros.rs#L11-L24

> Maybe both settings (-Xcheck:jni enabled / present, and -Xcheck:jni not present) should be tested.

I think there is no need to test without, as this option merely adds extra checks to the interface.

> What about other JDKs / non HotSpot VMs ? Would it make sense to test against Graal, OpenJ9, etc. for compatibility?

Possibly, though, if they implement the spec correctly (and test their implementation), there must be no difference. Graal for Java, IIRC, uses Hotspot with a different C2 compiler.

tests/jni_api.rs
maurolacy: Exactly. That's why I wanted to do a minimal test. Just to confirm that the bug is there. And, that it's not related to calling any other JNI methods (at least, explicitly).

maurolacy:
> Thanks for putting this test together!
> 
> > " Warning: Calling other JNI functions in the scope of Get/ReleasePrimitiveArrayCritical or Get/ReleaseStringCritical"
> 
> Good catch! I see get_primitive_array_critical uses a macro that checks exceptions with is a JNI function: `jni_non_void_call`. If we use a macro that doesn't do that (`jni_unchecked`), but check the result of this function to be [non-null](https://docs.oracle.com/en/java/javase/11/docs/specs/jni/functions.html#getprimitivearraycritical-releaseprimitivearraycritical), the warning will be gone:
> 
> https://github.com/jni-rs/jni-rs/blob/1d51a37d5c1591e13fcb8612a412e3ac6297807b/src/wrapper/macros.rs#L11-L24

Good to know. Will modify the test to use just that.

> 
> > Maybe both settings (-Xcheck:jni enabled / present, and -Xcheck:jni not present) should be tested.
> 
> I think there is no need to test without, as this option merely adds extra checks to the interface.

True. But in these particular cases (Release<Type>ArrayElements / ReleasePrimitive) by example, the implementation (and behaviour) is clearly different. With the flag enabled the buffer is a copy, and without, it's not. That's in fact why we're hittting the bug with the flag enabled. At least, according to the OpenJDK guys.

I agree that this may be an exceptional / rare case, though.

maurolacy:
> Good to know. Will modify the test to use just that.

In fact, instead of modifying the test, I've modified the `get_primitive_array_critical()` impl.

I think it makes sense to use `jni_unchecked!`, as we're already checking for non null just after calling the method. And, perhaps more importantly, we don't want to call other JNI functions, which may cause unexpected behaviour, and triggers a wwarning. Third, it's also more efficient to avoid checking for non null using JNI.

maurolacy:
So, I would say, let's disable this currently broken test with `#ignore`, and let's merge for documentation, correctness, and probably also performance.

maurolacy:
> Lastly, checking for non null / exceptions using JNI would be relatively costly.

In fact I've done some very specific flamegraphs of the "array critical" calls under different JVMs, and for HotSpot VMs in particular, a good portion of the time was spent in the "checking exceptions" part.

dmitry-timofeev (approved):
👍

src/wrapper/jnienv.rs
```diff
@@ -1994,7 +1994,7 @@ impl<'a> JNIEnv<'a> {
     ) -> Result<AutoByteArray> {
         non_null!(array, "get_byte_array_elements array argument");
         let mut is_copy: jboolean = 0xff;
-        let ptr = jni_non_void_call!(self.internal, GetByteArrayElements, array, &mut is_copy);
+        let ptr = jni_unchecked!(self.internal, GetByteArrayElements, array, &mut is_copy);
```
dmitry-timofeev, line 1997: If we replace jni_non_void_call that checks exceptions with jni_unchecked, we have to check the result to be non-null:

> RETURNS:
> Returns a pointer to the array elements, or NULL if the operation fails.
> THROWS:
> OutOfMemoryError: if the system runs out of memory.

dmitry-timofeev (changes requested):

dmitry-timofeev:
> In fact, I've done some specific, precise flamegraphs of the "array critical" calls under different JVMs in the past, and for OpenJDK 8 in particular, a good portion of the time was spent in the "checking exceptions" part.

@skletsun did a review of macro usages some time ago and fixed the usages where it was unnecessary, but some could remain. Many JNI methods do have to check for exceptions, according to the spec :)

src/wrapper/jnienv.rs
maurolacy: We're checking, because we're using `NonNull` wrapper inside `AutoArray`, so this is fine, and it will be more efficient.

maurolacy:
> > In fact, I've done some specific, precise flamegraphs of the "array critical" calls under different JVMs in the past, and for OpenJDK 8 in particular, a good portion of the time was spent in the "checking exceptions" part.
> 
> @skletsun did a review of macro usages some time ago and fixed the usages where it was unnecessary, but some could remain. Many JNI methods do have to check for exceptions, according to the spec :)

Sure. Here I was speaking only of these methods in particular.

src/wrapper/jnienv.rs
maurolacy: Or do you prefer to keep `GetXArrayElements` with `jni_non_void_call!`?

I've seen that you made a request for changes, but I'm not sure which ones.

stanislav-tkach (approved):

src/wrapper/jnienv.rs
dmitry-timofeev: Sorry, `jni_unchecked` is fine if the result is checked, I missed that code. Thanks for clarifications!

dmitry-timofeev (approved):

PR #304, opened 2021-02-24
Add support for weak references

### LINKED ISSUE #167: Support for weak globals refs

### LINKED ISSUE #301: `JNIEnv::new_local_ref` has seemingly-unnecessary lifetime restriction
The JNI `NewLocalRef` function can be used to create a local reference from a global reference. [The spec specifically allows it](https://docs.oracle.com/en/java/javase/11/docs/specs/jni/functions.html#newlocalref), and I'm guessing that's its main purpose. However, the wrapper method `JNIEnv::new_local_ref` currently can't be used that way because it requires the provided `JObject` to have the same lifetime as the `JNIEnv` itself (which won't be the case for a `JObject` belonging to a `GlobalRef`). This restriction does not seem necessary and should probably be removed (that is, the type of `obj` should be changed from `JObject<'a>` to `JObject<'_>`).

Also, `JNIEnv::new_local_ref` has a type parameter `T` that doesn't seem to be used at all.

argv-minus-one:
## Overview

This PR adds a `WeakRef` type and `JNIEnv::new_weak_ref` method, allowing the user to create weak references to Java objects. Fixes #167.

This is an alternative to #266. I wrote this before checking if someone had already implemented weak references for the `jni` crate, but I decided to submit a PR anyway because this is different from #266.

### Differences From #266

* This PR [doesn't introduce][no JWeak comment] a `JWeak` type. It only introduces `WeakRef`.
* `WeakRef` can be upgraded to either a local or global strong reference.
* Upgrading returns `Result<Option<JObject>>` or `Result<Option<GlobalRef>>`, where the `Option` is `None` if the object was garbage collected. Returning `Option` [is debated][Option comment] in #266, but I think it's the right approach because:
	1. It's immediately clear what happens if the reference has been garbage collected.
	2. `std::{rc,sync}::Weak::upgrade` also returns `Option`.
* `JNIEnv::new_weak_ref` returns `Result<Option<WeakRef>>`, where the `Option` is `None` if the original reference is null. That way, there's no [ambiguity between a weak reference that was originally null and one that's been garbage collected][null weak ref comment]. The JVM won't actually create a null weak reference anyway ([`NewWeakGlobalRef`] returns a null pointer in this case), and this code follows that behavior.
* A test verifies that weak references can be used across different threads.
* A test verifies that an object can be garbage-collected if there is only a weak reference to it.

### Note: Also Fixes #301

Note that this PR contains a fix for #301 (separately, as the first commit). That's needed because `WeakRef::upgrade_local` calls `JNIEnv::new_local_ref` to create the new local reference, which isn't otherwise possible due to #301.

This is technically a **breaking change**, although `JNIEnv::new_local_ref` doesn't seem very useful in its current form, so there may not be any existing uses of that method for this change to break.

[`NewWeakGlobalRef`]: https://docs.oracle.com/en/java/javase/11/docs/specs/jni/functions.html#newweakglobalref
[no JWeak comment]: https://github.com/jni-rs/jni-rs/pull/266#discussion_r504867022
[null weak ref comment]: https://github.com/jni-rs/jni-rs/pull/266#discussion_r504878736
[Option comment]: https://github.com/jni-rs/jni-rs/pull/266#discussion_r502779236

### Definition of Done

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - `JavaStr::into_raw()` which drops the `JavaStr` and releases ownership of the raw string pointer ([#374](https://github.com/jni-rs/jni-rs/pull/374))
 - `JavaStr::from_raw()` which takes ownership of a raw string pointer to create a `JavaStr` ([#374](https://github.com/jni-rs/jni-rs/pull/374))
 - `JNIEnv::get_string_unchecked` is a cheaper, `unsafe` alternative to `get_string` that doesn't check the given object is a `java.lang.String` instance. ([#328](https://github.com/jni-rs/jni-rs/issues/328))
+- `WeakRef` and `JNIEnv#new_weak_ref`. ([#304](https://github.com/jni-rs/jni-rs/pull/304))
 
 ### Changed
 - `JNIEnv::get_string` checks that the given object is a `java.lang.String` instance to avoid undefined behaviour from the JNI implementation potentially aborting the program. ([#328](https://github.com/jni-rs/jni-rs/issues/328))
--- a/src/wrapper/jnienv.rs
+++ b/src/wrapper/jnienv.rs
@@ -16,7 +16,7 @@ use crate::{
     objects::{
         AutoArray, AutoLocal, AutoPrimitiveArray, GlobalRef, JByteBuffer, JClass, JFieldID, JList,
         JMap, JMethodID, JObject, JStaticFieldID, JStaticMethodID, JString, JThrowable, JValue,
-        ReleaseMode, TypeArray,
+        ReleaseMode, TypeArray, WeakRef,
     },
     signature::{JavaType, Primitive, TypeSignature},
     strings::{JNIString, JavaStr},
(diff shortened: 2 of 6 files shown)


argv-minus-one:
I've rebased this against current master and added some more methods to `WeakRef`.

rib:
Thanks for the rebase - I've added this to the 0.21 milestone and will try and find some time to review this soon

rib (approved):
These changes all look good to me.

Sorry I didn't included this in the last 0.20 release (though it's also good to have the additional workaround for badly behaved jvms here)

I had assumed (since there was also another separate attempt at enabling weak references) that this might end up being contentious or awkward to review but that wasn't the case at all.

PR #318, opened 2021-05-18
Don't tie the lifetime of `AutoArray` to a particular `JNIEnv` ref

### LINKED ISSUE #302: get_array_elements and friends unnecessarily restrict lifetimes
Now that `JNIEnv<'a>` is `Copy`, it's clear that the lifetime of the current context is the lifetime parameter `'a`, not the lifetime of any individual (Rust) reference to it. However, `get_array_elements` et al have an inferred result type that matches the lifetime of `&self`, not `'a`:

```rust
pub fn get_array_elements<T: TypeArray>(
    &self,
    array: jarray,
    mode: ReleaseMode
) -> Result<AutoArray<'_, '_, T>>
```

This could be fixed to include `'a` in the result type, but now that `JNIEnv<'a>` is `Copy`, there's no reason for AutoArray (and AutoPrimitiveArray, and JList/JMap/JavaStr) to store a reference to one at all.

```rust
pub struct AutoArray<'a, T: TypeArray> {
    obj: JObject<'a>,
    ptr: NonNull<T>,
    mode: ReleaseMode,
    is_copy: bool,
    env: JNIEnv<'a>,
}

pub fn get_array_elements<T: TypeArray>(
    &self,
    array: jarray,
    mode: ReleaseMode
) -> Result<AutoArray<'a, T>>
```

This would allow the AutoArray to outlast the current JNIEnv value, which makes sense when passing them around by Copy instead of by reference.

jrose-signal:
## Overview

Now that JNIEnv is publicly Copy, it's clear that the important lifetime is its parameter, not the lifetime of any particular reference. That means that AutoArray can live as long as the actual JNIEnv.

Fixes #302.

### Definition of Done

### DIFF
--- a/src/wrapper/jnienv.rs
+++ b/src/wrapper/jnienv.rs
@@ -2042,7 +2042,7 @@ impl<'a> JNIEnv<'a> {
         &self,
         array: jarray,
         mode: ReleaseMode,
-    ) -> Result<AutoArray<T>> {
+    ) -> Result<AutoArray<'a, T>> {
         non_null!(array, "get_array_elements array argument");
         AutoArray::new(self, array.into(), mode)
     }
@@ -2052,7 +2052,7 @@ impl<'a> JNIEnv<'a> {
         &self,
         array: jintArray,
         mode: ReleaseMode,
-    ) -> Result<AutoArray<jint>> {
+    ) -> Result<AutoArray<'a, jint>> {
         self.get_array_elements(array, mode)
     }
 
@@ -2061,7 +2061,7 @@ impl<'a> JNIEnv<'a> {
         &self,
         array: jlongArray,
         mode: ReleaseMode,
-    ) -> Result<AutoArray<jlong>> {
+    ) -> Result<AutoArray<'a, jlong>> {
         self.get_array_elements(array, mode)
     }
 
@@ -2070,7 +2070,7 @@ impl<'a> JNIEnv<'a> {
         &self,
         array: jbyteArray,
         mode: ReleaseMode,
-    ) -> Result<AutoArray<jbyte>> {
+    ) -> Result<AutoArray<'a, jbyte>> {
         self.get_array_elements(array, mode)
     }
 
@@ -2079,7 +2079,7 @@ impl<'a> JNIEnv<'a> {
         &self,
         array: jbooleanArray,
         mode: ReleaseMode,
-    ) -> Result<AutoArray<jboolean>> {
+    ) -> Result<AutoArray<'a, jboolean>> {
         self.get_array_elements(array, mode)
     }
 
@@ -2088,7 +2088,7 @@ impl<'a> JNIEnv<'a> {
         &self,
         array: jcharArray,
         mode: ReleaseMode,
-    ) -> Result<AutoArray<jchar>> {
+    ) -> Result<AutoArray<'a, jchar>> {
         self.get_array_elements(array, mode)
     }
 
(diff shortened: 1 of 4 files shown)


src/wrapper/jnienv.rs
```diff
@@ -1996,83 +1996,83 @@ impl<'a> JNIEnv<'a> {
     /// [`get_double_array_elements`](struct.JNIEnv.html#method.get_double_array_elements)
     /// And the associated [`AutoArray`](struct.objects.AutoArray) struct.
     pub fn get_array_elements<T: TypeArray>(
-        &self,
+        self,
```
maurolacy, line 1999: I would leave all these as they are. Even though the `JNIEnv` wrapper is minimal, using references is better.

And, handle all the changes just inside `AutoArray`.

src/wrapper/objects/auto_array.rs
```diff
@@ -64,16 +64,16 @@ type_array!(jdouble, GetDoubleArrayElements, ReleaseDoubleArrayElements);
...
 /// AutoArray provides automatic array release through a call to appropriate
 /// Release<Type>ArrayElements when it goes out of scope.
-pub struct AutoArray<'a: 'b, 'b, T: TypeArray> {
+pub struct AutoArray<'a, T: TypeArray> {
     obj: JObject<'a>,
     ptr: NonNull<T>,
     mode: ReleaseMode,
     is_copy: bool,
-    env: &'b JNIEnv<'a>,
+    env: JNIEnv<'a>,
```
maurolacy, line 72: _This_ is the meaningful change. Instead of storing a reference to `JNIEnv`, which comes with its own lifetime, we just store a copy of it. That allows us to get rid of `'b` entirely.

src/wrapper/objects/auto_array.rs
```diff
@@ -64,16 +64,16 @@ type_array!(jdouble, GetDoubleArrayElements, ReleaseDoubleArrayElements);
...
 }
 
-impl<'a, 'b, T: TypeArray> AutoArray<'a, 'b, T> {
-    pub(crate) fn new(env: &'b JNIEnv<'a>, obj: JObject<'a>, mode: ReleaseMode) -> Result<Self> {
+impl<'a, T: TypeArray> AutoArray<'a, T> {
+    pub(crate) fn new(env: JNIEnv<'a>, obj: JObject<'a>, mode: ReleaseMode) -> Result<Self> {
```
maurolacy, line 76: We can just use a "plain" (i.e. without lifetime marker) reference here.
```suggestion
    pub(crate) fn new(env: &JNIEnv<'a>, obj: JObject<'a>, mode: ReleaseMode) -> Result<Self> {
```

And de-reference it below, on line 86:
`            env: *env,`

That way, **all the rest can stay the same** (except for removing references to `'b'` (and `'_` when required).

maurolacy (changes requested):
Looks good to me.

Still, I'm requesting changes because I think it that the same can be done with less external changes.

src/wrapper/jnienv.rs
jrose-signal: I had made this change because passing by reference means extra pointer chasing, which is a tiny bit slower everywhere. But it *is* everywhere, since all the other APIs still work by reference, so I guess it makes sense to change that all in one go (and for now to leave it as is).

maurolacy: Yes. In any case, let's create an issue to change all those references at once.

Measuring / benchmarking wouldn't be bad either, to confirm which option is faster

jrose-signal:
Updated to not change public signatures, and also to add a test.

stanislav-tkach (approved):

Issue #303, opened 2021-02-23
Consider marking `JNIEnv::new_direct_byte_buffer` `unsafe`


kjvalencik:
If I understand correctly, `NewDirectByteBuffer` does not copy the buffer and instead maintains a pointer to data owned by Rust.

Since the lifetime of the argument cannot be enforced (there's no way to ensure the `ByteBuffer` doesn't outlive the Rust data, it may be worth marking this method as `unsafe`.

Currently, UB can be invoked purely through safe Rust.

```rust
use jni::*;
use jni::objects::*;

struct MyData {
    data: Vec<u8>,
}

impl MyData {
    fn new(s: &str) -> Self {
        Self {
            data: s.as_bytes().to_vec(),
        }
    }
}

impl AsMut<[u8]> for MyData {
    fn as_mut(&mut self) -> &mut [u8] {
        self.data.as_mut()
    }
}

impl Drop for MyData {
    fn drop(&mut self) {
        println!("Dropping data!");
    }
}

fn print_byte_buffer(env: &JNIEnv, buffer: &JByteBuffer) {
    let bytes = env.get_direct_buffer_address(buffer.clone()).unwrap();
    let s = String::from_utf8_lossy(bytes);

    println!("{}", s);
}

fn main() {
    // Build the VM properties
    let jvm_args = InitArgsBuilder::new()
        .build()
        .unwrap();

    let jvm = JavaVM::new(jvm_args).unwrap();
    let env = jvm.attach_current_thread().unwrap();
    let mut data = MyData::new("Hello, World!");
    let buffer = env.new_direct_byte_buffer(data.as_mut()).unwrap();

    print_byte_buffer(&env, &buffer);

    std::mem::drop(data);

    let x = MyData::new("Replace");
    // This is UB!!
    print_byte_buffer(&env, &buffer);

    let _ = x;
}
```

```
Hello, World!
Dropping data!
ReplaceWorld!
Dropping data!
```

Similarly, it may be valuable for `get_direct_buffer_address` to be marked `unsafe` since there's no guarantees that the pointer will be valid.

stanislav-tkach:
Thanks you for the report! I think that you are right and these functions should be marked as unsafe.

maurolacy:
Thanks. Can you submit a PR with the changes?

PR #320, opened 2021-06-02
safety: Mark `new_direct_byte_buffer` as `unsafe`

kjvalencik:
## Overview

**:warning: Breaking Change! :warning:**

Currently, undefined behavior may be invoked with `JNIEnv::new_direct_byte_buffer` because it does not verify the lifetime of the received byte slice. Unfortunately, since the JVM may maintain a reference to the `ByteBuffer` for an unknown amount of time, it's not possible to statically enforce this.

Since we cannot lean on the borrow checker, this method is marked _unsafe_ which is a breaking change.

An argument can be made that `JNIEnv:: get_direct_buffer_address` should _also_ be marked unsafe. However, this method is safe assuming the invariants of `NewDirectByteBuffer` are held. It's typical practice in Rust to mark these methods safe.

Resolves https://github.com/jni-rs/jni-rs/issues/303

### Definition of Done

### DIFF
--- a/src/wrapper/jnienv.rs
+++ b/src/wrapper/jnienv.rs
@@ -317,6 +317,7 @@ impl<'a> JNIEnv<'a> {
     /// `ByteBuffer`. The JVM may maintain references to the `ByteBuffer` beyond the lifetime
     /// of this `JNIEnv`.
     pub unsafe fn new_direct_byte_buffer(&self, data: &mut [u8]) -> Result<JByteBuffer<'a>> {
+        #[allow(unused_unsafe)]
         let obj: JObject = jni_non_null_call!(
             self.internal,
             NewDirectByteBuffer,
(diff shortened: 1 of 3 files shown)


src/wrapper/jnienv.rs
```diff
@@ -304,7 +304,19 @@ impl<'a> JNIEnv<'a> {
...
+    /// let direct_buffer = unsafe { env.new_direct_byte_buffer(buf.leak()) };
+    /// ```
+    ///
+    /// # Safety
+    ///
+    /// Caller must ensure the lifetime of `data` extends to all uses of the returned
+    /// `ByteBuffer`. The JVM may maintain references to the `ByteBuffer` beyond the lifetime
+    /// of this `JNIEnv`.
+    pub unsafe fn new_direct_byte_buffer(&self, data: &mut [u8]) -> Result<JByteBuffer<'a>> {
         let obj: JObject = jni_non_null_call!(
```
stanislav-tkach, line 320:
```suggestion
        #[allow(unused_unsafe)]
        let obj: JObject = jni_non_null_call!(
```

stanislav-tkach (commented):
Thanks!

stanislav-tkach (approved):

PR #321, opened 2021-06-08
get_direct_buffer_address(): Return *mut [u8]

gedgygedgy:
Returning a &mut [u8] implies that the lifetime of the underlying
data is known. Return a raw pointer instead, to acknowledge that
the lifetime may be unknown.

### Definition of Done

### DIFF
--- a/src/wrapper/jnienv.rs
+++ b/src/wrapper/jnienv.rs
@@ -316,13 +316,13 @@ impl<'a> JNIEnv<'a> {
 
     /// Returns the starting address of the memory of the direct
     /// java.nio.ByteBuffer.
-    pub fn get_direct_buffer_address(&self, buf: JByteBuffer) -> Result<&mut [u8]> {
+    pub fn get_direct_buffer_address(&self, buf: JByteBuffer) -> Result<*mut [u8]> {
         non_null!(buf, "get_direct_buffer_address argument");
         let ptr: *mut c_void =
             jni_unchecked!(self.internal, GetDirectBufferAddress, buf.into_inner());
         non_null!(ptr, "get_direct_buffer_address return value");
         let capacity = self.get_direct_buffer_capacity(buf)?;
-        unsafe { Ok(slice::from_raw_parts_mut(ptr as *mut u8, capacity as usize)) }
+        unsafe { Ok(slice::from_raw_parts_mut(ptr as *mut u8, capacity as usize) as *mut [u8]) }
     }
 
     /// Returns the capacity of the direct java.nio.ByteBuffer.


src/wrapper/jnienv.rs
```diff
@@ -316,13 +316,13 @@ impl<'a> JNIEnv<'a> {
...
     /// java.nio.ByteBuffer.
-    pub fn get_direct_buffer_address(&self, buf: JByteBuffer) -> Result<&mut [u8]> {
+    pub fn get_direct_buffer_address(&self, buf: JByteBuffer) -> Result<*mut [u8]> {
         non_null!(buf, "get_direct_buffer_address argument");
         let ptr: *mut c_void =
             jni_unchecked!(self.internal, GetDirectBufferAddress, buf.into_inner());
         non_null!(ptr, "get_direct_buffer_address return value");
         let capacity = self.get_direct_buffer_capacity(buf)?;
-        unsafe { Ok(slice::from_raw_parts_mut(ptr as *mut u8, capacity as usize)) }
+        unsafe { Ok(slice::from_raw_parts_mut(ptr as *mut u8, capacity as usize) as *mut [u8]) }
```
rib, line 325: This is still unsound I think...

This is constructing a slice on the stack (which is effectively a struct with a pointer and a size) and then returning a pointer to that slice on the stack frame which is about to be returned from.

I guess this works with some luck just so long as you only deref that pointer before calling another function that uses the stack ;-p

I guess this API should just return a `Result<*mut u8>` (i.e. cast `ptr` after the null pointer check) and the caller will need to determine the bounds separately e.g. via `get_direct_buffer_capacity`

rib (changes requested):

src/wrapper/jnienv.rs
maurolacy: Just seen this. I think you're right, that's working out of luck / call patterns; and compiling because of the `unsafe`.

I think the idea of returning just the wrapped pointer is good. It's also more in line with the method's name.

maurolacy: In fact, if returning a pointer, why build the slice from raw parts at all?

maurolacy: > I guess this works with some luck just so long as you only deref that pointer before calling another function that uses the stack ;-p

Unless... the slice type is "special", and the (still cheap) fat pointer under `&[]` is being treated as a kind of value by the compiler, and copied to the caller's stack. I mean, that `&` in this case doesn't mean a reference in the usual sense, but it is more like convenience syntax (`&[]`) / syntactic sugar for a **value** of type "slice".

Will definitely need to check / test this to understand it better.

rib: > In fact, if returning a pointer, why build the slice from raw parts at all?

right; conceptually creating a slice on the stack here (no matter its representation) is unnecessary since the combined information can't be returned in a sound way through a single pointer.

maurolacy: As I suspected, slices are primitive types in Rust. See by example https://stackoverflow.com/a/71517192/3768429.

So, that code works because it's not returning a reference, but (erm) a _slice_. Which is something like an integer, and has value semantics.

rib: > > I guess this works with some luck just so long as you only deref that pointer before calling another function that uses the stack ;-p
> 
> Unless... the slice type is "special", and the (still cheap) fat pointer under `&[]` is being treated as a kind of value by the compiler, and copied to the caller's stack. I mean, that `&` in this case doesn't mean a reference in the usual sense, but it is more like convenience syntax (`&[]`) / syntactic sugar for a **value** of type "slice".
> 
> Will definitely need to check / test this to understand it better.

hmm, right, actually since the `[u8]` is a dynamically sized type then the original `&mut [u8]` would have basically been a fat pointer that would have combined the pointer and size.

So even though the previous method looks suspect from the pov of the lifetime of the returned slice it was effectively returning a tuple containing both the pointer into the JByteBuffer and its capacity.

Oof, sometime else I didn't know about Rust raw pointers is that they can _also_ be fat if they point to dynamically sized types, hmmm.

E.g. testing this assertion passes:
```rust
assert_eq!(std::mem::size_of::<Result<*mut [u8]>>(), std::mem::size_of::<Result<&mut [u8]>>());
```
and this fails:
```rust
assert_eq!(std::mem::size_of::<*mut [u8]>(), std::mem::size_of::<usize>());
```
On x64 then a `*mut [u8]` is 16 bytes.

I was definitely assuming that a raw pointer couldn't be fat, hmmm.

So actually it seems like this doesn't end up pointing back into the old stack frame like I was thinking but maybe does highlight that this is relying on some subtle, non-obvious details and we should maybe either:
1. Much more clearly documented (at least some comments in the code to explain exactly why this is considered safe, and maybe some assertions for in-variants that are being relied on).
2. Change this API to more-simply return a `*mut u8` instead and let the caller create their own slice if that's what they want (this would still be my preference since I think it would also better match the name of the method too)

maurolacy: Both assertions are failing for me (MAC M1):

```
$ evcxr
Welcome to evcxr. For help, type :help
>> use std::result::Result;
>> use core::fmt::Error;
>> 
>> assert_eq!(std::mem::size_of::<Result<*mut [u8], Error>>(), std::mem::size_of::<Result<&mut [u8], Error>>());
thread '<unnamed>' panicked at 'assertion failed: `(left == right)`
  left: `24`,
 right: `16`', src/lib.rs:7:1
```

Funny that an array pointer is fatter than a slice when wrapped. Some crazy optimisation going on with slices.

And
```
>> assert_eq!(std::mem::size_of::<*mut [u8]>(), std::mem::size_of::<*mut u8>());

thread '<unnamed>' panicked at 'assertion failed: `(left == right)`
  left: `16`,
 right: `8`', src/lib.rs:7:1
>>
```
It seems pointers to arrays are being handled internally like slices (fat pointers), as you said.

Given that this works, I vote for leaving it just as it is. Returning a slice is definitely better than returning an array pointer, both in terms of space, and safeness. And, the overall unsafeness is related to the lifecycle of the original buffer, which is our of scope / unrelated to the slice.

Documenting this is definitely a good idea, yes. Funny that I got this right when I implemented it, given that only now I understand what's going on under the hood. ;-)

rib: I think the case for changing this to return a `*mut u8` seems quite compelling compared to leaving this as is:

1. It seems like this is working partly through luck currently, considering the roller-coaster discussion we've had to investigate what is happening here. I think this is a good sign that others will also not be familiar with what is being relied on here.
2. I'm not currently sure if this relies on undefined behaviour or not. (e.g. I'm wondering about aliasing rules and the possibility that something could could call `get_direct_buffer_address` multiple times (if I'm not mistaken) and get a bunch of aliasing mutable slices to the same buffer?)
3. Since there is already the `get_direct_buffer_capacity` method then for minimal symmetry and to only do what the name implies then just returning the address pointer itself, instead of the address and size (via a slice) sounds simpler + less surprising to me.
4. The original issue description still stands here I think; which is that returning a reference implies that the slice has a lifetime that's implicitly connected with the JNIEvent &self reference that's given to this method, which isn't going to be a correct representation of its lifetime.

rib: Trying to clarify my understanding of what the real lifetime of the underlying buffer is I'm looking at `new_direct_byte_buffer` and a bit worried that that's no safe / correct...

```rust
    pub fn new_direct_byte_buffer(&self, data: &mut [u8]) -> Result<JByteBuffer<'a>> {
        let obj: JObject = jni_non_null_call!(
            self.internal,
            NewDirectByteBuffer,
            data.as_mut_ptr() as *mut c_void,
            data.len() as jlong
        );
        Ok(JByteBuffer::from(obj))
    }
```

My understanding is that `NewDirectByteBuffer` doesn't allocate its own storage or copy anything from the given address (e.g. it can be used to wrap mmaped files and it's the responsibility of native code to ensure the associated buffer stays valid so long as it's accessible from java) so it's similarly confusing this is passed a slice with an implied lifetime that is then pulled apart to get a pointer and size for a buffer that has a completely disconnected lifetime.

maurolacy: I sincerely think this is much ado about nothing.

Handling pointers from other language's memory space will by definition be an unsafe operation. Nothing will change that. And I don't think "signalling" that by returning a pointer instead of a slice will make much of a difference. Particularly when they have pretty similar semantics, and handling a slice is a safe operation as we've seen.

It would be interesting to check what advantages returning a slice brings. Perhaps the compiler can use the capacity information for signalling and preventing some errors / overflows at compile time.

The whole argument of this PR, that returning a pointer is better because it's "explicitly unsafe" is moot in my opinion. `get_direct_byte_buffer()` _is already_ an explicitly unsafe thing to do.

rib: I think it's maybe not so much the unsafety of messing with pointers that's too concerning (as you say there is an inherent amount of risk that's accepted with that) but there are a few things here that feel like they may be straying into undefined behaviour for the Rust language and I think that's more concerning for us to check. In particular with the correctness of lifetimes and mutable references that alias.

With Rust's `unsafe` mechanism we have to assume responsibility to avoid undefined behaviour that can potentially break assumptions in the compiler.

One of the big ones that's easy to get caught out by is the no-aliasing rule for mutable references - and can lead to miss-compiled code when the compiler generally assumes that if it has a mutable reference to something then it can trust that it doesn't alias some other mutable reference.

For example a function that like `fn copy(src: &mut [u8], dest: &mut [u8])` in Rust can assume that `src` and `dest` don't overlap and that could affect how it might be inlined / optimized etc. 

I'm more concerned here that we've got an API that looks like it will let you generate any number of mutable slices to the exact same memory which would be breaking those aliasing rules and could lead to incorrectly compiled code.

And then I think the other concern here is with the associated lifetimes for these slices. It seems a bit like slices are being used as a tool to pack together a pointer and a size currently (both for creating a buffer and for querying the buffer's address) but the lifetime that Rust is tracking is actually not related to the real lifetime of the buffer (I think this is what the original issue was opened for).

Hope that makes sense.

rib: > All that said, I'm OK with changing the signature to returning an array pointer, if you think it's better / worth it.

I think that sounds best to me atm - it'd be great if you're happy to update this patch!

I think we should also consider a similar change to `new_direct_byte_buffer` so it takes a pointer and size instead of a slice too. Since that's also about the lifetime it could be reasonable to deal with this together.

rib: Since I just came across this separate issue / PR I'll also link them here while they also seem relevant here: https://github.com/jni-rs/jni-rs/issues/303 and https://github.com/jni-rs/jni-rs/pull/320

rib: Also found this PR is very closely related to this too: https://github.com/jni-rs/jni-rs/pull/351

maurolacy: I'm still not convinced. You are free to change it of course. Just don't expect me to do it, and / or be happy about it. :-)

rib: okey, sorry, earlier you said "I think the idea of returning just the wrapped pointer is good. It's also more in line with the method's name." and I thought were offering to update it.

do you not agree with my concern that there may be UB as a result of multiple mutable references, or maybe don't agree with the concerns around the associated lifetime?

I suppose I feel pretty convinced atm that the current API is a high risk for UB so I'm pretty inclined to go ahead with making the changes above - but I may have overlooked something still.

for now I'm just going through existing PRs, (especially the simpler ones to start with) and aiming to rebase and merge those and I can circle back to this one later.

For now I've just merged the two PRs I linked above which don't change this, but they do make some related changes to the new_direct_byte_buffer API.

Maybe it'd be possible to use Miri to get a clearer answer about the UB concerns.

maurolacy: I agreed with returning just the pointer while thinking the current impl was broken.

Now that we've seen it's not, I'm not so sure. And then, don't see a good / compelling reason for the change anymore.

Having multiple mutable references is in the end equivalent to something you would be able to do with raw pointers as well. Just don't do it.
 
Now, using Miri sounds great. That, or a code snippet that shows a specific problem, limitation, or bug related to the current impl, as compared with raw pointers.

rib: > Having multiple mutable references is in the end equivalent to something you would be able to do with raw pointers as well.

just to note here that Rust specifically doesn't assume noaliasing with raw pointers and won't make the same assumptions as it does for mutable references. So it's not considered UB to have aliasing mutable (raw) pointers to something (in the sense that the compiler will make sure to avoid assuming noaliasing when optimizing your code).

If you're familiar with C then it's a bit like Rust defaults to adding the `restrict` keyword for mutable references but doesn't for raw pointers.

I can try and dig up some pointers for this if your not convinced by my assertions about aliasing mutable references.

rib: I thought this was a good watch recently for comparing how Rust and C/C++ handle these aliasing rules - though it's quite a long video it has some nice examples: https://www.youtube.com/watch?v=DG-VLezRkYQ the description also links to some godbolt examples.

There is also some discussion of the importance of avoiding aliasing in the rustonomicon here: https://doc.rust-lang.org/stable/nomicon/aliasing.html

Here's the list of things we need to be careful of in unsafe code: https://doc.rust-lang.org/stable/nomicon/what-unsafe-does.html and the second rule is to not "Breaking the pointer aliasing rules".

maurolacy: Now that makes sense. Using slices would be like adding `restrict` in C, when you shouldn't.

It will be nice to try and make this fail. By example by trying something like the optimisation in https://doc.rust-lang.org/stable/nomicon/aliasing.html over a direct buffer.

rib: my guess at the easiest way of trying to show an issue would be to use the `get_direct_buffer_address` api to get two slices that point to the same memory and then try and do a copy to different (but overlapping) sub-slices.

Rust is allowed to assume that if it's given two slices to copy between that they don't overlap. An optimized copy function, e.g. using SIMD type instructions may copy contiguous, aligned chunk from region A to B without synchronizing how it reads and writes from the overlapping region between the subslices.

On the other hand a similar copy function in Rust that took raw pointers would not be allowed to assume that the source and destination don't overlap and, like `memmove` in C it should take care about how it orders reads and writes for any overlap.

At least in theory I'm fairly sure that's one way the current API could lead to UB but I'm not sure how fiddly it will be to demonstrate in practice.

maurolacy: Tried to "miri" `jni-rs` to no avail, as ffi support is not there yet.

It's being worked on. See https://github.com/rust-lang/miri/issues/2365 and https://docs.google.com/document/d/1JuvORO1_EtYBH1IsjJroxIMbWAgSGXOWvRiQyvF443I.

So, perhaps we can test and confirm UB in `jni-rs` with miri soon!

rib: oh, interesting yeah.

I guess if we still wanted to poke at the situation here with miri we could potentially try and come up with an approximation of the buffer APIs but without and JNI.

For the issue I was thinking of above I don't think it needs to necessarily involve any JNI. a stub `JByteBuffer` that just holds an address pointer and a size could be enough, along with a stub JNIEnv buffer API.

either way though, will be good to keep an eye on that work. I'm also working on android atm so nice to see that that's the OS they are targeting with that.

rib: Okey, as a small illustration I just checked that this compiles:

```rust
pub fn direct_buffer_address_undefined_behaviour() {
    let env = attach_current_thread();
    let mut vec: Vec<u8> = vec![0, 1, 2, 3];
    let buf = unsafe { env.new_direct_byte_buffer(vec.as_mut_slice()) }.unwrap();

    let alias_a = env.get_direct_buffer_address(buf).unwrap();
    let alias_b = env.get_direct_buffer_address(buf).unwrap();

    assert_eq!(alias_a, alias_b);
    assert_eq!(alias_a, vec.as_mut_slice());
}
```

This gives an example of how easily the API lets you break out of Rust's memory safety guarantees by getting any number of mutable references to the same buffer from a single thread. 

Prior to landing #320 the other day the same thing was also possible without needing the `unsafe` around `new_direct_byte_buffer()`.

I've got [a patch](https://github.com/jni-rs/jni-rs/commit/994be46345e1d69fb92cda295833883b01824647) now that hopefully addresses this and I'll look at making a PR in a bit.

PR #335, opened 2022-03-15
Implement Kotlin variants of get/set/take_rust_field

### LINKED ISSUE #333: {get,set,take}_rust_field not compatible with Kotlin
Hi,

Due to the way Kotlin works internally, `get_rust_field`, `set_rust_field`, and `take_rust_field` are not always compatible.

Kotlin uses classes for storage of types rather than primitives, specifically `Long` in this case, and it converts to a primitive "if it can" (so won't always).

If you declare a non-nullable Kotlin field `private var foo: Long = 0`, then the bytecode generated will generate a Java primitive `J`, however, your functions expect to be able to check/set based on the field being `null` or not. In Kotlin, types are not nullable unless declared explicitly as such. So the correct way to declare the field would be `private var foo: Long? = null`. Unfortunately, due to what I mentioned before, this causes Kotlin's bytecode to generate the type as `Ljava/lang/Long;`, which means the code expecting type `J` no longer works.

The way to get and set the field in this case, for Kotlin specifically, would be
```rust
    let res = env.get_field(obj, "handle", "Ljava/lang/Long;").unwrap();
    // Ljava/lang/Long; object
    let ob = res.l().unwrap();
    // the primitive long is stored in the `value` field
    let value = env.get_field(ob, "value", "J").unwrap();
    println!("res {:?} ", value);
    // now we can just set this value field, and it'll be reflected back in Java
    env.set_field(ob, "value", "J", JValue::Long(1234)).unwrap();
```
As for setting it back to null (or checking for null), that's obvious since the type was nullable. Just set `obj` to null.

Note: It would be theoretically possible to set a field to null even though it wasn't marked as such in Kotlin. However, since kotlin won't allow null unless specified, the null check would still fail (since we had to fill it with a real value). Not withstanding the fact that the field would be null but the Kotlin typechecker wouldn't know anything about it (creating false positives - not a good idea)

MolotovCherry:
## Overview

This PR implements the Kotlin version of the get/set/take_rust_field functions.

For more info on why this is required, please see #333 , but to sum it up, the current methods rely on `null` and `J` field, which Kotlin can't use due to the way it works. Kotlin represents all primitive types as classes, and internally it may or may not use a `J`. While in Kotlin, using a `Long` results in a `J` internally, Kotlin does not allow `null` unless you specify it as a nullable type `Long?`. When you do this, it internally changes to use `java/lang/Long` instead of primitives, thus rendering the original functions incompatible.

Technically it IS possible to let it use the non-nullable `Long` type which results in primitive `J`, however we'd be setting null under the hood without any typecheck for the user (bad bad idea, besides, Kotlin may do optimizations based on the fact that it's never null, and we would be violating the types contract). The user could assume that the field can never be null, which would lead to bad situations.

(Completely unrelated note, but these get/set/take functions should all be marked `unsafe`, because they require the caller to uphold a guarantee that `T` is valid, otherwise it'll be UB. Thus, the unsafety is spilling onto the callers side)

(This fixes #333)

### Definition of Done

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Added
 - `Default` trait implemented for `JObject`, `JString`, `JClass`, and `JByteBuffer` (#199)
+- Kotlin specific methods for working with Rust objects and JNI: `take_rust_field_kt`, `get_rust_field_kt`, `set_rust_field_kt`
 
 ### Changed
 - The `release_string_utf_chars` function has been marked as unsafe. (#334)
--- a/src/wrapper/jnienv.rs
+++ b/src/wrapper/jnienv.rs
@@ -21,7 +21,7 @@ use crate::{
     sys::{
         self, jarray, jboolean, jbooleanArray, jbyte, jbyteArray, jchar, jcharArray, jdouble,
         jdoubleArray, jfloat, jfloatArray, jint, jintArray, jlong, jlongArray, jobjectArray,
-        jshort, jshortArray, jsize, jvalue, JNINativeMethod,
+        jshort, jshortArray, jsize, jvalue, JNINativeMethod, jobject
     },
     JNIVersion, JavaVM,
 };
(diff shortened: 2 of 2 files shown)


maurolacy:
I don't have time / enough knowledge of the impls to review it, but thanks for your contribution. 

Feel free to create another issue for addressing the doc comments.

cherryleafroad:
> I don't have time / enough knowledge of the impls to review it, but thanks for your contribution.
> 
> Feel free to create another issue for addressing the doc comments.

Yeah, the doc comments aren't really related to this PR honestly, just wanted to make sure I said that before I forgot. I really should put that in a separate issue

MolotovCherry:
> Hi @MolotovCherry although [I tried to incorporate your suggestions](https://github.com/jni-rs/jni-rs/pull/367) for improving the docs with these APIs in the 0.20 release, sorry that I didn't try and resolve this issue/PR before the release.
> 
> Although I'm not familiar with Kotlin; your description of a non-nullable Long field sounds like it would result in a suitable ("J") field for using these APIs from Rust.
> 
> The Rust code simply stores `null` values as a zero in the `jlong` and there's no expectation that null is understood or treated as a special case on the Java/Kotlin side - it should just be a Long with the value zero from the pov of Kotlin.
> 
> There shouldn't be any concern around how Kotlin optimizations for non-null fields will interact with this API because we aren't setting a null value as far as Kotlin is concerned. A `null` pointer on the Rust side just corresponds to storing the number zero in a long field as far as Java or Kotlin sees which isn't anything unusual/special as far as optimizations go.
> 
> Of course the whole interface is pretty dangerous as you've mentioned (at least the API is now marked as `unsafe`) but the only real requirement the Rust code has is that the field is long enough to store a pointer which a `jlong` field should always be.
> 
> Conceptually I think it's generally expected that the fields would only be read/updated from Rust but if they are touched from Java/Kotlin the the long value should be treated as an opaque handle.
> 
> From the Java/Kotlin side then zero could be considered a special empty handle to determine whether some Rust value has been attached but the field values probably shouldn't be modified from the Java/Kotlin side - so hopefully it works to just treat them as opaque handles in almost all situations.
> 
> A corner case might be where a Java/Kotlin object needs to be copied and it would be important to zero initialize these fields instead of copying them to make sure that Rust will recognize that the copied object doesn't have an associated Rust value.
> 
> Hopefully that makes sense, or else I've maybe misunderstood something here?

That makes sense, and is something I didn't originally consider. I can see it would be preferable to leave it as 0 instead of representing as null. I believe Kotlin should be able to use the primitive in a case such as this, which should render it compatible in that case.

Also, thanks a lot for taking care of the concerns around unsafety as well as the other comments!

At this current point, even if this PR's code is no longer needed, I'd probably want to see an example in the doc messages for Java and Kotlin, explaining how it sets things, so that it's easy to use from both (Java is pretty obvious, but Kotlin was a bit more confusing due to the seemingly magic way it works).

rib:
> At this current point, even if this PR's code is no longer needed, I'd probably want to see an example in the doc messages for Java and Kotlin, explaining how it sets things and the fields you need to provide, so that it's easy to reference for either one (Java is pretty obvious, but Kotlin was a bit more confusing due to the seemingly magic way it works).
> 
> The kotlin example of a field should be fairly simple now. `private var handle: Long = 0`

Thanks, yeah I've taken a pass at adding Java + Kotlin examples for how to declare a suitable property [here](https://github.com/jni-rs/jni-rs/pull/369) and hopefully that will help better explain this API. Please let me know if you think the revised docs are ok.

> 
> (I think that it might be taboo to try to alter private fields from jni though)

Looking in to this, then as far as I can tell then JNI basically isn't aware of whether fields are public/private. It could be a matter of preference but one benefit of using a private field might be that you'd reduce the chance of rogue Java/Kotlin code interfering with these associated handles/pointers which could lead to use-after-free errors if they are copied or leaks if they are cleared.

MolotovCherry:
> > At this current point, even if this PR's code is no longer needed, I'd probably want to see an example in the doc messages for Java and Kotlin, explaining how it sets things and the fields you need to provide, so that it's easy to reference for either one (Java is pretty obvious, but Kotlin was a bit more confusing due to the seemingly magic way it works).
> > The kotlin example of a field should be fairly simple now. `private var handle: Long = 0`
> 
> Thanks, yeah I've taken a pass at adding Java + Kotlin examples for how to declare a suitable property [here](https://github.com/jni-rs/jni-rs/pull/369) and hopefully that will help better explain this API. Please let me know if you think the revised docs are ok.
> 
> > (I think that it might be taboo to try to alter private fields from jni though)
> 
> Looking in to this, then as far as I can tell then JNI basically isn't aware of whether fields are public/private. It could be a matter of preference but one benefit of using a private field might be that you'd reduce the chance of rogue Java/Kotlin code interfering with these associated handles/pointers which could lead to use-after-free errors if they are copied or leaks if they are cleared.

I read it awhile back. I think I was mistakenly thinking of `final`. Thanks for the clarification in docs!

Issue #270, opened 2020-10-10
GlobalRef for JMethodID


maxv-rust:
I have a question about jni-rs's architecture.

There is [JMethodID](https://docs.rs/jni/0.17.0/jni/objects/struct.JMethodID.html) struct which bounded to the lifetime.

* From my point of view it's not clear why it's bounded to lifetime? It looks like JMethodID can be safely saved and reused with another JNIEnv. I didn't found proof in the official documentation but it was covered as a part of [SO question](https://stackoverflow.com/questions/2093112/why-i-should-not-reuse-a-jclass-and-or-jmethodid-in-jni)
* Even if JMethodID has to be bounded, what should be the right way to create GlobalRef to reuse the same JMethodID multiple times?

Please let me know if I should you another way of communication for that kind of question.

Thank you in advance for your help.

maxv-rust:
It looks like `JMethodID` is not `Sync` as well :-(

dmitry-timofeev:
> I didn't found proof in the official documentation

I think this could work: https://docs.oracle.com/en/java/javase/11/docs/specs/jni/design.html#accessing-fields-and-methods 

> A field or method ID does not prevent the VM from unloading the class from which the ID has been derived. _After the class is unloaded, the method or field ID becomes invalid._

I.e., these IDs (pointers) are valid until the class is unloaded, which is not directly related to thread attachment (= `JNIEnv` lifetime).

> Even if JMethodID has to be bounded, what should be the right way to create GlobalRef to reuse the same JMethodID multiple times?

Please note that they are not JNI references to Java objects (like JObject/GlobalRef), hence they must not be used as such. To keep them alive beyond JNIEnv, you must keep the corresponding class alive with a GlobalRef.

> It looks like JMethodID is not Send/Sync as well :-(

It is a pointer type, and JVM must not restrict their usage from different threads: https://docs.oracle.com/en/java/javase/11/docs/specs/jni/types.html#field-and-method-ids 

If they have to be Send/Sync to use them safely (I am not an expert in Rust, @stanislav-tkach ?), then it must make sense to add that. I see Exonum JNI cache used some unsafe, though it is a thread-safe implementation that guarantees their visibility in any thread with Once: https://github.com/exonum/exonum-java-binding/blob/4e00cd8cbae198ac0e8a49cb1405092537f306bc/exonum-java-binding/core/rust/src/utils/jni_cache.rs#L209

stanislav-tkach:
[...]

PR #346, opened 2022-06-06
JMethodID: remove lifetime and impl Send + Sync

rib:
## Overview

Method IDs by their nature are intended to be cached by applications
which was generally precluded by the lifetime parameter. According
to the JNI spec method IDs are also not garbage collected as previously
documented and they remain valid until their corresponding class is
unloaded. Since the real lifetime details can't be represented via a
Rust lifetime (and it would be an excessive cost at this level of
abstraction to pair every method ID with a global reference to the
class) then this patch removes the Rust lifetime and adds "Safety"
documentation to explain what developers need to do if caching method
IDs.

It's perhaps also worth noting that the very conservative lifetime
parameter that existed before wouldn't have provided any technical
safety against a method's class being unloaded within that lifetime.

As method IDs are not tied to a thread / JNIEnv it also makes sense to
declare them as Send + Sync safe.

Fixes: #38 
Fixes: #270  
Closes: #274 

### Definition of Done

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - The lifetime of `AutoArray` is no longer tied to the lifetime of a particular `JNIEnv` reference. (#302)
 - Relaxed lifetime restrictions on `JNIEnv::new_local_ref`. Now it can be used to create a local
   reference from a global reference. (#301 / #319)
+- `JMethodID` implements `Send` + `Sync` and no longer has a lifetime parameter, making method
+  IDs cacheable (with a documented 'Safety' note about ensuring they remain valid). ([#346](https://github.com/jni-rs/jni-rs/pull/346))
 
 ## [0.19.0] — 2021-01-24
 
--- a/benches/api_calls.rs
+++ b/benches/api_calls.rs
@@ -38,7 +38,7 @@ fn jni_abs_safe(env: &JNIEnv, x: jint) -> jint {
 fn jni_call_static_unchecked<'c, C>(
     env: &JNIEnv<'c>,
     class: C,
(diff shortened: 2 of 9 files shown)


stanislav-tkach:
I agree with your argumentation, though it is still bothers me that “Rust safe guaranties” are compromised here, though I'm not sure how to fix that and preserve the ergonomics. Perhaps we can have two separate types: safe JMethodID paired with a global reference to the class and  `UnsafeJMethodID` with the proposed implementation. But I am afraid it complicates both the library API definition and its usage.

rib:
Yeah, personally here I feel like it's reasonable to document how the IDs remain valid so long as the class isn't unloaded which I think is too high level of a thing to try and forcibly block at this level of abstraction.

On Android my understanding is that classes can't be unloaded so at least there then there's no risk of a method ID becoming invalid.

In case there is a real concern for some use case then I'd expect it's much easier for specific applications/libraries to create a single global ref for whatever class they are concerned could be unloaded, which wouldn't add require any kind of hidden cost/complexity for the method ids.

To some extent I feel like there's an inherent amount of unsafety I would expect with any ffi layer and although it's nice to have an abstraction like jni-rs that can guide you away from shooting yourself in the foot I also think performance + ergonomics are important at this level too.

In this case the risk is nuanced too. Since you would actually need a class reference to try and use the method ID then I'd hazard a guess that it's likely implied that the class you care about can't realistically be unloaded. For something to go wrong here, your class actually needs to get unloaded and then _reloaded_ (because otherwise how would you have a class reference to make a method call).

I'd probably bet that in any situation where some important class can theoretically be unloaded then whatever code that's looking up method IDs will also have been destroyed and would only be reinitialized after the class is reloaded - just guessing though.

rib:
Just noticed that the benchmarks aren't built by CI and `benches/api_calls.rs` needs some fixes before landing this

PR #351, opened 2022-06-23
Allow creation of byte buffer from mut pointer

TimeToogo:
## Overview

It can be useful to create a new byte buffer directly from a pointer and length if a `&mut [u8]` is not available.
In my case I am trying to share an immutable `&[u8]` to my java code and enforce immutable using [asReadOnlyBuffer](https://docs.oracle.com/javase/10/docs/api/java/nio/ByteBuffer.html#asReadOnlyBuffer()).
According to the [rustnomicon](https://doc.rust-lang.org/nightly/nomicon/transmutes.html) transmuting `&[u8]` to `&mut [u8]` is always UB, even if it doesn't get mutated so being able to initialise directly from the raw pointer is needed.

### Definition of Done

### DIFF
--- a/src/wrapper/jnienv.rs
+++ b/src/wrapper/jnienv.rs
@@ -317,12 +317,21 @@ impl<'a> JNIEnv<'a> {
     /// `ByteBuffer`. The JVM may maintain references to the `ByteBuffer` beyond the lifetime
     /// of this `JNIEnv`.
     pub unsafe fn new_direct_byte_buffer(&self, data: &mut [u8]) -> Result<JByteBuffer<'a>> {
+        self.new_direct_byte_buffer_raw(data.as_mut_ptr(), data.len())
+    }
+
+    /// Create a new instance of a direct java.nio.ByteBuffer from a pointer and size directly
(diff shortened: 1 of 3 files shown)


src/wrapper/jnienv.rs
```diff
@@ -305,11 +305,21 @@ impl<'a> JNIEnv<'a> {
 
     /// Create a new instance of a direct java.nio.ByteBuffer.
     pub fn new_direct_byte_buffer(&self, data: &mut [u8]) -> Result<JByteBuffer<'a>> {
+        unsafe { self.new_direct_byte_buffer_raw(data.as_mut_ptr(), data.len()) }
+    }
+
+    /// Create a new instance of a direct java.nio.ByteBuffer from a pointer and size directly
+    #[allow(unused_unsafe)]
```
stanislav-tkach, line 312: Does this attribute really needed?

stanislav-tkach (approved):

src/wrapper/jnienv.rs
stanislav-tkach: Nevermind. 🙃

Issue #197, opened 2019-07-31
Unsound conversion methods from jni_sys::jobject to jni::JObject and other wrappers


MaulingMonkey:
`jni_sys::jobject` and friends are simply raw pointers, whereas `jni::JObject` is generally assumed to be a valid object.  This causes soundness problems if there are safe conversion methods from the former to the latter - this crashes in an access violation without any `unsafe`:
```rust
use jni_sys;
use jni;

fn main() {
    let args = jni::InitArgsBuilder::new().version(jni::JNIVersion::V8).build().unwrap();
    let vm = jni::JavaVM::new(args).unwrap();
    let env = vm.attach_current_thread().unwrap();
    let bad_object = jni::objects::JObject::from(42 as jni_sys::jobject);
    let hash_code = env.call_method(bad_object, "hashCode", "()I", &[]).unwrap().i().unwrap();
    println!("Hello, world!  bad_object.hashCode(): {}", hash_code);
}
```

Thanks to conversion methods like this: https://github.com/jni-rs/jni-rs/blob/3ad3c9db5d74e1c693e4f523ceb7654d3defad7d/src/wrapper/objects/jobject.rs#L20-L27

I'd recommend replacing these with unsafe fns ala `JavaVM::from_raw`... which unfortunately will be a breaking change.

DarkEld3r:
Thanks for the observations. It is definitely a soundness issue.

PR #362, opened 2022-10-15
Replace `From<jni pointers>` with `unsafe` `::from_raw()` functions

rib:
This ensures that it's not possible to trivially convert arbitrary numbers as invalid pointers into types like JObject, JString, JClass etc in safe code due to `From` trait implementations for `jni_sys` types.

I.e this kind of thing shouldn't be possible in safe Rust:

```rust
    let bad_object = jni::objects::JObject::from(42 as jni_sys::jobject);
    let hash_code = env.call_method(bad_object, "hashCode", "()I", &[]).unwrap().i().unwrap();
```

This removes the `From<jni pointer>` trait implementation for `JObject`-based types and method and field IDs.

The same types now consistently have an `unsafe` `::from_raw()` functions instead of implementing the `From` trait and a corresponding `::into_raw()` method. Any pre-existing `::into_inner()` method was renamed to `::into_raw()` for consistency / symmetry.

Fixes: #197

### Definition of Done


rib:
@maurolacy I wonder if you might be able to take a look over this if possible? It'd be nice to hopefully have another pair of eyes check it before it lands.

rib:
okey, I'll go ahead and land this even though we haven't found someone to review the change. The change isn't really controversial and it's also very systematic. More importantly it addresses a fairly significant soundness issue.

PR #364, opened 2022-10-15
direct_buffer: Avoid mutable reference aliasing (UB)

rib:
This aims to address some soundness issues with the direct_buffer wrapper functions which:
1. discard the slice lifetime when creating a buffer
2. enable shared mutable reference aliasing (which is considered undefined behaviour in Rust.)

For example the following was previously possible:
```rust
let mut vec: Vec<u8> = vec![0, 1, 2, 3];
let buf = env.new_direct_byte_buffer(&mut vec)?;
let slice_alias_a: &mut [u8] = env.get_direct_buffer_address();
let slice_alias_b: &mut [u8] = env.get_direct_buffer_address();
// ...
```
Where the same memory would be mutable via the original `vec` and any of the additional mutable slice aliases.

This doesn't uphold Rust's safety model since it enables shared mutable access to state from a single thread.

This patch removes the version of `new_direct_byte_buffer()` that takes a slice with a lifetime and replaces it with the recently added `new_direct_byte_buffer_raw()` implementation that takes an address pointer and size. This is an `unsafe` function that documents that the address needs to remain valid as long as the `JByteBuffer` may be accessed.

This adds some additional null pointer checks (and corresponding tests) that `new_direct_byte_buffer()` can't be provided a null address and `get_direct_buffer_capacity()` can't be given a null buffer.

This updates `get_direct_buffer_address()` to simply return the raw address pointer given to `new_direct_byte_buffer()` instead of returning a mutable slice that may lead to undefined behaviour.

If an application knows it has exclusive access to the buffer then it can use `unsafe` code to create it's own mutable slice from the raw pointer and size if needed.

Closes: #321

### Definition of Done

### DIFF
--- a/src/wrapper/jnienv.rs
+++ b/src/wrapper/jnienv.rs
@@ -1,7 +1,7 @@
 use std::{
     marker::PhantomData,
     os::raw::{c_char, c_void},
-    ptr, slice, str,
+    ptr, str,
     str::FromStr,
     sync::{Mutex, MutexGuard},
 };
@@ -304,33 +304,31 @@ impl<'a> JNIEnv<'a> {
         Ok(check)
     }
 
-    /// Create a new instance of a direct java.nio.ByteBuffer.
+    /// Create a new instance of a direct java.nio.ByteBuffer
     ///
     /// # Example
     /// ```rust,ignore
     /// let buf = vec![0; 1024 * 1024];
-    /// let direct_buffer = unsafe { env.new_direct_byte_buffer(buf.leak()) };
+    /// let (addr, len) = { // (use buf.into_raw_parts() on nightly)
+    ///     let buf = buf.leak();
+    ///     (buf.as_mut_ptr(), buf.len())
+    /// };
+    /// let direct_buffer = unsafe { env.new_direct_byte_buffer(addr, len) };
     /// ```
     ///
     /// # Safety
     ///
+    /// Expects a valid (non-null) pointer and length
+    ///
     /// Caller must ensure the lifetime of `data` extends to all uses of the returned
     /// `ByteBuffer`. The JVM may maintain references to the `ByteBuffer` beyond the lifetime
     /// of this `JNIEnv`.
-    pub unsafe fn new_direct_byte_buffer(&self, data: &mut [u8]) -> Result<JByteBuffer<'a>> {
-        self.new_direct_byte_buffer_raw(data.as_mut_ptr(), data.len())
-    }
-
-    /// Create a new instance of a direct java.nio.ByteBuffer from a pointer and size directly
-    ///
-    /// # Safety
-    ///
-    /// Expects a valid pointer and length
-    pub unsafe fn new_direct_byte_buffer_raw(
+    pub unsafe fn new_direct_byte_buffer(
         &self,
         data: *mut u8,
         len: usize,
     ) -> Result<JByteBuffer<'a>> {
+        non_null!(data, "new_direct_byte_buffer data argument");
         #[allow(unused_unsafe)]
         let obj: JObject = jni_non_null_call!(
             self.internal,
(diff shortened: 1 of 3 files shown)


rib:
Considering the discussion under https://github.com/jni-rs/jni-rs/pull/321, it'd be good if you wouldn't mind taking a look at this @maurolacy

src/wrapper/jnienv.rs
```diff
@@ -304,33 +304,31 @@ impl<'a> JNIEnv<'a> {
...
-    /// # Safety
-    ///
-    /// Expects a valid pointer and length
-    pub unsafe fn new_direct_byte_buffer_raw(
+    pub unsafe fn new_direct_byte_buffer(
         &self,
         data: *mut u8,
         len: usize,
     ) -> Result<JByteBuffer<'a>> {
+        non_null!(data, "new_direct_byte_buffer_raw data argument");
```
maurolacy, line 331:
```suggestion
        non_null!(data, "new_direct_byte_buffer data argument");
```

src/wrapper/jnienv.rs
```diff
@@ -343,21 +341,28 @@ impl<'a> JNIEnv<'a> {
 
     /// Returns the starting address of the memory of the direct
     /// java.nio.ByteBuffer.
-    pub fn get_direct_buffer_address(&self, buf: JByteBuffer) -> Result<&mut [u8]> {
+    pub fn get_direct_buffer_address(&self, buf: JByteBuffer) -> Result<*mut u8> {
```
maurolacy, line 344: 👍🏼

src/wrapper/jnienv.rs
```diff
@@ -343,21 +341,28 @@ impl<'a> JNIEnv<'a> {
...
         non_null!(buf, "get_direct_buffer_address argument");
-        let ptr: *mut c_void =
-            jni_unchecked!(self.internal, GetDirectBufferAddress, buf.into_inner());
+        let ptr = jni_unchecked!(self.internal, GetDirectBufferAddress, buf.into_inner());
         non_null!(ptr, "get_direct_buffer_address return value");
-        let capacity = self.get_direct_buffer_capacity(buf)?;
-        unsafe { Ok(slice::from_raw_parts_mut(ptr as *mut u8, capacity as usize)) }
+        Ok(ptr.cast())
```
maurolacy, line 348: What an expensive way to write `ptr as _`. ;-)

maurolacy (approved):
Looks good.

My only "concern" would be, I remember I measured the performance of these direct buffers in its time, and they were blazingly fast.

I hope they continue to be. And in any case, this is the right thing to do: first correctness, then performance.

src/wrapper/jnienv.rs
rib: yeah I guess I use `as _` and `cast()` arbitrarily, interchangeably - couldn't really say how come I used `cast` in this case

rib:
> Looks good.
> 
> My only "concern" would be, I remember I measured the performance of these direct buffers in its time, and they were blazingly fast.
> 
> I hope they continue to be. And in any case, this is the right thing to do: first correctness, then performance.

yeah, it's hard to imagine it making a measurable difference in most circumstances.

If there were a measurable difference for some operations that would presumably come down to the difference in what the compiler is allowed to assume/optimize between references vs pointers but previously those optimizations would have come at a risk of a mis-compiled / undefined program due to the violation of the aliasing rules.

If an application or higher-level abstraction did know that it had _exclusive_ access to a buffer it could potentially still use `unsafe` code to take the pointer and size and create its own slice in case that helped with using some better optimized code somewhere.

src/wrapper/jnienv.rs
rib: ended up switching it when I made the other fix for the comment

rib:
thanks for the review @maurolacy

PR #368, opened 2022-10-17
Release 0.20.0

rib:
Bump the version to 0.20.0


rib:
To further test this I also checked that this https://github.com/rib/bluey/tree/main/bluey-ui app could be built and run against this release candidate, and it seems to work as expected.

Issue #328, opened 2021-08-12
JNIEnv#get_string() panics when given a non-string, rather than returning an Err


TobiasDeBruijn:
As the title suggests, the get_string fn on JNIEnv panics when given an object which is not a String.

The panic seems to occur within the JVM, so I dont think that panic can be avoided, but a check could be an option.

I'm unsure if it's intented behaviour for it to panic, likely it is, since that's just a programmer's fault. Though I suppose an invalid signature is too, which doesn't panic.

Here's the code that triggered it for me, which obviously has a bug:
```rs
        let return_type = env.call_method(method, "getReturnType", "()Ljava/lang/Class;", &[])?.l()?;
        let return_type_name: String = env.get_string(return_type.into())?.into();
```
where method is an instance of `java.lang.reflect.Method`,  I should have been calling getName, rather than trying to stringify a `java.lang.Class`.

stanislav-tkach:
I think that even if we decide against adding the check, at least the `get_string` function documentation should be updated to mention a potential panic.

mjh316:
Is there an update on this? Tried to run through the example on docs.rs, but encountered a SIGSEV from JNIEnv#get_string (not sure if it's completely related).

TobiasDeBruijn:
Only the docs have been updated. I think we should still add a check that checks the object passed in is of java.lang.String. we can then also add an unchecked variant (possibly unsafe?) Which developers van use to avoid the performance penalty.

mjh316:
For the check, I was thinking we could put
```rust
match self.is_instance_of(obj, self.find_class("java/lang/String").unwrap()) {
            Ok(false) => return Err(Error::JavaException), // <- Not sure what type of Error to return
            Err(err) => return Err(err),
            _ => (),
        };
```
after the null check. Would this be sufficient?

TobiasDeBruijn:
I think that looks OK, though does that work for classes that extend String (is that even possible, but if it is, we should allow it I'd think?), maybe we need to use something like is_assignable_from?

I think we should also have a new error type, something similar to what HTTP does with Bad Request, but then Bad Input or something?

Possible optimization: Can we put the find_class in a Once? or is that unsound due to the JNIEnv?

rib:
Just for reference here I ended up creating these tests to see what's happening here:

```rust
#[test]
pub fn test_null_get_string() {
    let env = attach_current_thread();
[...]

PR #370, opened 2022-10-19
Checked version of JNIEnv::get_string

TobiasDeBruijn:
## Overview
This PR:
- Renames `JNIEnv::get_string` to `JNIEnv::get_string_unchecked`, and marks it unsafe
- Adds a new `JNIEnv::get_string` which checks whether the object passed in is actually a java.lang.String
- Adds a new error variant describing that the type of the parameter passed in is incorrect.

See also: #328 

This PR is backwards compatible, though the code will become slightly slower due to the addition of the class check

Open questions:
- Can we optimize the `find_class` call for `java/lang/String` to be put in a once cell? Or must the lifetime of the `JClass` be tied to the `JNIEnv`'s lifetime?

### Definition of Done

### DIFF
--- a/benches/api_calls.rs
+++ b/benches/api_calls.rs
@@ -26,6 +26,9 @@ static SIG_MATH_ABS: &str = "(I)I";
 static SIG_OBJECT_HASH_CODE: &str = "()I";
 static SIG_LOCAL_DATE_TIME_OF: &str = "(IIIIIII)Ljava/time/LocalDateTime;";
 
+// 32 characters
+static TEST_STRING_UNICODE: &str = "_񍷕㳧~δ򗊁᪘׷ġ˥쩽|ņ/򖕡ٶԦ萴퀉֒ٞHy󢕒%ӓ娎񢞊ăꊦȮ񳗌";
+
 #[inline(never)]
 fn native_abs(x: i32) -> i32 {
     x.abs()
@@ -312,22 +315,22 @@ mod tests {
     #[bench]
     fn jni_get_string(b: &mut Bencher) {
         let env = VM.attach_current_thread().unwrap();
-        let string = env.new_string("test").unwrap();
+        let string = env.new_string(TEST_STRING_UNICODE).unwrap();
 
         b.iter(|| {
             let s: String = env.get_string(string).unwrap().into();
-            assert_eq!(s, "test");
+            assert_eq!(s, TEST_STRING_UNICODE);
         });
     }
 
     #[bench]
     fn jni_get_string_unchecked(b: &mut Bencher) {
         let env = VM.attach_current_thread().unwrap();
-        let string = env.new_string("test").unwrap();
+        let string = env.new_string(TEST_STRING_UNICODE).unwrap();
 
         b.iter(|| {
             let s: String = unsafe { env.get_string_unchecked(string) }.unwrap().into();
-            assert_eq!(s, "test");
+            assert_eq!(s, TEST_STRING_UNICODE);
         });
     }
 
--- a/benches/api_calls.rs
+++ b/benches/api_calls.rs
@@ -309,6 +309,28 @@ mod tests {
         });
     }
 
+    #[bench]
+    fn jni_get_string(b: &mut Bencher) {
+        let env = VM.attach_current_thread().unwrap();
+        let string = env.new_string("test").unwrap();
+
+        b.iter(|| {
+            let s: String = env.get_string(string).unwrap().into();
+            assert_eq!(s, "test");
+        });
+    }
+
+    #[bench]
+    fn jni_get_string_unchecked(b: &mut Bencher) {
+        let env = VM.attach_current_thread().unwrap();
+        let string = env.new_string("test").unwrap();
+
+        b.iter(|| {
+            let s: String = unsafe { env.get_string_unchecked(string) }.unwrap().into();
+            assert_eq!(s, "test");
+        });
+    }
+
     /// A benchmark measuring Push/PopLocalFrame overhead.
     ///
     /// Such operations are *required* if one attaches a long-running
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Changed
+- `JNIEnv::get_string` checks that the given object is a `java.lang.String` instance to avoid undefined behaviour from the JNI implementation potentially aborting the program. ([#328](https://github.com/jni-rs/jni-rs/issues/328))
+
+### Added
+- `JNIEnv::get_string_unchecked` is a cheaper, `unsafe` alternative to `get_string` that doesn't check the given object is a `java.lang.String` instance. ([#328](https://github.com/jni-rs/jni-rs/issues/328))
+
 
 ## [0.20.0] — 2022-10-17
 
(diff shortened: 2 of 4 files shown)


rib:
Thanks for looking at this. I just had a poke at #328 and wrote a quick test to reproduce the issue and check what's happening.

I think if we're updating this it could be good to further clarify that `get_string` does not "panic" if it's not passed a `String`. A panic is a Rust mechanism that can also be caught in some circumstances. It's maybe better to say that you can hit "undefined behavior". Essentially the JNI implementation may or may not recognise an error internally and either way there is no guarantee what the JNI implementation will do.

When I test on Windows I see the message "FATAL ERROR in native method: JNI string operation received a non-string" and the program immediately exits but I could imagine a different JNI implementation might behave differently.

src/wrapper/errors.rs
```diff
@@ -39,6 +39,8 @@ pub enum Error {
     ParseFailed(#[source] combine::error::StringStreamError, String),
     #[error("JNI call failed")]
     JniCall(#[source] JniError),
+    #[error("The class of the parameter passed in is invalid.")]
+    InvalidParameterClass
```
rib, line 43: It seems like it might be reasonable to use `JniError::InvalidArguments`?

I.e. 
```rust
JniCall(JniError::InvalidArguments)
```

The crash/segv/abort happens within a JNI function due to it being passed an invalid argument.

TobiasDeBruijn: Oh that is a good point actually. Missed that one!

rib:
> Open questions:
> 
> * Can we optimize the `find_class` call for `java/lang/String` to be put in a once cell? Or must the lifetime of the `JClass` be tied to the `JNIEnv`'s lifetime?

For now `JNIEnv` is a `#[transparent]` wrapper over a `JNINativeInterface_` vtable pointer and so with the current design of `jni-rs` there isn't an obvious place that could be used to cache the jclass to avoid these kinds of repeated look ups. We can't really change the representation of `JNIEnv` since it's a core part of the JNI function call ABI that we are passed a `JNIEnv` pointer when the JVM calls native methods.

Introducing global static state for caching state doesn't sound good either (I guess you're imagining the `OnceCell` in a static variable?) since that would kind of subvert the notion that a `JavaVM` encapsulates all the state for a JVM instance.

I could sort of imagine using thread local storage to track a cache of state for `jni-rs`  - but probably a bit of a rabbit hole of details to figure out.

The initial way I would imagine it:
1. add an api to create a (ref-counted) cache
2. add an api to attach/detach a cache to/from a thread
3. update `jni-rs` to internally, speculatively look up a cache for the thread whenever it needs to look up something costly and simply fallback gracefully if there's no cache.

My gut instinct is that it would need to be the responsibility of the app to deal with creating that cache and attach/detach to/from threads similar to how the app has to attach/detach threads to the JVM.

I think it could be tricky to have `jni-rs` automatically allocate this kind of cache and make sure it gets cleaned up while it doesn't have a top-level object that's a suitable place to anchor the lifetime of the cache too.

TobiasDeBruijn:
Good point. Even if we want it in `jni-rs`, it certainly is out of scope for this PR

rib:
> Good point. Even if we want it in `jni-rs`, it certainly is out of scope for this PR

right

TobiasDeBruijn:
Hm, thats a good point.

I think a big warning in the get_string function is definitely required.

Alternatively, we can use the original implementation in get_string, mark it unsafe, and add a safe get_string_checked function doing what get_string does in this branch?

rib:
> Hm, thats a good point.
> 
> I think a big warning in the get_string function is definitely required.
> 
> Alternatively, we can use the original implementation in get_string, mark it unsafe, and add a safe get_string_checked function doing what get_string does in this branch? Of course also mentioning the performance penalty

Yeah I was also half wondering about there being a `get_string_checked()` but for now I tend to think it's good to avoid the API break for `get_string` since I only published 0.20 yesterday. It will be slightly more awkward if we want to make a point release if we immediately start landing breaking API changes.

It's probably reasonable that APIs without a suffix will er on the side of doing more validation and being safe by default, but still it would be nice to highlight the cost of validation for something that will probably be used a lot and also sign post to an API that can avoid that overhead if needed.

rib:
Hey @TobiasDeBruijn - I took the liberty to squash your commits, so I could see the changes more clearly (since we e.g. backtracked on the addition of the error enum).

I ended up adding a bit more to the docs too and added myself as a co-author since it didn't seem worth splitting into multiple patches.

Interestingly while looking at the change I saw there is also a `get_string_utf_chars()` which also looks pretty concerning - I'm not really sure what the purpose of that API is and it _also_ looks super unsafe. We can consider it a separate issue probably but it does seem somewhat related too.

rib:
Oh, and I also added the unit tests and benchmark I wrote the other day.

If you're ok with the combined changes I made I think we can probably merge this - otherwise I can split my doc changes out into a separate patch if you'd prefer.

rib:
I think I'll also add a bench for `get_string_unchecked` that can then easily be run to compare the checked vs unchecked version.

TobiasDeBruijn:
Looks good to me! Feel free to merge.

We definitely need to look into `get_string_utf_chars`, I'll open an issue for thet.

benches/api_calls.rs
```diff
@@ -309,6 +309,28 @@ mod tests {
         });
     }
 
+    #[bench]
+    fn jni_get_string(b: &mut Bencher) {
+        let env = VM.attach_current_thread().unwrap();
+        let string = env.new_string("test").unwrap();
```
maurolacy, line 315: Will be nice to put this into a constant, for referencing / robustness.

Also, using a longer string with UTF-8 chars in it would be a good idea.

rib:
keeping the static test string I updated the last benchmark change.

NB: the benchmarks aren't intended as unit tests so I don't think it's really necessary to have duplicate `get_string` benchmarks where the only difference is whether the string used is all ascii or not - as far as benchmarking goes they will go through the same code.

I preferred to refer to the string as `_UNICODE` to highlight that it happens to contain non-ascii characters. Since it's a Rust `str` then it's implicit that it's represented as UTF-8.

Even though this isn't a unit test the fact that it's non-ascii seems reasonable and it doubles as a test that confirms that the string correctly survives a `Rust -> Java -> Rust` round trip.

The longer string means that the benchmark is now more weighted towards testing the cost of format conversions even though the original aim was to focus more on the difference in validation cost between the `get_string` and `get_string_unchecked` APIs. This seems reasonable though and the difference in validation cost is still clearly visible. If I was to guess then I'd probably guess that 32 chars is actually longer than the average string that applications will query but it's still short enough that we aren't purely benchmarking format conversion here.

These are the results I see with the longer test string:
```
test tests::jni_get_string                                    ... bench:       1,397 ns/iter (+/- 115)
test tests::jni_get_string_unchecked                          ... bench:         474 ns/iter (+/- 40)
```
The gap between them is closed somewhat because they are now both spending more time on format conversion, but that's fair enough.

TobiasDeBruijn:
Looks good to me. I added the ASCII and Unicode variants to check for differences in performance, though I suppose that is better tested on the conversion machinery, rather than the `get_string` function.

rib:
Yeah I think if we were worried there could be a difference with that then it would maybe warranted to add some dedicated format conversion benchmarks but my gut feeling is that that could feel perhaps a bit out of scope for `jni-rs` since we'd effectively either be benchmarking the `cesu8` crate or the JNI implementation which we can't change.

TobiasDeBruijn:
Yup good point. Will leave that out then

Issue #371, opened 2022-10-21
Get_string_utf_chars appears unsafe 


TobiasDeBruijn:
We need to look into `get_string_utf_chars`, it appears super unsafe. [More here](https://github.com/jni-rs/jni-rs/pull/370#issuecomment-1286150322)

PR #372, opened 2022-10-21
Remove `get_string_utf_chars` and `release_string_utf_chars` from `JNIEnv`

TobiasDeBruijn:
## Overview

This PR updates the documentation for `JNIEnv::get_string_utf_chars`, providing more details on safety and adding error documentation. This PR also marks the function as `unsafe`

### Open questions
- Do we want to rename this function to `get_string_utf_chars_unchecked` and add a checked variant, like done with `get_string` in #370?
- Do we want to add a newtype wrapper around the pointer with a `Drop` impl to avoid the potential memory leak? 

See also: #371 
### Definition of Done

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - The `JNIEnv::new_object_unchecked` function now takes arguments as `&[jni::sys::jvalue]` to avoid allocating, putting it inline with changes to `JniEnv::call_*_unchecked` from 0.20.0 ([#382](https://github.com/jni-rs/jni-rs/pull/382))
 - The `get_superclass` function now returns an Option instead of a null pointer if the class has no superclass ([#151](https://github.com/jni-rs/jni-rs/issues/151))
 
+### Removed
+- `get_string_utf_chars` and `release_string_utf_chars` from `JNIEnv` (See `JavaStr::into_raw()` and `JavaStr::from_raw()` instead) ([#372](https://github.com/jni-rs/jni-rs/pull/372))
+
 ## [0.20.0] — 2022-10-17
 
 ### Added
--- a/src/wrapper/jnienv.rs
+++ b/src/wrapper/jnienv.rs
@@ -1262,46 +1262,6 @@ impl<'a> JNIEnv<'a> {
         unsafe { self.get_string_unchecked(obj) }
     }
 
-    /// Get a pointer to the character array beneath a JString.
-    ///
-    /// Array contains Java's modified UTF-8.
-    ///
-    /// # Attention
-    /// This will leak memory if `release_string_utf_chars` is never called.
-    pub fn get_string_utf_chars(&self, obj: JString) -> Result<*const c_char> {
-        non_null!(obj, "get_string_utf_chars obj argument");
-        let ptr: *const c_char = jni_non_null_call!(
-            self.internal,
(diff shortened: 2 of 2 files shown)


rib:
tbh, I'm not entirely clear at the moment why this API exists. It's a very unsafe shim that makes me think that anything using this might just as well use `jni_sys` at this point?

Instead of this I wonder if it could be better if there were an unsafe way to consume a `JniStr` that would just return the internal JNI string pointer and then it's your responsibility to release that and then there would be a symmetric way of re-constructing a `JniStr` from such a pointer. That would be more like `Box::into_raw()` and `Box::from_raw()` or the similar `Arc` APIs.

TobiasDeBruijn:
That is probably the better approach yeah. Allowing users who really need the API to use it as JniStr::into_raw, JniStr::from_raw, then either JniStr::free or a Drop impl I think. While also not confusing users who don't use it yet.

TobiasDeBruijn:
I've split out the functions into a seperate struct `JniStr`. Some open questions:
- Should the JniStr hold the `JString` to use for the release function?
- Should the file maybe live in a different place (might it be more appropriate in sys/?)

rib:
Oh, I think I was imagining this would live under the existing `JavaStr`: https://github.com/jni-rs/jni-rs/blob/master/src/wrapper/strings/java_str.rs which is what `get_string` returns. Writing "JniStr" was a typo above sorry!

TobiasDeBruijn:
Ah that explains a lot! Will fix

src/wrapper/strings/java_str.rs
```diff
@@ -27,11 +28,91 @@ impl<'a: 'b, 'b> JavaStr<'a, 'b> {
...
+    ///
+    /// # Errors
+    ///
+    /// If the release call fails
+    pub unsafe fn release(mut self) -> Result<()> {
+        Self::release_string_utf_chars(&mut self)
+
+        // TODO do we even need this function? Or is it better to just use `Drop` for this?
+        // TODO do we need to use a ManuallyDrop here to avoid running the `Drop` code?
+    }
```
rib, line 97: Right, I don't think we should have this API. That would be a bit like if `Box` had a `free()` method that would free the backing memory which would leave it in an inconsistent state.

rib:
Yeah this update generally looks good to me, except for the `release` method which I don't think we need.

rib:
Can you perhaps squash the changes into a single commit considering the significant changes in approach?

Or it could make sense to split into two patches: 1) to add the `JavaStr` `into_raw` and `from_raw` APIs (which is a backwards compatible change that's easy to land now) and 2) the removal of the `JNIEnv::get_string_utf_chars` API which we can review and hold back from merging for a little while,

src/wrapper/strings/java_str.rs
TobiasDeBruijn: Good point. I've removed it.

TobiasDeBruijn:
Sure thing.
I'll split out the changes to `JavaStr` into a seperate PR, this PR will then only remove the methods from `JNIEnv`

TobiasDeBruijn:
Think I'll need to rebase this PR on #374 once that lands, otherwhise I break `JavaStr` in its old form

rib:
Okey I've just rebased this but for now I'll leave it unmerged since it's a breaking change. It's associated with milestone 0.21 so we'll hopefully remember to merge this before the next major release.

TobiasDeBruijn:
Sounds good! Thanks

PR #373, opened 2022-10-21
JNIEnv::get_superclass return none instead of a nullpointer

### LINKED ISSUE #151: Improve JNIEnv#get_superclass()
The current implementation of JNIEnv#get_superclass() uses jni_non_null_call!() for GetSuperclass which is a little redundant since GetSuperclass does not throw any exception thus jni_unchecked!() could be used. But it may return NULL for an object/interface that doesn't have a superclass. The suggested behavior for this case is to still treat this as an error but probably it worth introducing another entity like Err(NoSuperClass("<class/interface name> can't have a super class")) instead of the NullPtr.

TobiasDeBruijn:
## Overview
This PR changes the return type of `JNIEnv::get_superclass` to `Result<Option<JClass<'a>>>` from `Result<JClass<'a>>`.
`None` is returned when the class `class` has no superclass (i.e. is `java.lang.Object` an interface).

Documentation of the function was updated accordingly, and now also mentions `Errors`

Fixes #151

### Open questions
- Might it be better to check if `class` is `java.lang.Object` or an interface, rather than calling the method and checking if the pointer is null? (Probably need to benchmark that)

### Definition of Done

- [X] Change is covered by [automated tests](https://github.com/jni-rs/jni-rs/blob/master/CONTRIBUTING.md#tests)
- [X] User-visible changes are mentioned in the Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - `JNIEnv::get_string` checks that the given object is a `java.lang.String` instance to avoid undefined behaviour from the JNI implementation potentially aborting the program. ([#328](https://github.com/jni-rs/jni-rs/issues/328))
 - `JNIEnv::call_*method_unchecked` was marked `unsafe`, as passing improper argument types, or a bad number of arguments, can cause a JVM crash. ([#385](https://github.com/jni-rs/jni-rs/issues/385))
 - The `JNIEnv::new_object_unchecked` function now takes arguments as `&[jni::sys::jvalue]` to avoid allocating, putting it inline with changes to `JniEnv::call_*_unchecked` from 0.20.0 ([#382](https://github.com/jni-rs/jni-rs/pull/382))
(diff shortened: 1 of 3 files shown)


rib:
Cool, thanks, this looks good to me.

I may hold of from merging immediately just to keep a bit of breathing room after the recent 0.20 release before landing breaking API changes - just in case there's some cause to make a 0.20.1 point release with bug + documentation fixes it'll be slightly easier to handle.

rib:
> * Might it be better to check if `class` is `java.lang.Object` or an interface, rather than calling the method and checking if the pointer is null? (Probably need to benchmark that)

Seems fine to just call `GetSuperclass` and trust that JNI it will do this check for you. Explicitly checking for cases where there won't be a super class is effectively trying to do the job of `GetSuperclass` which could make sense if it were possible to be faster as an optimization but since we'd need to call into JNI to check if we have an `Object` anyway it's hard to imagine it being an optimization at that point.

TobiasDeBruijn:
Good point. Consider that question resolved then.

PR #374, opened 2022-10-22
Add `get_string_utf_chars` and `release_string_utf_chars` to `JavaStr` and update documentation

TobiasDeBruijn:
## Overview
See #372 

### Definition of Done

- [X] Change is covered by [automated tests](https://github.com/jni-rs/jni-rs/blob/master/CONTRIBUTING.md#tests)
- [X] The continuous integration build passes

### DIFF
--- a/src/wrapper/jnienv.rs
+++ b/src/wrapper/jnienv.rs
@@ -88,13 +88,18 @@ impl<'a> JNIEnv<'a> {
     ///
     /// # Safety
     ///
-    /// Expects a valid pointer retrieved from the `GetEnv` JNI function. Only does a null check.
+    /// Expects a valid pointer retrieved from the `GetEnv` JNI function or [Self::get_raw] function. Only does a null check.
     pub unsafe fn from_raw(ptr: *mut sys::JNIEnv) -> Result<Self> {
         non_null!(ptr, "from_raw ptr argument");
         Ok(JNIEnv {
             internal: ptr,
             lifetime: PhantomData,
         })
+    }
+
+    /// Get the raw JNIEnv pointer
+    pub fn get_raw(&self) -> *mut sys::JNIEnv {
+        self.internal
     }
 
     /// Get the java version that we're being executed from.
--- a/src/wrapper/strings/java_str.rs
+++ b/src/wrapper/strings/java_str.rs
@@ -1,3 +1,4 @@
+use jni_sys::{jboolean, JNI_TRUE};
 use std::{borrow::Cow, os::raw::c_char};
 
 use log::warn;
@@ -5,7 +6,7 @@ use log::warn;
 use crate::{errors::*, objects::JString, strings::JNIStr, JNIEnv};
 
 /// Reference to a string in the JVM. Holds a pointer to the array
-/// returned by GetStringUTFChars. Calls ReleaseStringUTFChars on Drop.
+/// returned by `GetStringUTFChars`. Calls `ReleaseStringUTFChars` on Drop.
 /// Can be converted to a `&JNIStr` with the same cost as the `&CStr.from_ptr`
 /// conversion.
 pub struct JavaStr<'a: 'b, 'b> {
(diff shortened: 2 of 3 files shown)


src/wrapper/strings/java_str.rs
```diff
@@ -27,11 +28,74 @@ impl<'a: 'b, 'b> JavaStr<'a, 'b> {
...
         self.internal
     }
+
+    /// Get the raw C string pointer from the [JavaStr]. This will be
+    /// encoded using the JVM internal `CESU-8`-style.
+    ///
+    /// # Warning
+    /// The programmer is responsible for making sure the backing string gets released
+    /// when they are done with it by resconstructing the [JavaStr] with [Self::from_raw],
+    /// which will release the backing string when it is dropped.
```
rib, line 63: I wonder if we could borrow more from the similar `Box::into_raw` docs here:

Docs for for `Box::into_raw` say:
> After calling this function, the caller is responsible for the memory previously managed by the Box. In particular, the caller should properly destroy T and release the memory, taking into account the [memory layout](https://doc.rust-lang.org/std/boxed/index.html#memory-layout) used by Box. The easiest way to do this is to convert the raw pointer back into a Box with the [Box::from_raw](https://doc.rust-lang.org/std/boxed/struct.Box.html#method.from_raw) function, allowing the Box destructor to perform the cleanup.

In particular it would be good to highlight that `from_raw` would be the recommended way to handle freeing the raw pointer later.

TobiasDeBruijn: Good point. Will fix

rib: oh, sorry was rushing here - the docs do actually highlight `from_raw` already

rib:
Cool, thanks for looking at this. Unit tests would be nice, but otherwise it just looks like it's missing a Changelog entry

src/wrapper/strings/java_str.rs
TobiasDeBruijn: Oops, accidentally unresolved it 😅

TobiasDeBruijn:
Unit tests shouldn't be too difficult. Will look into those.

src/wrapper/jnienv.rs
```diff
@@ -78,7 +78,7 @@ use crate::{
 #[derive(Clone, Copy)]
 #[repr(transparent)]
 pub struct JNIEnv<'a> {
-    internal: *mut sys::JNIEnv,
+    pub(crate) internal: *mut sys::JNIEnv,
```
rib, line 81: oh, how come this has been made public?

I would expect to leave this as private since it should be read-only - accessible via `get_raw()` if you want to do something `unsafe` with the raw pointer or `into_raw`

TobiasDeBruijn: Oh damn, missed the get_raw 😅 will change that back

TobiasDeBruijn: There seems to be no `get_raw` on JNIEnv. However, rather than making one, might it not be better to move the `get_string_utf_chars` back into JNIEnv, but change the visibility. This means keeping all JNI calls inside JNIEnv, then invoke them from JavaStr.

rib: Sorry I don't follow what the concern is here.

In this context isn't it enough that there's (already) a `get_raw()` for `JavaStr` in case an app has some `unsafe` thing it needs to do with the raw pointer that's stored in a `JavaStr`

There doesn't need to be any `get_raw` API under `JNIEnv` for strings does there?

TobiasDeBruijn: Currently (from what I've seen), all 'real' JNI calls are done in JNIEnv, this PR would move both string_utf_chars functions out of there into JavaStr (counting #372 as well). Would it not be better to leave them in `JNIEnv`, but make them non-pub, and 'expose' their functionality via JavaStr.

For `get_raw`, I couldn't find one on `JNIEnv` (to get at the inner pointer), hence one would need to be made if we want the 'real' JNI call to remain in `JavaStr`

rib: Ah, I see.

I guess either is reasonable though I personally like the encapsulation/locality from moving these functions into `JavaStr` if that's going to become the one and only thing that uses these.

Adding a `get_raw` for the env also sounds good , it's a bit surprised that's missing. Or else the env inner pointer could be made `pub(crate)` perhaps.

TobiasDeBruijn: I'll go with keeping it in JavaStr then, and add a get_raw to JNIEnv

rib: okey, cool, sounds good to me

tests/java_str.rs
```diff
@@ -0,0 +1,24 @@
...
+
+const TEST_STRING: &str = "";
+
+#[test]
+fn get_java_str() {
+    let env = attach_current_thread();
+    let string = unwrap(&env, env.new_string(TEST_STRING));
+    let javastr = unwrap(&env, JavaStr::from_env(&env, string.clone()));
+
+    let result = unsafe { JavaStr::from_raw(javastr.get_raw(), string, &env) };
```
rib, line 18: I think perhaps the docs for `from_raw` could make it clearer that for safety `from_raw` should only be giving a raw pointer that you own and maybe show an example of using `into_raw()` and then `from_raw().

The intention with the API is that there would be symmetry between taking ownership of the raw pointer via `into_raw()` and then returning ownership with `from_raw()`.

Although it's just a test that's self-aware of the imbalance it's possible someone might look at this code for guidance about how to use the API so it could be good if we test it in a "safe" way by calling `into_raw()` before calling `from_raw()` - so then there wouldn't be a need to use `ManuallyDrop` to avoid a double free.

rib:
I've just taken another pass over this with these updates:
1. Rebased on master
2. The implementation of `into_raw` uses `ManuallyDrop` to ensure the `Drop` destructor isn't called when consuming the `JavaStr` (since we are returning ownership of the string it's important that we shouldn't release it too) - When I tried updating the unit test to use `into_raw` + `from_raw` then I found this was leading to a crash due to a double free.
3. The unit test was updated to use `into_raw` paired with `from_raw`
4. The `get_string_utf_chars` utility exposes whether the implementation copied the string or not (just for completeness even though that information isn't used currently) The API was also made unsafe and the internal docs made consistent with the `get_string_unchecked` API.
5. Some arguments were re-ordered to have the more general, high-level objects first, such as `environment`, `string object`, `pointer`. This e.g. makes `from_raw` consistent with `from_env`.
6. Various doc additions/tweaks + tweaked the example to use `get_string()` instead of `JavaStr::from_env()`
7. The `get_string_utf_chars` + `release_string_utf_chars` utilities were moved to the top of the `JavaStr` impl - just a personal style preference since `get_string_utf_chars` is a pre-requisite for the implementation of `from_env` then I like seeing its implementation first.

I think this could be ok to land if you're happy with the updates @TobiasDeBruijn?

src/wrapper/strings/java_str.rs
```diff
@@ -15,10 +16,72 @@ pub struct JavaStr<'a: 'b, 'b> {
...
+        let mut is_copy: jboolean = 0;
+        let ptr: *const c_char = jni_non_null_call!(
+            env.get_raw(),
+            GetStringUTFChars,
+            obj.into_raw(),
+            &mut is_copy as *mut _
+        );
+
+        let is_copy = is_copy == JNI_TRUE;
+        println!("GetStringUTFChars, is_copy = {}", is_copy);
```
TobiasDeBruijn, line 56: This println probably should be removed

src/wrapper/strings/java_str.rs
```diff
@@ -27,11 +90,56 @@ impl<'a: 'b, 'b> JavaStr<'a, 'b> {
...
+    /// [Modified UTF-8](https://en.wikipedia.org/wiki/UTF-8#Modified_UTF-8) /
+    /// [CESU-8](https://en.wikipedia.org/wiki/CESU-8).
+    ///
+    /// # Warning
+    /// The programmer is responsible for making sure the backing string gets
+    /// released when they are done with it, for example by reconstructing a
+    /// [JavaStr] with [`Self::from_raw`], which will release the backing string
+    /// when it is dropped.
+    pub fn into_raw(self) -> *const c_char {
+        let _dont_call_drop = std::mem::ManuallyDrop::new(self);
```
TobiasDeBruijn, line 114: Would this not cause a memory leak?

At least, I think so. Since the `JavaStr` is never dropped anywhere now, I think?

Might make more sense to let the user handle the ManuallyDrop?

rib: I dont think this introduces a memory leak, rather it avoids a double free...

The JavaStr itself is still dropped it's just that Drop won't be run. Since only thing the Drop implementation frees is the string that we are trying to return to the caller, that's what we want here.

Previously we were returning a pointer to a string that would then get released by the Drop implementation before `into_raw` returned (I.e a pointer to memory that has been freed by the JNI implementation) and that would then lead to a double free.

I think the only thing that can be leaked here is the string we return in case they dont later call `from_raw` to ensure it gets properly dropped but that's made clear in the API docs.

src/wrapper/strings/java_str.rs
rib: oh, oops, yeah I was debugging whether I was getting a copy or or pinned string and forgot to remove that!

TobiasDeBruijn:
All good to merge I'd say

PR #382, opened 2022-11-12
allow no-alloc use of new_object_unchecked

chrismooredev:
## Overview

To align with the previous changes to allow `call_*_unchecked` methods without allocating, convert `new_object_unchecked` to do the same.

### Definition of Done


rib (approved):
Great; this all makes sense and looks good to me, thanks!

rib:
Thanks for the PR, the change looks good to land to me.

I was holding out from merging breaking API changes for a little while just for convenience in case we need to make any minor releases for 0.20 so for now I've added this PR to the 0.21 milestone.

Thanks again!

chrismooredev:
Sounds good, thanks!

chrismooredev:
It occurs to me that we should probably also mark all the call_*method_unchecked and new_object_unchecked methods as unsafe, as the caller could pass an array shorter than what the JVM expects to read...

rib:
> It occurs to me that we should probably also mark all the call_*method_unchecked and new_object_unchecked methods as unsafe, as the caller could pass an array shorter than what the JVM expects to read...
> 
> Let me know if you want me to push a commit for that to this PR and I can do it within a day or so

Yeah this sounds sensible too; though could probably warrant a separate PR since it's also somewhat orthogonal to this.

I'll at least create an issue for this from your comment and yeah if you were make a patch for that, that would be much appreciated thanks.

rib:
The updated changes here look good to me thanks @chrismooredev 

I think I'll probably start merging changes like this with API changes soon.

Issue #137, opened 2018-09-28
Consider type and method signature improvements


dmitry-timofeev:
## Overview
Currently one has to use the [JVM type descriptor format](https://docs.oracle.com/javase/specs/jvms/se11/html/jvms-4.html#jvms-4.3.2) in `call`/`get_field` methods. This format is designed to be used in Java class files and not in source code, hence, meets different requirements (it’s compact, fast and easy to parse, but not human-readable). That leaves us with such calls:
```rust
env.call_static_method(/* … */, "binarySearch", "([Ljava/lang/Object;IILjava/lang/Object;)Ljava/lang/Object;");
```
that are rather hard to read.

I’d consider various changes to support more user-friendly type and method signatures:
1. Types in Java source format.
1. Various ways to construct a method signature.
1. Update `call_x`/`get_x_field` to allow most convenient ways to pass type/method signatures.

### Types in the Java source format

Format | long | String[]
---|---|---
Java source | "long" | "java.lang.String[]"
JVM descriptor | "J" | "[Ljava/lang/String"
Enum (qualified) | `JavaType::Primitive(Primitive::Long)` | `JavaType::Array(JavaType::Object("java/lang/String"))`
Enum (imports) | `Primitive(Long)` | `Array(Object("java/lang/String"))`

### Constructing the method signature

Currently one can create a method signature (`TypeSignature`) either by parsing a JVM descriptor or as a structure. The first is used to construct the signature in `call_x` methods. If we provide an alternative way to specify the types, it might be necessary to provide new ways to construct the signature.

Some options:
```rust
let elementType = Object("java/lang/String");

// # Present APIs

// (0.1) JVM descriptors (present format)
// + Specification-compliant — no conversions needed
// - Horrible readability — no spaces, confusing type codes 
//   (do you know that "Z" stands for boolean and "J" for long?)
// - Highly error-prone
let signature = MethodSignature::from_str(
    "([Ljava/lang/String;Ljava/lang/String;Ljava/lang/Comparator;)Ljava/lang/String;");

// (0.2) As struct of enums (present format)
// + Explicit
// +/- Readable
// - A little noise, especially, if no imports are used/possible 
//   ("JavaType::Primitive(Primitive::Int)" or "Primitive(Int)" instead of
//   "I" in JVM or "int" in source-code version)
let signature = MethodSignature {
    args: !vec[Array(elementType), elementType, Object("java/lang/Comparator")],
    ret: elementType
}

// # API options to consider

[...]

PR #386, opened 2022-11-16
mark unchecked method calls as unsafe, fixes #385

### LINKED ISSUE #385: `call_*method_unchecked` and `new_object_unchecked` methods should be marked as `unsafe`
> It occurs to me that we should probably also mark all the `call_*method_unchecked` and `new_object_unchecked` methods as unsafe, as the caller could pass an array shorter than what the JVM expects to read...

_Originally posted by @chrismooredev in https://github.com/jni-rs/jni-rs/issues/382#issuecomment-1314763696_

chrismooredev:
## Overview

Marked the unchecked family of method calls as unsafe, per #385.

Passing in the incorrect number of arguments to a given method can cause the JVM to read past the argument buffer. Similarly, passing the wrong types of arguments can cause memory unsafety as well (ex: passing in a non-zero'd primitive when it expects an object pointer)

To reflect this, I also updated the safe wrappers to check the basic argument types that would case a JNI error - primitive types, and that objects are objects. If an object of the wrong type is provided, then the JVM should throw a regular exception.

Note too that the previous test, `tests::call_static_method_wrong_arg` was incorrect - it indeed threw an exception, but the exception was instead for a `java.lang.ArithmeticException: integer overflow` error when passing a `double` that the JVM interpreted as a `int`. This is unfortunately something the JVM itself can't really check for, so the safe wrappers were updated. The tests were then changed to ensure those checks worked.

I've kept the new_object_unchecked changes within my other PR in #382, so that+related tests aren't included here.

Please let me know if you spot any issues

### Definition of Done

### DIFF
--- a/benches/api_calls.rs
+++ b/benches/api_calls.rs
@@ -89,9 +89,8 @@ where
 {
     let x = JValue::from(x);
     let ret = ReturnType::Primitive(Primitive::Int);
-    let v = env
-        .call_static_method_unchecked(class, method_id, ret, &[x.into()])
-        .unwrap();
+    let v =
+        unsafe { env.call_static_method_unchecked(class, method_id, ret, &[x.into()]) }.unwrap();
     v.i().unwrap()
 }
 
@@ -100,7 +99,8 @@ where
     M: Desc<'m, JMethodID>,
 {
     let ret = ReturnType::Primitive(Primitive::Int);
-    let v = env.call_method_unchecked(obj, method_id, ret, &[]).unwrap();
+    // SAFETY: Caller retrieved method ID + class specifically for this use: Object.hashCode()I
+    let v = unsafe { env.call_method_unchecked(obj, method_id, ret, &[]) }.unwrap();
     v.i().unwrap()
 }
 
@@ -113,8 +113,8 @@ fn jni_object_call_static_unchecked<'c, C>(
 where
     C: Desc<'c, JClass<'c>>,
 {
-    let v = env
-        .call_static_method_unchecked(class, method_id, ReturnType::Object, args)
+    // SAFETY: Caller retrieved method ID and constructed arguments
+    let v = unsafe { env.call_static_method_unchecked(class, method_id, ReturnType::Object, args) }
         .unwrap();
     v.l().unwrap()
 }
--- a/src/wrapper/errors.rs
+++ b/src/wrapper/errors.rs
@@ -13,7 +13,7 @@ pub enum Error {
     WrongJValueType(&'static str, &'static str),
     #[error("Invalid constructor return type (must be void)")]
     InvalidCtorReturn,
-    #[error("Invalid number of arguments passed to java method: {0}")]
+    #[error("Invalid number or type of arguments passed to java method: {0}")]
     InvalidArgList(TypeSignature),
     #[error("Method not found: {name} {sig}")]
     MethodNotFound { name: String, sig: String },
--- a/src/wrapper/jnienv.rs
+++ b/src/wrapper/jnienv.rs
@@ -789,7 +789,12 @@ impl<'a> JNIEnv<'a> {
     ///
     /// Under the hood, this simply calls the `CallStatic<Type>MethodA` method
     /// with the provided arguments.
-    pub fn call_static_method_unchecked<'c, T, U>(
+    ///  
+    /// # Safety
+    ///
+    /// The provided JMethodID must be valid, and match the types and number of arguments, and return type.
+    /// If these are incorrect, the JVM may crash. The JMethodID must also match the passed type.
+    pub unsafe fn call_static_method_unchecked<'c, T, U>(
         &self,
         class: T,
         method_id: U,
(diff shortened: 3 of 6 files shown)


rib:
Thanks this change looks good to me. In particular thanks for the diligence with the `// SAFETY:` comments.

nirvdrum:
I just upgraded from 0.20 to 0.21-dev to get some of the array improvements and discovered that all of my `_unchecked` calls now require `unsafe`. I was porting some C++ code initially where you typically do the method lookup and store the method ID then pass that around. I converted most calls to just use the checked ones, but that results in repeated lookups when calling the same method multiple times. I'd like to avoid that overhead and I'd prefer not to litter my code with `unsafe`.

I understand you want to prevent making calls with invalid argument list lengths. Would it make sense to update `get_method_id` to return a compound object with the `JMethodID` and the `TypeSignature` and allow that to be passed around to `call_method_unchecked`? I don't mind the argument length check as that ought to be relatively cheap. At least it doesn't seem any more expensive than the existing "not null" and "no exception" checks. It'd be great if the happy path could avoid repeated semi-expensive work.

argv-minus-one:
You also get undefined behavior by calling a method by `JMethodID` on an object of the wrong class. To make that safe, the `JMethodID` would have to be packaged together with a `GlobalRef` pointing to a `java.lang.Class` and do an `is_instance_of` check upon use.

It may or may not be possible (I'm not sure) to get undefined behavior by calling a method with arguments of the wrong type (e.g. passing an object of type `BigDecimal` as an argument that's supposed to have type `String`), but [`jni` doesn't currently check for that](https://github.com/jni-rs/jni-rs/blob/a0c4c48f7b4304a4b109d3617213159c142d64c1/src/wrapper/jnienv.rs#L1384) anyway.

rib:
heya @nirvdrum, I do agree that the the ergonomics + safety trade offs for the `call_method_` APIs probably need to be revisited.

Personally I also don't like the current trade off and I currently choose to use the `_unchecked` call APIs in my projects because I find the redundant work done at runtime by the safe APIs to be excessive just to make a single function call.

That said though I also think it's appropriate that the jni crate should be honest about which APIs really are `unsafe`. The change here is just about being honest so that it's clear to users of the API when there is a high risk that code using this API can quite easily lead to undefined behaviour.

The `jni` crate has actually had quite a large number of public APIs that have frankly been extremely `unsafe` and that has given a very false sense that the `jni` crate is providing a safe JNI wrapper over the lower-level `jni-sys` crate.

Given that Rust crates usually strive to provide safe abstractions, I think it makes sense that the `jni` crate isn't purely about trying to provide an ergonomic abstraction over the raw `jni-sys` API but that it also tries to help you use the JNI API safely.

Although it's difficult (perhaps even impossible) to make a perfectly safe JNI abstraction (so there are pragmatic trade offs that needs to be made), I think for safe APIs I'd like to feel confident that you can't easily hit undefined behaviour due to simple programming mistakes - which was the case with these APIs.

Working towards the 0.21 release a number of changes have landed that have tried to make the API safer or at least more honest about what's safe. It does unfortunately mean this release is going to cause a bit more churn for users compared to the last release but it does seem like it's the appropriate direction to iterating the API.

If you were happy with how these APIs weren't marked as `unsafe` before then maybe Instead of switching your own projects over to using the safe APIs you can consider making your own "safe" wrapper for now that lets you avoid scattering your code with `unsafe {}` blocks. This is basically what I do. This way the `jni` crate (which is a relatively low-level API that doesn't have any context for your use case) can just be honest about which APIs are safe, and it's then your responsibility to decide (for your use case / needs) if you're happy to ignore certain risks of undefined behaviour if they are acceptable for your use case.

I think the real problem here isn't so much that the jni crate marks these APIs as `unsafe` its that there is currently no safe alternative that has a better performance trade off. Personally I'd be happy to trade off ergonomics in some other way that means we can make API calls more efficiently but still be safe.

One suggestion that has come up before has been to adopt a macro like Signal has here: https://github.com/jni-rs/jni-rs/issues/137#issuecomment-1292521322

It could also be nice if there was a safer API that didn't necessarily require a macro but was more efficient but certainly there could be lots of opportunities to get improved safety, performance and ergonomics via some macros here.

In general I think issue #137 is probably the best place to continue the discussion around how we can improve the ergonomics for making API calls, but would also be happy to see other ideas / issues / PRs that poke at this.

argv-minus-one:
The Holy Grail solution to this problem would be a binding generator that looks at JVM class files and generates safe Rust wrappers around them. That way, you get hard safety guarantees without the performance penalty we have now.

Unfortunately, that would be a ton of work. @MaulingMonkey once wrote one called [jni-bindgen](https://github.com/MaulingMonkey/jni-bindgen), but doesn't seem to have been able to get it past an “only barely kinda partially usable” state.

I notice that the jni-bindgen project includes jni-glue, another Rust wrapper around the JNI API. I wonder if there's anything we can do to make this library better suited for use by a binding generator, so that the next time someone tries writing one, it won't be quite as hard.

rib:
Yeah I was skimming the work that @MaulingMonkey did recently with jni-bindgen and it would be good to see more experimentation along those lines.

I can imagine there may be a number of different high level codegen approaches that may have different trade offs and it might be best that some of them would build on top of the `jni` crate instead of necessarily being built-in, but there is likely also a middle ground where we could have some macros that could help with more hand written bindings which can deal with signature parsing and argument type validation at compile time.

Issue #189, opened 2019-06-20
Test all documentation examples


dmitry-timofeev:
Consider enabling doc-tests for all examples present in JNIEnv.

Currently, the only tested example is in JavaVM docs. It demonstrates that quite some setup is required to launch a JVM and obtain JNIEnv; also, such tests must be disabled when no invocation feature is present. That can likely be extracted in a macro, enabling consice, maintainable and correct doc tests.

See:

- The relevant portion of JavaVM test, that is needed to launch a VM and obtain a JNIEnv:
https://github.com/jni-rs/jni-rs/blob/94baf211f4accf0532147e3b3971b4c1eb8e1f66/src/wrapper/java_vm/vm.rs#L57-L83
- #188

PR #388, opened 2022-11-20
fixing documentation example into_inner() => into_raw()

homelessnessbo:
## Overview

Just a simple documentation change, I see that the repo example has already migrated, but copying examples from documentation yields a compile-time error.


rib:
Thanks - it's a good reminder to address https://github.com/jni-rs/jni-rs/issues/189

Issue #396, opened 2022-12-09
All APIs like `env.new_*_array()` should return a `JObject` wrapper (like `JByteArray`) allowing safe JValue conversion


EliseChouleur:
Particulary to be able to call a methode with a byte array in parameter.
It was possible in JNI 0.19 thanks to `into` but the new conversion methode doesn't work with jbyteArray.

e.g:
jni-rs 0.19

```
    let buffer = env.new_byte_array(asset_size)?; 
    env.call_method(asset_stream, "read", "([B)I", &[buffer.into()])?;
```

rib:
Ah, right, I guess this was affected when addressing the soundness issues with allowing safe conversions from sys types: https://github.com/jni-rs/jni-rs/issues/197

The odd thing here is that `jni-rs` is directly returning a sys type from `env.new_byte_array` instead of a type like `JObject` that has an associated lifetime for the local reference.

We should also track this in the context of https://github.com/jni-rs/jni-rs/issues/392 where we're looking to ensure that local reference types are non-Copy.

Since the array is effectively just an object you can pass your array to method calls like this for now:

```rust
    let buffer = env.new_byte_array(asset_size)?;
    let buffer_obj = unsafe { JObject::from_raw(buffer) };
    env.call_method(stream, "read", "([B)I", &[buffer_obj.into()])?;
```

We probably shouldn't directly allow safe conversion between `jbyteArray` and `JValue` - similar to not allowing safe conversions of `jobject` to `JObject` otherwise safe code can can easily pass literal numbers as an invalid pointers.

What we probably need to do here is introduce a `JByteArray<'env>` type similar to `JString` or `JClass` that would be returned by `env.new_byte_array()` and then there should be a safe way to convert a `JByteArray` into a `JValue` (which should make your original code work again)

EliseChouleur:
Hello
Thanks for the temporary fix !

Issue #399, opened 2023-01-10
`JNIEnv::with_local_frame` should be able to return any value, not just `jni::results::Result<JObject>`


argv-minus-one:
This is to continue the conversation begun in https://github.com/jni-rs/jni-rs/pull/398#issuecomment-1372430485.

In brief, we want to change `JNIEnv::with_local_frame` to be able to return any result type, not just `jni::results::Result<JObject>`.

---

@rib, I'm mostly okay with [your proposed change](https://github.com/jni-rs/jni-rs/commit/22d126b9139300a57c7216171b7444ff0b53c23d), but I have one issue with it. Because `with_local_frame` returns `jni::results::Result<T>`, and because the closure will always perform JNI calls and therefore potentially fail with `jni::Results::Error`, `with_local_frame` will always return nested `Result`s. [This has already given you some trouble.](https://github.com/jni-rs/jni-rs/blob/22d126b9139300a57c7216171b7444ff0b53c23d/tests/java_integers.rs#L39)

There is a method `Result::flatten` to unnest `Result`s, but:

1. It's unstable.
2. It doesn't work if the error types are different, even if one error type can be converted to the other ([although that has been proposed](https://github.com/rust-lang/rust/issues/70142#issuecomment-699710803)).

The nested `Result`s can easily be unnested with the `?` operator in some situations:

```rust
let () = env.with_local_frame(1, |env| {
    some_fallible_function()?;
    Ok(())
})??;
```

But that's only true if `with_local_frame` is called in a function that returns `Result` and the error type implements both `From<jni::errors::Error>` and `From<E>` (where `E` is the error type returned by the closure). Otherwise, the `Result`s have to be unnested explicitly:

```rust
let result: Result<(), E> = env.with_local_frame(1, |env| {
    some_fallible_function()?;
    Ok(())
}).map_err(Into::into).and_then(std::convert::identity);
```

---

I'm not sure if this is actually better or not, but [here's another idea](https://github.com/jni-rs/jni-rs/commit/08a129d95cbfad01aafbf91b363b3177aa26e894): have the `with_local_frame` closure return a `Result`, but with any error type that implements `From<jni::errors::Error>`. That way, a different error type can be used without nesting `Result`s. If there is a problem pushing or popping the local frame, the error is converted into the closure's error type.

[...]

Issue #397, opened 2022-12-20
Report error when run HelloWord on jdk11?


lfygh:
$ java --version
openjdk 11.0.15 2022-04-19 LTS
OpenJDK Runtime Environment Zulu11.56+19-CA (build 11.0.15+10-LTS)
OpenJDK 64-Bit Server VM Zulu11.56+19-CA (build 11.0.15+10-LTS, mixed mode)

$ uname -a
Darwin MacBook-Pro.local 20.5.0 Darwin Kernel Version 20.5.0: Sat May  8 05:10:31 PDT 2021; root:xnu-7195.121.3~9/RELEASE_ARM64_T8101 arm64

```
javac HelloWorld.java && java -Djava.library.path=mylib/target/debug/ HelloWorld
Exception in thread "main" java.lang.UnsatisfiedLinkError: no mylib in java.library.path: [mylib/target/debug/]
        at java.base/java.lang.ClassLoader.loadLibrary(ClassLoader.java:2673)
        at java.base/java.lang.Runtime.loadLibrary0(Runtime.java:830)
        at java.base/java.lang.System.loadLibrary(System.java:1873)
        at HelloWorld.<clinit>(HelloWorld.java:13)

```

rib:
Since I was also in the middle of another PR which I realize also affects this example I've just made a patch to update `example/mylib` as part of this PR: https://github.com/jni-rs/jni-rs/pull/400

You can also fetch this branch: https://github.com/rib/jni-rs/tree/array-wrapper-types

It'd be good to know if you can run the example via `cd example && make java_run` with the above branch.

rib:
Since merging #400 then the `example/mylib` crate should compile now, so it should hopefully be possible to run `cd example && make java_run` with the current `master` branch.

I'm going to close this now that I think that the `mylib` crate builds since I think the original issue was that the `libmylib.so` hadn't been compiled prior to running `javac HelloWorld.java && java -Djava.library.path=mylib/target/debug/ HelloWorld`

PR #400, opened 2023-01-18
Add Array wrapper types with lifetimes

### LINKED ISSUE #41: Byte arrays in signature
Hello

I'm following your example. It has a signature like this:

```rust
pub extern "system" fn Java_HelloWorld_hello(env: JNIEnv,
                                             class: JClass,
                                             input: JString)
                                             -> jstring {
...
```

From looking at the signature, it seems that parameter types come from the `objects` module while the return value is a `sys` type.

How would I accept a byte array as input value instead of a `JString`? There's `sys::jbyteArray` but there doesn't seem to be a matching `objects` type?

rib:
## Overview

Instead of all the array-based APIs being based on sys types like
`jbyteArray` (which can easily lead to the use of invalid local
references) this introduces types like `JByteArray<'local>` that
have a lifetime parameter to ensure the local reference can't
be read if it is deleted or goes out of scope.

`JByteArray` and `JIntArray` etc are based on a generic
`JPrimitiveArray<T>` which takes any of the primitive types that
may be put into arrays, like `jbyte` or `jint`.

`JObjectArray` is handled separately, considering that it can't be used
with `AutoArray` or `AutoPrimitiveArray`.

The implementations of `AutoArray` and `AutoPrimitiveArray` were updated
so they can be created based on a `JPrimitiveAarray`.

While updating the lifetime parameters to the `AutoArray` and
`AutoPrimitiveArray` APIs numerous implementation inconsistencies were
found, even though they should only really differ in how they acquire /
release the pointer to array elements. This patch also normalizes the
implementation of both of these APIs, so they are easily comparable. It
may even be possible to unify these at some point.

Closes: #396
Closes #41

### Definition of Done

### DIFF
--- a/src/wrapper/jnienv.rs
+++ b/src/wrapper/jnienv.rs
@@ -2805,17 +2805,63 @@ impl<'local> JNIEnv<'local> {
         jni_error_code_to_result(res)
     }
 
-    /// Return an [`AutoArray`] of the given Java array.
+    /// Returns an [`AutoArray`] to access the elements of the given Java `array`.
     ///
-    /// The result is valid until the `AutoArray` object goes out of scope, when the
-    /// release happens automatically according to the `mode` parameter.
+    /// The elements are accessible until the returned auto-release guard is dropped.
+    ///
+    /// The returned array may be a copy of the Java array and changes made to
+    /// the returned array will not necessarily be reflected in the original
+    /// array until the [`AutoArray`] guard is dropped.
+    ///
+    /// If you know in advance that you will only be reading from the array then
+    /// pass [`ReleaseMode::NoCopyBack`] so that the JNI implementation knows
+    /// that it's not necessary to copy any data back to the original Java array
+    /// when the [`AutoArray`] guard is dropped.
     ///
     /// Since the returned array may be a copy of the Java array, changes made to the
     /// returned array will not necessarily be reflected in the original array until
     /// the corresponding `Release*ArrayElements` JNI method is called.
     /// [`AutoArray`] has a commit() method, to force a copy back of pending
     /// array changes if needed (and without releasing it).
-    pub fn get_array_elements<'other_local, 'array, T: TypeArray>(
+    ///
+    /// # Safety
+    ///
+    /// ## No data races
+    ///
+    /// This API has no built-in synchronization that ensures there won't be any data
+    /// races while accessing the array elements.
+    ///
+    /// To avoid undefined behaviour it is the caller's responsibility to ensure there
+    /// will be no data races between other Rust or Java threads trying to access the
+    /// same array.
+    ///
+    /// Acquiring a [`MonitorGuard`] lock for the `array` could be one way of ensuring
+    /// mutual exclusion between Rust and Java threads, so long as the Java threads
+    /// also acquire the same lock via `synchronized(array) {}`.
+    ///
+    /// ## No aliasing
+    ///
+    /// Callers must not create more than one [`AutoArray`] or
+    /// [`AutoPrimitiveArray`] per Java array at the same time - even if
+    /// there is no risk of a data race.
+    ///
+    /// The reason for this restriction is that [`AutoArray`] and
+    /// [`AutoPrimitiveArray`] implement `DerefMut` which can provide a
+    /// mutable `&mut [T]` slice reference for the elements and it would
+    /// constitute undefined behaviour to allow there to be more than one
+    /// mutable reference that points to the same memory.
+    ///
+    /// # jboolean elements
+    ///
+    /// Keep in mind that arrays of `jboolean` values should only ever hold
+    /// values of `0` or `1` because any other value could lead to undefined
+    /// behaviour within the JVM.
+    ///
+    /// Also see
+    /// [`get_primitive_array_critical`](Self::get_primitive_array_critical) which
+    /// imposes additional restrictions that make it less likely to incur the
+    /// cost of copying the array elements.
+    pub unsafe fn get_array_elements<'other_local, 'array, T: TypeArray>(
         &mut self,
         array: &'array JPrimitiveArray<'other_local, T>,
         mode: ReleaseMode,
@@ -2824,27 +2870,98 @@ impl<'local> JNIEnv<'local> {
         AutoArray::new(self, array, mode)
     }
 
-    /// Return an [`AutoPrimitiveArray`] of the given Java primitive array.
+    /// Returns an [`AutoPrimitiveArray`] to access the elements of the given Java `array`.
     ///
-    /// The result is valid until the corresponding AutoPrimitiveArray object goes out of scope,
-    /// when the release happens automatically according to the mode parameter.
+    /// The elements are accessible during the critical section that exists until the
+    /// returned auto-release guard is dropped.
     ///
-    /// Given that Critical sections must be as short as possible, and that they come with a
-    /// number of important restrictions (see GetPrimitiveArrayCritical JNI doc), use this
-    /// wrapper wisely, to avoid holding the array longer that strictly necessary.
-    /// In any case, you can:
-    ///  - Use std::mem::drop explicitly, to force / anticipate resource release.
-    ///  - Use a nested scope, to release the array at the nested scope's exit.
+    /// This API imposes some strict restrictions that help the JNI implementation
+    /// avoid any need to copy the underlying array elements before making them
+    /// accessible to native code:
     ///
-    /// Since the returned array may be a copy of the Java array, changes made to the
-    /// returned array will not necessarily be reflected in the original array until
-    /// ReleasePrimitiveArrayCritical is called; which happens at AutoPrimitiveArray
-    /// destruction.
+    /// 1. No other use of JNI calls are allowed (on the same thread) within the critical
+    /// section that exists while holding the [`AutoPrimitiveArray`] guard.
+    /// 2. No system calls can be made (Such as `read`) that may depend on a result
+    /// from another Java thread.
+    ///
+    /// The JNI spec does not specify what will happen if these rules aren't adhered to
+    /// but it should be assumed it will lead to undefined behaviour, likely deadlock
+    /// and possible program termination.
+    ///
+    /// Even with these restrictions the returned array may still be a copy of
+    /// the Java array and changes made to the returned array will not
+    /// necessarily be reflected in the original array until the [`AutoPrimitiveArray`]
+    /// guard is dropped.
+    ///
+    /// If you know in advance that you will only be reading from the array then
+    /// pass [`ReleaseMode::NoCopyBack`] so that the JNI implementation knows
+    /// that it's not necessary to copy any data back to the original Java array
+    /// when the [`AutoPrimitiveArray`] guard is dropped.
+    ///
+    /// A nested scope or explicit use of `std::mem::drop` can be used to
+    /// control when the returned [`AutoPrimitiveArray`] is dropped to
+    /// minimize the length of the critical section.
     ///
     /// If the given array is `null`, an `Error::NullPtr` is returned.
     ///
-    /// See also [`get_array_elements`](Self::get_array_elements)
-    pub fn get_primitive_array_critical<'other_local, 'array, 'env, T: TypeArray>(
+    /// # Safety
+    ///
+    /// ## Critical Section Restrictions
+    ///
+    /// Although this API takes a mutable reference to a [`JNIEnv`] which should
+    /// ensure that it's not possible to call JNI, this API is still marked as
+    /// `unsafe` due to the complex, far-reaching nature of the critical-section
+    /// restrictions imposed here that can't be guaranteed simply through Rust's
+    /// borrow checker rules.
+    ///
+    /// The rules above about JNI usage and system calls _must_ be adhered to.
+    ///
+    /// Using this API implies:
+    ///
+    /// 1. All garbage collection will likely be paused during the critical section
+    /// 2. Any use of JNI in other threads may block if they need to allocate memory
+    ///    (due to the garbage collector being paused)
+    /// 3. Any use of system calls that will wait for a result from another Java thread
+    ///    could deadlock if that other thread is blocked by a paused garbage collector.
+    ///
+    /// A failure to adhere to the critical section rules could lead to any
+    /// undefined behaviour, including aborting the program.
+    ///
+    /// ## No data races
+    ///
+    /// This API has no built-in synchronization that ensures there won't be any data
+    /// races while accessing the array elements.
+    ///
+    /// To avoid undefined behaviour it is the caller's responsibility to ensure there
+    /// will be no data races between other Rust or Java threads trying to access the
+    /// same array.
+    ///
+    /// Acquiring a [`MonitorGuard`] lock for the `array` could be one way of ensuring
+    /// mutual exclusion between Rust and Java threads, so long as the Java threads
+    /// also acquire the same lock via `synchronized(array) {}`.
+    ///
+    /// ## No aliasing
+    ///
+    /// Callers must not create more than one [`AutoArray`] or
+    /// [`AutoPrimitiveArray`] per Java array at the same time - even if
+    /// there is no risk of a data race.
+    ///
+    /// The reason for this restriction is that [`AutoArray`] and
+    /// [`AutoPrimitiveArray`] implement `DerefMut` which can provide a
+    /// mutable `&mut [T]` slice reference for the elements and it would
+    /// constitute undefined behaviour to allow there to be more than one
+    /// mutable reference that points to the same memory.
+    ///
+    /// ## jboolean elements
+    ///
+    /// Keep in mind that arrays of `jboolean` values should only ever hold
+    /// values of `0` or `1` because any other value could lead to undefined
+    /// behaviour within the JVM.
+    ///
+    /// Also see [`get_array_elements`](Self::get_array_elements) which has fewer
+    /// restrictions, but is is more likely to incur a cost from copying the
+    /// array elements.
+    pub unsafe fn get_primitive_array_critical<'other_local, 'array, 'env, T: TypeArray>(
         &'env mut self,
         array: &'array JPrimitiveArray<'other_local, T>,
         mode: ReleaseMode,
--- a/src/wrapper/jnienv.rs
+++ b/src/wrapper/jnienv.rs
@@ -12,9 +12,9 @@ use crate::{
     descriptors::Desc,
     errors::*,
     objects::{
-        AutoArray, AutoLocal, AutoPrimitiveArray, GlobalRef, JByteBuffer, JClass, JFieldID, JList,
-        JMap, JMethodID, JObject, JStaticFieldID, JStaticMethodID, JString, JThrowable, JValue,
-        JValueOwned, ReleaseMode, TypeArray, WeakRef,
+        AutoElements, AutoElementsCritical, AutoLocal, GlobalRef, JByteBuffer, JClass, JFieldID,
+        JList, JMap, JMethodID, JObject, JStaticFieldID, JStaticMethodID, JString, JThrowable,
+        JValue, JValueOwned, ReleaseMode, TypeArray, WeakRef,
     },
     signature::{JavaType, Primitive, TypeSignature},
     strings::{JNIString, JavaStr},
@@ -287,7 +287,7 @@ impl<'local> JNIEnv<'local> {
         &mut self,
         name: S,
         loader: &JObject,
-        buf: &AutoArray<'_, '_, '_, jbyte>,
+        buf: &AutoElements<'_, '_, '_, jbyte>,
     ) -> Result<JClass<'local>>
     where
         S: Into<JNIString>,
@@ -2805,23 +2805,23 @@ impl<'local> JNIEnv<'local> {
         jni_error_code_to_result(res)
     }
 
-    /// Returns an [`AutoArray`] to access the elements of the given Java `array`.
+    /// Returns an [`AutoElements`] to access the elements of the given Java `array`.
     ///
     /// The elements are accessible until the returned auto-release guard is dropped.
     ///
     /// The returned array may be a copy of the Java array and changes made to
     /// the returned array will not necessarily be reflected in the original
-    /// array until the [`AutoArray`] guard is dropped.
+    /// array until the [`AutoElements`] guard is dropped.
     ///
     /// If you know in advance that you will only be reading from the array then
     /// pass [`ReleaseMode::NoCopyBack`] so that the JNI implementation knows
     /// that it's not necessary to copy any data back to the original Java array
-    /// when the [`AutoArray`] guard is dropped.
+    /// when the [`AutoElements`] guard is dropped.
     ///
     /// Since the returned array may be a copy of the Java array, changes made to the
     /// returned array will not necessarily be reflected in the original array until
     /// the corresponding `Release*ArrayElements` JNI method is called.
-    /// [`AutoArray`] has a commit() method, to force a copy back of pending
+    /// [`AutoElements`] has a commit() method, to force a copy back of pending
     /// array changes if needed (and without releasing it).
     ///
     /// # Safety
@@ -2841,12 +2841,12 @@ impl<'local> JNIEnv<'local> {
     ///
     /// ## No aliasing
     ///
-    /// Callers must not create more than one [`AutoArray`] or
-    /// [`AutoPrimitiveArray`] per Java array at the same time - even if
+    /// Callers must not create more than one [`AutoElements`] or
+    /// [`AutoElementsCritical`] per Java array at the same time - even if
     /// there is no risk of a data race.
     ///
-    /// The reason for this restriction is that [`AutoArray`] and
-    /// [`AutoPrimitiveArray`] implement `DerefMut` which can provide a
+    /// The reason for this restriction is that [`AutoElements`] and
+    /// [`AutoElementsCritical`] implement `DerefMut` which can provide a
     /// mutable `&mut [T]` slice reference for the elements and it would
     /// constitute undefined behaviour to allow there to be more than one
     /// mutable reference that points to the same memory.
@@ -2858,19 +2858,19 @@ impl<'local> JNIEnv<'local> {
     /// behaviour within the JVM.
     ///
     /// Also see
-    /// [`get_primitive_array_critical`](Self::get_primitive_array_critical) which
+    /// [`get_array_elements_critical`](Self::get_array_elements_critical) which
     /// imposes additional restrictions that make it less likely to incur the
     /// cost of copying the array elements.
     pub unsafe fn get_array_elements<'other_local, 'array, T: TypeArray>(
         &mut self,
         array: &'array JPrimitiveArray<'other_local, T>,
         mode: ReleaseMode,
-    ) -> Result<AutoArray<'local, 'other_local, 'array, T>> {
+    ) -> Result<AutoElements<'local, 'other_local, 'array, T>> {
         non_null!(array, "get_array_elements array argument");
-        AutoArray::new(self, array, mode)
+        AutoElements::new(self, array, mode)
     }
 
-    /// Returns an [`AutoPrimitiveArray`] to access the elements of the given Java `array`.
+    /// Returns an [`AutoElementsCritical`] to access the elements of the given Java `array`.
     ///
     /// The elements are accessible during the critical section that exists until the
     /// returned auto-release guard is dropped.
@@ -2880,7 +2880,7 @@ impl<'local> JNIEnv<'local> {
     /// accessible to native code:
     ///
     /// 1. No other use of JNI calls are allowed (on the same thread) within the critical
-    /// section that exists while holding the [`AutoPrimitiveArray`] guard.
+    /// section that exists while holding the [`AutoElementsCritical`] guard.
     /// 2. No system calls can be made (Such as `read`) that may depend on a result
     /// from another Java thread.
     ///
@@ -2890,16 +2890,16 @@ impl<'local> JNIEnv<'local> {
     ///
     /// Even with these restrictions the returned array may still be a copy of
     /// the Java array and changes made to the returned array will not
-    /// necessarily be reflected in the original array until the [`AutoPrimitiveArray`]
+    /// necessarily be reflected in the original array until the [`AutoElementsCritical`]
     /// guard is dropped.
     ///
     /// If you know in advance that you will only be reading from the array then
     /// pass [`ReleaseMode::NoCopyBack`] so that the JNI implementation knows
     /// that it's not necessary to copy any data back to the original Java array
-    /// when the [`AutoPrimitiveArray`] guard is dropped.
+    /// when the [`AutoElementsCritical`] guard is dropped.
     ///
     /// A nested scope or explicit use of `std::mem::drop` can be used to
-    /// control when the returned [`AutoPrimitiveArray`] is dropped to
+    /// control when the returned [`AutoElementsCritical`] is dropped to
     /// minimize the length of the critical section.
     ///
     /// If the given array is `null`, an `Error::NullPtr` is returned.
@@ -2942,12 +2942,12 @@ impl<'local> JNIEnv<'local> {
     ///
     /// ## No aliasing
     ///
-    /// Callers must not create more than one [`AutoArray`] or
-    /// [`AutoPrimitiveArray`] per Java array at the same time - even if
+    /// Callers must not create more than one [`AutoElements`] or
+    /// [`AutoElementsCritical`] per Java array at the same time - even if
     /// there is no risk of a data race.
     ///
-    /// The reason for this restriction is that [`AutoArray`] and
-    /// [`AutoPrimitiveArray`] implement `DerefMut` which can provide a
+    /// The reason for this restriction is that [`AutoElements`] and
+    /// [`AutoElementsCritical`] implement `DerefMut` which can provide a
     /// mutable `&mut [T]` slice reference for the elements and it would
     /// constitute undefined behaviour to allow there to be more than one
     /// mutable reference that points to the same memory.
@@ -2961,13 +2961,13 @@ impl<'local> JNIEnv<'local> {
     /// Also see [`get_array_elements`](Self::get_array_elements) which has fewer
     /// restrictions, but is is more likely to incur a cost from copying the
     /// array elements.
-    pub unsafe fn get_primitive_array_critical<'other_local, 'array, 'env, T: TypeArray>(
+    pub unsafe fn get_array_elements_critical<'other_local, 'array, 'env, T: TypeArray>(
         &'env mut self,
         array: &'array JPrimitiveArray<'other_local, T>,
         mode: ReleaseMode,
-    ) -> Result<AutoPrimitiveArray<'local, 'other_local, 'array, 'env, T>> {
+    ) -> Result<AutoElementsCritical<'local, 'other_local, 'array, 'env, T>> {
         non_null!(array, "get_primitive_array_critical array argument");
-        AutoPrimitiveArray::new(self, array, mode)
+        AutoElementsCritical::new(self, array, mode)
     }
 }
 
--- a/src/wrapper/jnienv.rs
+++ b/src/wrapper/jnienv.rs
@@ -8,7 +8,6 @@ use std::{
 
 use log::warn;
 
-use crate::signature::ReturnType;
 use crate::{
     descriptors::Desc,
     errors::*,
@@ -32,6 +31,7 @@ use crate::{
         JObjectArray, JPrimitiveArray, JShortArray,
     },
 };
+use crate::{objects::AsJArrayRaw, signature::ReturnType};
 
 /// FFI-compatible JNIEnv struct. You can safely use this as the JNIEnv argument
 /// to exported methods that will be called by java. This is where most of the
@@ -1648,10 +1648,13 @@ impl<'local> JNIEnv<'local> {
         Ok(unsafe { JString::from_raw(s) })
     }
 
-    /// Get the length of a java array
-    pub fn get_array_length(&self, array: jarray) -> Result<jsize> {
-        non_null!(array, "get_array_length array argument");
-        let len: jsize = jni_unchecked!(self.internal, GetArrayLength, array);
+    /// Get the length of a [`JPrimitiveArray`] or [`JObjectArray`].
+    pub fn get_array_length<'other_local, 'array>(
+        &self,
+        array: &'array impl AsJArrayRaw<'other_local>,
+    ) -> Result<jsize> {
+        non_null!(array.as_jarray_raw(), "get_array_length array argument");
+        let len: jsize = jni_unchecked!(self.internal, GetArrayLength, array.as_jarray_raw());
         Ok(len)
     }
 
--- a/src/wrapper/jnienv.rs
+++ b/src/wrapper/jnienv.rs
@@ -8,7 +8,6 @@ use std::{
 
 use log::warn;
 
-use crate::errors::Error::JniCall;
 use crate::signature::ReturnType;
 use crate::{
     descriptors::Desc,
@@ -21,11 +20,17 @@ use crate::{
     signature::{JavaType, Primitive, TypeSignature},
     strings::{JNIString, JavaStr},
     sys::{
-        self, jarray, jboolean, jbooleanArray, jbyte, jbyteArray, jchar, jcharArray, jdouble,
-        jdoubleArray, jfloat, jfloatArray, jint, jintArray, jlong, jlongArray, jobjectArray,
-        jshort, jshortArray, jsize, jvalue, JNINativeMethod,
+        self, jarray, jboolean, jbyte, jchar, jdouble, jfloat, jint, jlong, jshort, jsize, jvalue,
+        JNINativeMethod,
     },
     JNIVersion, JavaVM,
+};
+use crate::{
+    errors::Error::JniCall,
+    objects::{
+        JBooleanArray, JByteArray, JCharArray, JDoubleArray, JFloatArray, JIntArray, JLongArray,
+        JObjectArray, JPrimitiveArray, JShortArray,
+    },
 };
 
 /// FFI-compatible JNIEnv struct. You can safely use this as the JNIEnv argument
@@ -282,7 +287,7 @@ impl<'local> JNIEnv<'local> {
         &mut self,
         name: S,
         loader: &JObject,
-        buf: AutoArray<'_, jbyte>,
+        buf: &AutoArray<'_, '_, '_, jbyte>,
     ) -> Result<JClass<'local>>
     where
         S: Into<JNIString>,
@@ -1663,59 +1668,58 @@ impl<'local> JNIEnv<'local> {
         length: jsize,
         element_class: T,
         initial_element: U,
-    ) -> Result<jobjectArray>
+    ) -> Result<JObjectArray<'local>>
     where
         T: Desc<'local, JClass<'other_local_2>>,
         U: AsRef<JObject<'other_local_1>>,
     {
         let class = element_class.lookup(self)?;
 
-        let ret = Ok(jni_non_null_call!(
+        let array: jarray = jni_non_null_call!(
             self.internal,
             NewObjectArray,
             length,
             class.as_ref().as_raw(),
             initial_element.as_ref().as_raw()
-        ));
+        );
+
+        let array = unsafe { JObjectArray::from_raw(array) };
 
         // Ensure that `class` isn't dropped before the JNI call returns.
         drop(class);
 
-        ret
+        Ok(array)
     }
 
-    /// Returns an element of the `jobjectArray` array.
-    pub fn get_object_array_element(
-        &self,
-        array: jobjectArray,
+    /// Returns a local reference to an element of the [`JObjectArray`] `array`.
+    pub fn get_object_array_element<'other_local>(
+        &mut self,
+        array: impl AsRef<JObjectArray<'other_local>>,
         index: jsize,
     ) -> Result<JObject<'local>> {
-        non_null!(array, "get_object_array_element array argument");
+        non_null!(array.as_ref(), "get_object_array_element array argument");
         Ok(unsafe {
             JObject::from_raw(jni_non_void_call!(
                 self.internal,
                 GetObjectArrayElement,
-                array,
+                array.as_ref().as_raw(),
                 index
             ))
         })
     }
 
-    /// Sets an element of the `jobjectArray` array.
-    pub fn set_object_array_element<'other_local, O>(
+    /// Sets an element of the [`JObjectArray`] `array`.
+    pub fn set_object_array_element<'other_local_1, 'other_local_2>(
         &self,
-        array: jobjectArray,
+        array: impl AsRef<JObjectArray<'other_local_1>>,
         index: jsize,
-        value: O,
-    ) -> Result<()>
-    where
-        O: AsRef<JObject<'other_local>>,
-    {
-        non_null!(array, "set_object_array_element array argument");
+        value: impl AsRef<JObject<'other_local_2>>,
+    ) -> Result<()> {
+        non_null!(array.as_ref(), "set_object_array_element array argument");
         jni_void_call!(
             self.internal,
             SetObjectArrayElement,
-            array,
+            array.as_ref().as_raw(),
             index,
             value.as_ref().as_raw()
         );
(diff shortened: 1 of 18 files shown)


src/wrapper/objects/jobject_array.rs
```diff
@@ -0,0 +1,82 @@
...
+}
+
+impl<'local> std::default::Default for JObjectArray<'local> {
+    fn default() -> Self {
+        Self(JObject::null())
+    }
+}
+
+impl<'local> JObjectArray<'local> {
+    /// Creates a [`JObjectArray`] that wraps the given `raw` [`jclass`]
```
argv-minus-one, line 59: It should say `jobjectArray` here, not `jclass`.

src/wrapper/objects/auto_array.rs
```diff
@@ -17,9 +21,9 @@ use crate::{errors::*, objects::JObject, sys, JNIEnv};
 /// The `get` method must return a valid pointer to the beginning of the JNI array.
 ///
 /// The `release` method must not invalidate the `ptr` if the `mode` is [`sys::JNI_COMMIT`].
-pub unsafe trait TypeArray {
+pub unsafe trait TypeArray: Copy {
     /// getter
-    fn get(env: &mut JNIEnv, obj: &JObject, is_copy: &mut jboolean) -> Result<*mut Self>;
+    fn get(env: &mut JNIEnv, array: jarray, is_copy: &mut jboolean) -> Result<*mut Self>;
```
argv-minus-one, line 26: This method now takes and uses a raw pointer (`jarray`), and it's publicly accessible, so it should be made `unsafe`.

src/wrapper/objects/auto_array.rs
```diff
@@ -92,35 +90,42 @@ type_array!(jdouble, GetDoubleArrayElements, ReleaseDoubleArrayElements);
 ///
 /// AutoArray provides automatic array release through a call to appropriate
 /// Release<Type>ArrayElements when it goes out of scope.
-pub struct AutoArray<'local, T: TypeArray> {
-    obj: JObject<'local>,
+pub struct AutoArray<'a, 'local: 'a, 'other_local: 'a, T: TypeArray + 'other_local> {
```
argv-minus-one, line 93: What's the lifetime `'a` for?

argv-minus-one:
Nice. My project mostly uses Java byte arrays to move data between Java and Rust, so this is just what I needed. I'll give this a try tomorrow.

rib:
just to add; I had intended to open this as a "draft" PR but just noticed I must have clicked the wrong button in the end when opening the PR. This was a first pass that I did fairly quickly yesterday but wanted to post here to get some initial feedback. Thanks for taking a look.

src/wrapper/objects/auto_array.rs
rib: Right. In general it looks like the intent is that this trait is effectively private but the methods should still be marked as `unsafe`.

I wonder if the trait should also be moved into a private module and marked with `#[doc(hidden)]` to further hide the trait as an implementation detail.

src/wrapper/objects/auto_array.rs
rib: Maybe naming this `'ptr` would be better, with docs something like this?:

```rust
/// # Lifetimes
///
/// While accessing the elements of an array there are potentially two local
/// reference frames involved and the temporary pointer to access the elements
/// is constrained to being valid within the intersection of these local frames.
///
/// - `'other_local` represents the local reference frame that `array` is owned by
/// - `'local` represents the current local reference frame (which may be stacked
///    on top of `'other_local`)
/// - `'ptr` represents the lifetime of the acquired pointer to the `array` elements.
```

rib:
Something I had meant to highlight too was that I ended up removing the `AsRef<JObject<'local>>` implementation for `AutoArray` and `AutoPrimitiveArray`.

On the one hand I wanted to remove the owned `JObject`s that were being created for these structs which looked amiss but, more generally, I think of `.as_ref()` a bit like a type cast and since I don't think either of these types really represent the array itself it feels like a misuse to use `.as_ref()` to query a kind of property instead of acting like a type cast.

I currently feel like `AutoArray` and `AutoPrimitiveArray` are poorly named types because they don't so much represent arrays as much as they represent an auto-release guard for a temporary pointer for accessing the elements of an array.

I think of them more like `AutoElements` and `AutoElementsCritical`.

I could see that it might be desirable to be able to query the link between these and the `array` that's being accessed but I'd think of it like querying a relationship/link to a parent object and I'd expect that to have a more explicit API instead of using `.as_ref()`.

src/wrapper/objects/auto_array.rs
rib: It's also notable that `AutoPrimitiveArray` and `AutoArray` should be consistent here which isn't currently the case

src/wrapper/objects/auto_array.rs
argv-minus-one: People still need to see which types implement the trait.

You might want to use the sealed trait pattern here, like this:

```rust
mod type_array_sealed {
	pub unsafe trait Sealed {
		// Move the `TypeArray` methods into this trait.
	}
	
	impl Sealed for jint {
		// Implement the `TypeArray` methods for `jint` here.
	}
}

pub trait TypeArray: type_array_sealed::Sealed {}

impl TypeArray for jint {}
```

This allows users to see that there is a trait and which types implement it, but hides the actual implementation details.

rib: Thanks, this was the pattern I was roughly recalling but this is a nice clear illustration. It seems like this would make sense for this trait.

argv-minus-one:
Here are my thoughts so far. Let me know if you'd like me to work on any of them.


## Merge `AutoArray` and `AutoPrimitiveArray`

I wonder if we could make `JNIEnv::get_primitive_array_critical` return `AutoArray` too, and remove `AutoPrimitiveArray`. They're pretty similar in API and behavior.

There is one important difference, though: `AutoPrimitiveArray` holds a `&mut JNIEnv`, to prevent the caller from making JNI calls inside the critical region, whereas `AutoArray` contains its own `unsafe_clone`d `JNIEnv`. I think this can be dealt with by adding a lifetime parameter `'env` to `AutoArray` that represents:

* If critical, the lifetime of the `&mut JNIEnv`. This prevents other use of the `JNIEnv` while the `AutoArray` is live.
* If not critical, `'static`. This allows other use of the `JNIEnv` while the `AutoArray` is live.

```rust
pub struct AutoArray<'local, 'other_local, 'env, T> {
    env: AutoArrayKind<'local, 'env>,
    ..
}

enum AutoArrayKind<'local, 'env> {
    // Not a critical `AutoArray`. Contains a `JNIEnv` created by `JNIEnv::unsafe_clone`.
    Normal(JNIEnv<'local>),
    
    // A critical `AutoArray`. Mutably borrows the `JNIEnv` it was created with.
    Critical(&'env mut JNIEnv<'local>),
}

impl<'local> JNIEnv<'local> {
    pub fn get_array_elements<'other_local, T>(
        &mut self,
        ..
    ) -> Result<AutoArray<'local, 'other_local, 'static, T>> {_}

    pub fn get_primitive_array_critical<'env, 'other_local, T>(
        &'env mut self,
        ..
    ) -> Result<AutoArray<'local, 'other_local, 'env, T>> {_}
}
```

The destructor for `AutoArray` would need to look at what variant `self.env` is to decide which JNI function to release the array with. For `AutoArrayKind::Normal`, use `Release<Type>ArrayElements`; for `AutoArrayKind::Critical`, use `ReleasePrimitiveArrayCritical`.


## `Deref` to primitive slice

I'd really like to see an implementation of `Deref<Target = [T]>` and `DerefMut` for `AutoArray<T>` (and `AutoPrimitiveArray<T>`, if we're keeping it). Currently, there is no way to actually *use* `AutoArray` without `unsafe`ly dereferencing the raw array pointer.

To make that work, `AutoArray::new` would need to get the size of the array, convert it to `usize`, and store that where the `Deref` implementation can find it along with the array pointer.


## Use after free

I see that `AutoArray` and `AutoPrimitiveArray` no longer take ownership of the local reference to the array. I agree with you that this is a good idea, but in that case, they should store a borrowed `&JPrimitiveArray`, not just the raw `jarray`. Currently, it looks like you can delete the local reference to the array while an `AutoArray` is still live, causing use-after-free in `AutoArray`'s methods and destructor:

```rust
let mut env: JNIEnv;
let array_ref: JPrimitiveArray;

let auto_array = env.get_int_array_elements(&array_ref, _)?;
env.delete_local_ref(array_ref)?;
auto_array.size()?; // UB: uses the local reference after it is freed!
```


## `JNIEnv::get_<type>_array_elements`

Should the `JNIEnv::get_<type>_array_elements` methods be removed? I don't see any need for them now that there's a generic `JNIEnv::get_array_elements` method.

rib:
For reference, I've made a second pass over this now.

For now I've tried to use a more consistent lifetime name of `'ptr` to name the lifetime of the pointer to the array elements.

I'll look at splitting out the change to add `as_raw()` methods to types like `JString` which I added for consistency.

This iteration also ends up making quite a lot of changes to both `AutoArray` and `AutoPrimitiveArray` for the sake of normalizing their implementation, and I was also starting to consider that they could potentially be unified later through the `TypeArray` trait.

src/wrapper/objects/auto_array.rs
argv-minus-one: I'm afraid I don't understand. The lifetime of the pointer isn't separate from the lifetime of the `AutoArray` wrapping it. The pointer is created by `AutoArray::new` and destroyed by the `AutoArray` destructor.

What's more, the lifetime `'a` comes from a lifetime parameter on `JNIEnv::get_array_elements`, which has no constraints other than that `'local` outlives it, so this lifetime can in fact be *longer* than that of the `AutoArray` itself.

argv-minus-one:
`JString` already has an `as_raw` method from `Deref<Target = JObject>`.

rib:
> ## Merge `AutoArray` and `AutoPrimitiveArray`
> I wonder if we could make `JNIEnv::get_primitive_array_critical` return `AutoArray` too, and remove `AutoPrimitiveArray`. They're pretty similar in API and behavior.
> 
> There is one important difference, though: `AutoPrimitiveArray` holds a `&mut JNIEnv`, to prevent the caller from making JNI calls inside the critical region, whereas `AutoArray` contains its own `unsafe_clone`d `JNIEnv`. I think this can be dealt with by adding a lifetime parameter `'env` to `AutoArray` that represents:

Yeah, I was having similar thoughts.

I was also pondering whether the API to access elements with a "critical" section should simply be marked as `unsafe` (along with clearer docs explaining that JNI can't be used in the critical section) instead of holding mutable env reference to try and block any use of JNI. In the documentation for `GetPrimitiveArrayCritical` there is a notable exception that allows nested critical sections in case you want to access multiple arrays, and if we take the mutable reference to the env then we limit that use case.

rib:
> `JString` already has an `as_raw` method from `Deref<Target = JObject>`.

ah, thanks for the reminder. I guess I was initially adding `as_raw()` implementations that would return the more specific sys types for the arrays and then thought it was maybe an oversight that we hadn't added `as_raw()` methods for `JString` and `JThrowable`. I've dropped those now.

rib:
> I see that `AutoArray` and `AutoPrimitiveArray` no longer take ownership of the local reference to the array. I agree with you that this is a good idea, but in that case, they should store a borrowed `&JPrimitiveArray`, not just the raw `jarray`. Currently, it looks like you can delete the local reference to the array while an `AutoArray` is still live, causing use-after-free in `AutoArray`'s methods and destructor:

Ah yep, makes sense.

argv-minus-one:
I notice that this PR also contains https://github.com/jni-rs/jni-rs/pull/400/commits/910ee0d35e5048b97fdb587dfd24b3d9f6dabedc and https://github.com/jni-rs/jni-rs/pull/400/commits/eb82c38cfb016201042fcca634f1032bd58d911b, which don't have anything to do with arrays. Is that intentional? If not, you can remove them from this branch with `git rebase --onto master eb82c38` (assuming your local `master` is the same as what's in this GitHub repository).

rib:
> I notice that this PR also contains [910ee0d](https://github.com/jni-rs/jni-rs/commit/910ee0d35e5048b97fdb587dfd24b3d9f6dabedc) and [eb82c38](https://github.com/jni-rs/jni-rs/commit/eb82c38cfb016201042fcca634f1032bd58d911b), which don't have anything to do with arrays. Is that intentional? If not, you can remove them from this branch with `git rebase --onto master eb82c38` (assuming your local `master` is the same as what's in this GitHub repository).

Yeah, sorry for the distraction - I originally noted this in the PR description. When I first made the PR it was intended to be a draft to test out the feasibility of adding these wrapper types and I just did that based on the branch for #399. I used that branch as a base mainly for the sake of being based on the `non-Copy` local ref changes which would have led to lots of conflicts to rebase. I've just rebased on master.

src/wrapper/objects/auto_array.rs
rib: hmm I've possibly made a bit of a muddle of this `'a`/`'ptr` lifetime.

At least in the case of `AutoPrimitiveArray` it seems worth noting that previously the pointer was actually acquired _before_ calling `AutoPrimitiveArray::new()` so in some sense the lifetime was longer (though I also didn't initially add a named pointer lifetime there)

As you say the lifetime of the pointer isn't separate from the lifetime of the `AutoArray` wrapping it but I think by naming the pointer lifetime I had thought I was constraining that the `AutoArray` itself wouldn't be able to move outside the intersection between `'local` and `'other_local`. I'm a bit confused now about why I thought that was necessary though tbh.

I'll aim to drop the `'ptr` lifetime and also borrow the `JPrimitiveArray` as suggested

rib: okey the latest iteration removes the `'ptr` lifetime and adds an `'array` lifetime for borrowing the `JPrimitiveArray`.

rib:
@argv-minus-one I've just added a patch that implements `Deref` and `DerefMut` and updated the tests to check these. This makes it so the array length is implicitly called during `::new()` to ensure we always know the array length upfront. These internally call a `new_with_len()` constructor in case we ever want to create an `AutoArray` with a pre-determined length.

`.size()` is replaced with `.len()` that doesn't return a `Result`.

rib:
Regarding the idea to try and consolidate `AutoArray` and `AutoPrimitiveArray` I'm currently a little reluctant to look at that here currently (though still open to experimenting later via a separate PR) my current gut feeling is that it may end up kind of klunky / over complicated having to deal with the env ownership differences. It's also notable that `.commit()` is another difference since `JNI_COMMIT` is fraught with issues across different JNI implementations when using critical sections.

That said - I'd be very tempted to rename `AutoPrimitiveArray` to something like `AutoArrayCritical` to at least try and better represent the difference between these two APIs.

argv-minus-one:
I've changed my project to use these new features, and they work very nicely.

I did need to use the [bytemuck](https://crates.io/crates/bytemuck) library to get a `[u8]` from a Java byte array, which normally gives `[i8]`. I imagine this will be a common need whenever using Java byte arrays to move serialized data across JNI, so it might be worth recommending bytemuck in the documentation.

maurolacy:
> Regarding the idea to try and consolidate `AutoArray` and `AutoPrimitiveArray` I'm currently a little reluctant to look at that here currently (though still open to experimenting later via a separate PR) my current gut feeling is that it may end up kind of klunky / over complicated having to deal with the env ownership differences.

I remember trying to do that at the time, and in the end deciding that it was simpler / clearer to have just two different impls.

> It's also notable that `.commit()` is another difference since `JNI_COMMIT` is fraught with issues across different JNI implementations when using critical sections.

Yes. JNI_COMMIT is not even consistent in openjdk, depending on the flags, in particular `-Xcheck:jni`.

> 
> That said - I'd be very tempted to rename `AutoPrimitiveArray` to something like `AutoArrayCritical` to at least try and better represent the difference between these two APIs.

Go ahead. I tried to come up with representative names at the time; having `Critical` in the name sounds like a good idea.

rib:
Okey, not really wanting to bike shed naming but still; @argv-minus-one + @maurolacy can you let me know if you'd be ok with the names `AutoElements` / `AutoElementsCritical` or would you rather go with the more similar `AutoArray` / `AutoArrayCritical` names?

Personally I think that the use of `Array` in the name is a little misleading - more so now that this PR adds `JPrimitiveArray` + `JByteArray`/`JIntArray` etc as a wrapper type for Array references.

The `AutoArray` type doesn't own the array itself - it's an auto-release guard that gives temporary access to the elements of an array so I tend to think that `AutoElements` / `AutoElementsCritical` help to intuitively differentiate what these are for compared to `JPrimitiveArray` + `JByteArray` etc.

Please react with 🎉 for `AutoArray` / `AutoArrayCritical` or 🚀 for `AutoElements` / `AutoElementsCritical`

maurolacy:
> Please react with 🎉 for `AutoArray` / `AutoArrayCritical` or 🚀 for `AutoElements` / `AutoElementsCritical`

Both work for me, but I tend to stick with 'Array' for historical reasons.

rib:
> I've changed my project to use these new features, and they work very nicely.
> 
> I did need to use the [bytemuck](https://crates.io/crates/bytemuck) library to get a `[u8]` from a Java byte array, which normally gives `[i8]`. I imagine this will be a common need whenever using Java byte arrays to move serialized data across JNI, so it might be worth recommending bytemuck in the documentation.

It could even be worth adding a `sys` type for casting bytes as unsigned. If we added a `jubyte` that would allow Rust code to interpret any Java byte as unsigned if that's what makes sense.

rib:
> > Please react with 🎉 for `AutoArray` / `AutoArrayCritical` or 🚀 for `AutoElements` / `AutoElementsCritical`
> 
> Both work for me, but I tend to stick with 'Array' for historical reasons.

One other little thing to note is that the C JNI functions used by `AutoArray` for mapping the elements are named like `Get<Type>ArrayElements` so I also think that `AutoElements` has a reasonable enough nod back to naming used in the C API. (On the other hand `AutoArrayCritical` or `AutoElementsCritical` are less connected with the C API naming but tbh I think that's in part because the C API would have been overly verbose if they had used the name `GetPrimitiveArrayElementsCritical`, even though that would have been more consistent with the `Get<Type>ArrayElements` functions.)

I'd usually be very reluctant to change API names that introduce API breakages just for the sake of if but I'm not massively concerned about that in this case, mainly because:
1. the 0.21 release is going to be a fairly major release considering the `non-Copy` changes for references
2. this PR is making a sweeping change to these APIs that will require a little bit of migration effort anyway.
3. I guess at the moment these APIs aren't particularly widely used yet

so I feel like this would be the sensible time to rename if it would be beneficial

rib:
One other thing I'm a little unsure about atm is: what to do with `JNIEnv::get_array_length()`?

It needs to be possible to query the length of both `Primitive` and `Object` arrays and so currently this is the last `JNIEnv` array API left that still takes a `jarray` `sys` type.

Maybe for now we can make `get_array_lenth()` accept a `JPrimitiveArray` and then add a `get_object_array_length()` as a special case that takes a `JObjectArray` reference.

rib:
Thinking about adding some migration guide notes about the `Deref`/`DerefMut` trait implementations has got me thinking about what safety issues should be discussed here, if any.

`jni-rs` is obviously an FFI abstraction and so there is inherently some shared idea of what constitutes "initialized" data when talking about accessing array elements.

Considering that we're only talking about primitive types (`JObjectArray` is handled as a special case) I think it's pragmatic that `jni-rs` would simply assumes it's always valid to treat array values as "initialized" but I think there is one case which could potentially lead to undefined behaviour within the JVM which is `jboolean` arrays.

I _think_ as far as Rust goes then there's nothing that concerning with `jboolean` elements which map to `u8` (and so there's no bitwise representation that can be considered an invalid `u8`) but on the _Java_ side only the values `0` and `1` are valid boolean values as far as I understand it and so code that uses `AutoArray` to access a `jboolean` array has some responsibility to ensure that it only stores `0` or `1` in any element to avoid undefined behaviour when that data is read by the JVM.

It seems like it would be a pain to special case `jboolean` and also excessive to keep it `unsafe` to access all primitive array elements so I'm not really thinking this realization needs to affect the introduction of `Deref`/`DerefMut` technically. It's also notably not the only place where code using JNI needs to be careful to only pass booleans to Java that have a value of `0` or `1` (any native JNI function returning a `jboolean` also needs to be careful about this).

Mostly just thinking out loud here since this all just occurred to me.

rib:
considering above; I suppose potentially it could make sense if `sys:jboolean` was instead represented as:

```rust
#[repr(u8)]
enum jboolean {
    False=0,
    True=1
}
```

so it would be equally undefined on the Rust and Java side to have boolean values that don't equal `0` or `1`

rib:
Another detail is that `get_array_elements` probably needs to take a mutable reference to the `JPrimitiveArray` - otherwise it would be very easy to create multiple `AutoArray`s for the same array and then use `DerefMut` to create mulitple mutable references to slices that alias each other - which would be unsound.

rib:
Okey I've taken another pass over this with these notable updates:


1. `get_primitive_array_critical` is now marked as `unsafe` and renamed `get_array_elements_critical` and I have tried to add more detailed documentation about the safety considerations within the critical section defined by this API (also updating the docs for `get_array_elements` to be consistent)
2. `AutoArray` and `AutoArrayCritical`  now take a mutable reference to the source array. As noted above the introduction of `DerefMut` could lead to unsound code if `get_array_elements` was called multiple times to create aliasing `AutoArray`s (since you could get multiple mutable references to slices that alias each other).
3. the file `primitive_array.rs` was renamed `jprimitive_array.rs` for consistency
4. `get_array_length` expects a type that implements `AsJArrayRaw` which provides `.as_jarray_raw()` similar to `JObject::as_raw()`. This trait is implemented by `JPrimitiveArray` and `JObjectArray`.


_P.S Since I just noticed that we have a tie over the naming preference and also considering that @maurolacy did suggest that either name would be acceptable - I'm probably going also add a patch to rename to `AutoElements` + `AutoElementsCritical`_

rib:
Okey, I feel like this is hopefully in good shape to land at this point if looks good to you @argv-minus-one + @maurolacy ?

rib:
After seeing this issue: https://github.com/jni-rs/jni-rs/issues/397 and quickly trying to build `example/mylib` I realized we recently broke that example with the `non-Copy` reference API changes and I also broke it further with this PR.

Ideally that example should be built by CI to avoid this but for now I've added a patch on top of this PR that updates the example based on these changes.

Actually in this case the broken example helped highlight that `convert_byte_array` was taking ownership of the `JByteArray` argument instead of borrowing it. I've updated `convert_byte_array` to accept an `array` as `array: &impl AsRef<JByteArray<'other_local>>`

argv-minus-one:
> It could even be worth adding a `sys` type for casting bytes as unsigned. If we added a `jubyte` that would allow Rust code to interpret any Java byte as unsigned if that's what makes sense.

Wouldn't that be incorrect? The Java `byte` type is always signed.

> `jni-rs` is obviously an FFI abstraction and so there is inherently some shared idea of what constitutes "initialized" data when talking about accessing array elements.
> 
> Considering that we're only talking about primitive types (`JObjectArray` is handled as a special case) I think it's pragmatic that `jni-rs` would simply assumes it's always valid to treat array values as "initialized"

Java arrays are *always* initialized upon allocation with the default value of their type (zero, false, or null for numeric primitive, boolean, and reference types, respectively), so yes, it's safe to make that assumption.

> but I think there is one case which could potentially lead to undefined behaviour within the JVM which is `jboolean` arrays.
> 
> I _think_ as far as Rust goes then there's nothing that concerning with `jboolean` elements which map to `u8` (and so there's no bitwise representation that can be considered an invalid `u8`) but on the _Java_ side only the values `0` and `1` are valid boolean values as far as I understand it and so code that uses `AutoArray` to access a `jboolean` array has some responsibility to ensure that it only stores `0` or `1` in any element to avoid undefined behaviour when that data is read by the JVM.
> 
> It seems like it would be a pain to special case `jboolean` and also excessive to keep it `unsafe` to access all primitive array elements so I'm not really thinking this realization needs to affect the introduction of `Deref`/`DerefMut` technically. It's also notably not the only place where code using JNI needs to be careful to only pass booleans to Java that have a value of `0` or `1` (any native JNI function returning a `jboolean` also needs to be careful about this).
> 
> Mostly just thinking out loud here since this all just occurred to me.
>
> ---
>
> considering above; I suppose potentially it could make sense if `sys:jboolean` was instead represented as:
> 
> ```rust
> #[repr(u8)]
> enum jboolean {
>     False=0,
>     True=1
> }
> ```
> 
> so it would be equally undefined on the Rust and Java side to have boolean values that don't equal `0` or `1`

Per [the Rust Reference](https://doc.rust-lang.org/reference/types/boolean.html), that's exactly how the Rust `bool` type is defined, so we should be able to get away with simply redefining `sys::jboolean` as a type alias for `bool`.

> Another detail is that `get_array_elements` probably needs to take a mutable reference to the `JPrimitiveArray` - otherwise it would be very easy to create multiple `AutoArray`s for the same array and then use `DerefMut` to create mulitple mutable references to slices that alias each other - which would be unsound.

Yikes. I didn't realize that this would be a problem, but now that you mention it, it's actually worse than that. Java code on another thread can read/write the array at the same time as Rust code thinks it has exclusive/immutable access, resulting in undefined behavior on the Rust side.

I've spent a few hours doing some reading and thinking on the subject, and it looks like there isn't any way to read or write a memory address in Rust without undefined behavior if there is a data race, not even with `std::ptr::{read,write}_volatile`.

The problem is that Java has a memory model that says what happens when there's a data race. Rust is a bit underspecified in this area: in Rust, a data race causes undefined behavior, no exceptions, end of story. This leaves us only one way to read/write a Java primitive array in the presence of a data race from other Java threads without causing undefined behavior: copying through `{Get,Set}<Type>ArrayRegion`, which *does* implement Java's memory model. Unfortunately, that's slow.

For this reason, anything that allows reading/writing through the pointer returned by `GetPrimitiveArrayCritical` or `Get<Type>ArrayElements` has to be `unsafe`, and the caller has to promise that, until the `AutoElements`/`AutoElementsCritical` is dropped, no Java or Rust code will access the array by any other means.

Which means `JNIEnv::get_array_elements` and its wrappers need to be `unsafe` too. :disappointed:

> Okey I've taken another pass over this with these notable updates:
> 
> 1. `get_primitive_array_critical` is now marked as `unsafe` and renamed `get_array_elements_critical` and I have tried to add more detailed documentation about the safety considerations within the critical section defined by this API (also updating the docs for `get_array_elements` to be consistent)

Looks mostly good to me. You're right, critical adds a bunch of rules that Rust borrowing can't enforce and will cause UB if broken (in addition to the data race problems mentioned above), so it'll have to be `unsafe`.

One problem, though: the documentation you wrote says, “[a]ny use of system calls that will wait for a result from another thread could deadlock if that other thread is blocked by the paused garbage collector.” This implies that a deadlock is the worst that can happen if you break the rules. That might be true in practice with the current version of HotSpot, but the JNI spec doesn't actually make such a promise, or any promises at all. [Anything](http://www.catb.org/jargon/html/N/nasal-demons.html) could happen with other/future JVMs, and the documentation shouldn't imply otherwise.

> 2. `AutoArray` and `AutoArrayCritical`  now take a mutable reference to the source array. As noted above the introduction of `DerefMut` could lead to unsound code if `get_array_elements` was called multiple times to create aliasing `AutoArray`s (since you could get multiple mutable references to slices that alias each other).

Since that's not enough to prevent UB from the array being read/written by other Java threads, I'm not sure this change is actually useful.

> 3. the file `primitive_array.rs` was renamed `jprimitive_array.rs` for consistency
>
> 4. `get_array_length` expects a type that implements `AsJArrayRaw` which provides `.as_jarray_raw()` similar to `JObject::as_raw()`. This trait is implemented by `JPrimitiveArray` and `JObjectArray`.

Looks good to me.

> _P.S Since I just noticed that we have a tie over the naming preference and also considering that @maurolacy did suggest that either name would be acceptable - I'm probably going also add a patch to rename to `AutoElements` + `AutoElementsCritical`_

OK. I don't have a strong preference either way. The important part is that the names are “*X*” and “*X*Critical”. This makes the difference between them clear.

rib:
> > It could even be worth adding a `sys` type for casting bytes as unsigned. If we added a `jubyte` that would allow Rust code to interpret any Java byte as unsigned if that's what makes sense.
> 
> Wouldn't that be incorrect? The Java `byte` type is always signed.

those signed bytes are 8 bits and twos compliment, so there's not really any such thing as an invalid byte. A valid signed byte is also a valid unsigned byte and that's also how / why bytemuck can let you do the casting safely too.

practically I guess there might be some annoying fiddly cases though if we introduced a `jubyte` so I'm not exactly sure it'd be the best approach even though it could avoid the need to use bytemuck in this case.

> Java arrays are always initialized upon allocation with the default value of their type (zero, false, or null for numeric primitive, boolean, and reference types, respectively), so yes, it's safe to make that assumption.

Although this is relevant too, my primary basis for saying that it's ok for Rust to assume they are initialized is because they are primitive types that (with the exception of jboolean) have no invalid values.

Terminology is a bit tricky here since we're talking about the intersection of memory models between two languages and it's easy to switch up what we mean when saying memory is "initialized" (and Rust doesn't even have an official model yet :-))

I'm more thinking about the Rust side though and effectively considering the Java side a black box for simplicity (e.g. imagine there's an obscure Java API that lets you `mmap` a shared file and somehow wrap that as an array).

As an illustration of the different terminology: Although Java will make sure to zero initialize new allocations, it potentially wouldn't be unreasonable to expose such an allocation to Rust as "uninitialized" in some situations, from the pov that the allocation doesn't contain anything you care about because you want to explicitly tell the compiler that it can apply optimizations that mean it doesn't need to worry about readying back and preserving anything from that allocation. (so any operation doing a read/modify/write could be free to skip the read)

> Per [the Rust Reference](https://doc.rust-lang.org/reference/types/boolean.html), that's exactly how the Rust bool type is defined, so we should be able to get away with simply redefining sys::jboolean as a type alias for bool

Yeah we basically want the same as `bool` though I'm not sure if we should directly use bool. Rust doesn't really guarantee much about ABI yet - e.g. that reference says "this book is not normative. It may include details that are specific to rustc itself, and should not be taken as a specification for the Rust language." and so to some extent it's conceptually an implementation detail that `bool` is what we want here.

Practically though it seems pretty likely that it'd be fine to rely on this so maybe it would be OK.

If `jni-rs` defined its own enum then we can be a little bit more transparent about it being `#[repr(u8)]` and clearer that `True` = `1` and `False` = `0` without anyone needing to look at a language reference to check these details.

> Yikes. I didn't realize that this would be a problem, but now that you mention it, it's actually worse than that. Java code on another thread can read/write the array at the same time as Rust code thinks it has exclusive/immutable access, resulting in undefined behavior on the Rust side.

> I've spent a few hours doing some reading and thinking on the subject, and it looks like there isn't any way to read or write a memory address in Rust without undefined behavior if there is a data race, not even with std::ptr::{read,write}_volatile.

right, any read/write races are canonically "undefined behaviour". That's not necessarily a huge issue if its something you're aware of (and have some locking scheme to avoid) but it probably would make sense to make `get_array_elements` unsafe too and add `#Safety` notes (for `get_array_elements_critical` too) that say it's currently your responsibility to manage any data races from concurrent access to the array.

We could potentially create a `MonitorGuard` when constructing an `AutoElements[Critical]` as a way to help ensure mutual exclusion. This way `AutoElements` would act similar to a `MutexGuard` and would unlock the array when dropped (though I expect it will be awkward knowing what to do with any JNI errors within Drop).

This would still require some cooperation on the Java side to use `synchronized(array) { }` locks when modifying shared arrays, which we can't guarantee from Rust but it would at least offer mutual exclusion for Rust code in different threads and make it very easy for Java code to participate. This would preclude more fine grained shared access to arrays which might be desirable in some cases.

> Which means JNIEnv::get_array_elements and its wrappers need to be unsafe too. 😞

Luckily we were able to delete all the wrappers, as you suggested (as redundant) so actually the only change is making `get_array_elements` `unsafe` (since `get_array_elements_critical` is already unsafe)

This isn't actually that onerous, since we can front-load the `unsafe` action of creating an `AutoElements` guard  (where the `# Safety` docs can explain that it's unsafe because you're responsible for managing any shared access) and then I think it'd be reasonable that `AutoElements` would still provide "safe" `Deref`/`DerefMut` for more ergonomic access to the elements.

> One problem, though: the documentation you wrote says, “[a]ny use of system calls that will wait for a result from another thread could deadlock if that other thread is blocked by the paused garbage collector.” This implies that a deadlock is the worst that can happen if you break the rules. That might be true in practice with the current version of HotSpot, but the JNI spec doesn't actually make such a promise, or any promises at all. [Anything](http://www.catb.org/jargon/html/N/nasal-demons.html) could happen with other/future JVMs, and the documentation shouldn't imply otherwise.

I'm not generally a big fan of being overly hand wavy - otherwise it's easy to lose sight of why the constraints/rules exist without understanding them. The JNI spec is a typical design-by-committee spec that's written in more open-ended language but I've tried to distill that into a more practical break down of _why_ the JNI spec has those rules.

Maybe I can find a compromise that both states what the JNI spec says - and the possibility of a deadlock can serve as an illustration of what could happen. I'm also trying to avoid copying wording from the spec since Oracle is such a terrible company regarding copyright ;-p

The spec language doesn't really go out of it's way to claim that "Anything" could happen and doesn't have any formal idea of "undefined behaviour" itself. It also makes it clear enough that garbage collection may be disabled in the critical section and that the problem with making system calls is that they may create a blocking dependency between the current thread and some other Java thread (that may be blocked by the disabled garbage collection) which would cause a deadlock: "or any system call that may cause the current thread to block and wait for another Java thread".

A JNI implementation could also potentially detect the same condition and `abort()` instead of actually deadlocking or anything else but the constraints boil down to avoiding a deadlock condition (regardless of how that would technically manifest if you broke the rules)

I'll take another pass over these docs to see if I can convey the more open ended wording from the JNI spec too.

> Since that's not enough to prevent UB from the array being read/written by other Java threads, I'm not sure this change is actually useful

It's the difference between undefined behaviour and _unsound_ Rust code.

All the data race concerns with undefined behaviour can potentially be addressed by initially marking `get_array_elements` as `unsafe` and making the caller responsible for considering these. undefined behaviour isn't generally desirable and ideally shouldn't be possible with safe code but it's often an acceptable trade off too.

The mutable borrow here was actually to avoid _unsound_ Rust which is a much more serious concern to my mind - even unsafe code is not allowed to be unsound.

I'm not sure if there's really a standard definition for "undefined behaviour" vs "unsound Rust code" but the issue here is that the Rust compiler would be free to make a total mess if you you start trying to juggle multiple mutable references that alias each other in the same code. That's what I'm referring to as "unsound" - it's fundamentally breaking the rules of the language.

The implementation of `DerefMut` now makes it possible to get a `&mut [T]` instead of a raw pointer and unlike raw pointers Rust mutable references are not allowed to alias.

If you're familiar with C and the `restrict` keyword then in Rust it's as if `&mut [T]` is a pointer that is also implicitly marked with `restrict` which tells the compiler that you're promising that it doesn't alias with any other reference pointers it has. E.g. considering the difference between `memmove` and `memcpy` if you want to copy data with a source and destination reference then Rust can assume it doesn't have to worry about those memory regions overlapping. If you did the same kind of copy in Rust via raw pointers though the compiler would have to consider that some portion of the src/dest regions might overlap.

I.e. in Rust aliasing raw pointers can be OK (even though you might be able to observe undefined behaviour via data races) but aliasing mutable references are unsound and can lead to code being compiled incorrectly.

Even if we require `unsafe` code and potentially start taking a `MonitorGuard` lock I think it could still make sense to have this further mutable borrow restriction to ensure you can't easily create multiple `AutoElements` instances that alias each other (in case code would then _also_ use `DerefMut` and create multiple mutable references that alias)

Hope that clarifies where my original concern came from - I was actually completely overlooking the data race side of things at the time.

argv-minus-one:
> We could potentially create a `MonitorGuard` when constructing an `AutoElements[Critical]` as a way to help ensure mutual exclusion. This way `AutoElements` would act similar to a `MutexGuard` and would unlock the array when dropped (though I expect it will be awkward knowing what to do with any JNI errors within Drop).

If you do that, there should be a way to opt out if data races are being prevented some other way. Keep in mind that acquiring a lock is pretty costly and that the whole point of `Get<Type>ArrayElements` and `GetPrimitiveArrayCritical` is performance.

For instance, in my project, byte arrays are used for passing messages between Rust and Java code. The sender allocates an array, writes to it, calls the receiver, and then drops the reference to the array. The sender never reads or writes the array after calling the receiver and the array is never read or written on another thread, so unless I'm mistaken, locking is not needed.

> This isn't actually that onerous, since we can front-load the `unsafe` action of creating an `AutoElements` guard (where the `# Safety` docs can explain that it's unsafe because you're responsible for managing any shared access) and then I think it'd be reasonable that `AutoElements` would still provide "safe" `Deref`/`DerefMut` for more ergonomic access to the elements.

Sounds good.

> I'm not sure if there's really a standard definition for "undefined behaviour" vs "unsound Rust code"

[The Unsafe Code Guidelines has a definition.](https://github.com/rust-lang/unsafe-code-guidelines/blob/5325cdedeca312dd242227fc8899c150356ec965/reference/src/glossary.md#undefined-behavior)

> Even if we require `unsafe` code and potentially start taking a `MonitorGuard` lock I think it could still make sense to have this further mutable borrow restriction to ensure you can't easily create multiple `AutoElements` instances that alias each other (in case code would then _also_ use `DerefMut` and create multiple mutable references that alias)

It can still be done by duplicating the reference to the array with `JNIEnv::new_local_ref`. Java monitor locks are reentrant, so acquiring a lock won't prevent aliasing on the same thread either.

rib:
> If you do that, there should be a way to opt out if data races are being prevented some other way. Keep in mind that acquiring a lock is pretty costly and that the whole point of `Get<Type>ArrayElements` and `GetPrimitiveArrayCritical` is performance.

yeah this is generally why I'd be a bit reluctant to do this - vs simply marking `get_array_elements` as `unsafe` and documenting that its the caller's responsibility to consider data races.

> It can still be done by duplicating the reference to the array with JNIEnv::new_local_ref. Java monitor locks are reentrant, so acquiring a lock won't prevent aliasing on the same thread either.

About the re-entrancy I had been thinking that was another reason why it could be beneficial to have the mutable borrow - since it would help block reentrant access to the same array in the same thread while the monitor locks would avoid concurrent access across threads.

Yeah, having to consider the possibility of `new_local_ref` (or other various ways code can get multiple references for the same array) would certainly let you bypass the restriction from the mutable borrow, so it's true that it's not a water-tight solution but it still seems sensible to leverage the borrow checker to at least make it harder to trigger undefined behavior unwittingly and to also help document the intent.

Since the api would be `unsafe` we can cover the other cases with the `# Safety` rules but to try and minimize the burden of responsibility away from those `unsafe` rules it seems good to try and utilize the language where we practically can?

We could potentially utilize some thread local storage in the future (would want to defer this to a separate PR I think) to try and address this issue arising from multiple references that point to the same array. E.g. a stack of array references that are currently being accessed may be feasible to maintain and easy (+ relatively cheap) to check against via `env.IsObjectEqual`. In conjunction with `MonitorGuard` for blocking other threads, that could be quite reliable for a safer API.

# What should we do here?

Although I'm tempted to consider the idea of utilizing a `MonitorGuard` I also worry that that will block too many practical use cases that could make sure they avoid data races in a more fine-grained fashion.

For now I think I'd rather reserve that idea for a later addition - maybe we could add `.get_array_elements_locked` for 0.22 if others show an interest in that.

I'm thinking we should just mark `get_array_elements` as `unsafe` and in light of the additional safety observations recently we should extend the `# Safety` docs to note:
1. callers are responsible for avoiding any undefined behaviour as a result of data races while accessing the elements of an array. (can e.g. suggest the use of their own `MonitorGuard`)
2. callers are responsible for not allowing more than one `AutoElements[Critical]` instance to exist per Java array object at any time (even if no data race would exist).

(2) is a broader restriction than is strictly necessary but it also seems a whole lot simpler to convey than wording the restriction in terms of avoiding mutable reference aliasing via`DerefMut`

Regarding the brief discussion about `jboolean` arrays above I also tend to think this issue should be considered separate from this. I like the idea of defining `jboolean` as either `bool` or our own enum but that's a broader / orthogonal issue.


Overall though - dragons beware with these APIs!

rib:
oh btw, one other detail I noticed is that `define_class_bytearray` was taking ownership of its `buf: AutoElements` argument instead of taking a reference, so I'll also fix that.

rib:
Okey:

- I've made `get_array_elements` `unsafe` and taken another pass over the docs, adding the additional safety notes discussed above
- define_class_bytearray now borrows the `buf` argument.

@argv-minus-one do we agree now that it's reasonable to have `get_array_elements[_critical]` borrow the array reference mutably as a way to help avoid aliasing - even though it's understood that it's not a water-tight solution? (it sounded like you were thinking about it from the pov of data races / locking before which wasn't really the reason for the change)

As noted above - this is still punting the question of making `jboolean` an alias for `bool` or an enum, to be considered in a separate issue.

Potentially this is good to land now.

maurolacy:
> Okey, I feel like this is hopefully in good shape to land at this point if looks good to you @argv-minus-one + @maurolacy ?

OK with me about the naming and stuff. I didn't took a detailed look at the code, and thought about it, so cannot really comment about all the other changes you introduced.

argv-minus-one:
I've pushed commit https://github.com/jni-rs/jni-rs/pull/400/commits/1e19b0108e1f7ed79080e59a7b6e38a20c0d0139 to your branch with some minor documentation adjustments. I hope you don't mind.

In the 0.21 migration guide, I changed “to ensure you can't create more than one mutable `&mut [T]`” to “to **help** ensure …” to make clear that this doesn't completely protect from that possibility.

The rest of your changes look good to me.

> We could potentially utilize some thread local storage in the future (would want to defer this to a separate PR I think) to try and address this issue arising from multiple references that point to the same array. E.g. a stack of array references that are currently being accessed may be feasible to maintain and easy (+ relatively cheap) to check against via `env.IsObjectEqual`. In conjunction with `MonitorGuard` for blocking other threads, that could be quite reliable for a safer API.

That should prevent data races and invalid aliasing in Rust code, yes. Data races with Java code are still possible, though.

Be careful that array references are removed from the stack at the right time even if a destructor doesn't get called (`std::mem::forget`, etc). Otherwise a dangling pointer could end up in the stack, causing undefined behavior next time a check is performed.

The only way I can think of to make that work is for `JNIEnv::get_array_elements{,_critical}` to take a closure that takes a parameter of type `&mut AutoElements{,Critical}` (and `&mut JNIEnv`). `JNIEnv::get_array_elements{,_critical}` would:

1. Add the pointer to the stack.
2. Run the closure using `std::panic::catch_unwind`.
3. Remove the pointer from the stack.
4. If the closure panicked, `std::panic::resume_unwind`. Otherwise, pass through whatever value the closure returned.

Benefits of this approach:

* It *should* ensure that the pointer is reliably removed from the stack.
* `JNIEnv::get_array_elements{,_critical}` won't have to search the stack for the correct pointer to remove. It will always be the last one. (Checking with a `debug_assert_eq!` might be wise.)

This approach has the drawback of being less ergonomic, though.

> @argv-minus-one do we agree now that it's reasonable to have `get_array_elements[_critical]` borrow the array reference mutably as a way to help avoid aliasing - even though it's understood that it's not a water-tight solution?

It just occurred to me that requiring `&mut JPrimitiveArray` makes it impossible to use `JNIEnv::get_array_elements{,_critical}` if the array reference is a `GlobalRef`. Users who have a global reference to a Java primitive array will have to create a local reference from it before they can use `get_array_elements{,_critical}`. Is that acceptable? If so, should it be documented?

Other than that, yes.

rib:
> I've pushed commit [1e19b01](https://github.com/jni-rs/jni-rs/commit/1e19b0108e1f7ed79080e59a7b6e38a20c0d0139) to your branch with some minor documentation adjustments. I hope you don't mind.

cool, thanks I'll take a look


> > We could potentially utilize some thread local storage in the future (would want to defer this to a separate PR I think) to try and address this issue arising from multiple references that point to the same array. E.g. a stack of array references that are currently being accessed may be feasible to maintain and easy (+ relatively cheap) to check against via `env.IsObjectEqual`. In conjunction with `MonitorGuard` for blocking other threads, that could be quite reliable for a safer API.
> 
> That should prevent data races and invalid aliasing in Rust code, yes. Data races with Java code are still possible, though.

yeah on the java side the assumption is that it would have to cooperate with taking the same monitor locks (e.g. via `synchronized(array) { }` blocks when accessing the array - but yeah, we can't control that from the Rust side.

> 
> Be careful that array references are removed from the stack at the right time even if a destructor doesn't get called (`std::mem::forget`, etc). Otherwise a dangling pointer could end up in the stack, causing undefined behavior next time a check is performed.
> 
> The only way I can think of to make that work is for `JNIEnv::get_array_elements{,_critical}` to take a closure that takes a parameter of type `&mut AutoElements{,Critical}` (and `&mut JNIEnv`). `JNIEnv::get_array_elements{,_critical}` would:

moving the critical section to a closure might be helpful for this yep.

> 
> 1. Add the pointer to the stack.
> 2. Run the closure using `std::panic::catch_unwind`.
> 3. Remove the pointer from the stack.
> 4. If the closure panicked, `std::panic::resume_unwind`. Otherwise, pass through whatever value the closure returned.
> 
> Benefits of this approach:
> 
> * It _should_ ensure that the pointer is reliably removed from the stack.
> * `JNIEnv::get_array_elements{,_critical}` won't have to search the stack for the correct pointer to remove. It will always be the last one. (Checking with a `debug_assert_eq!` might be wise.)
> 
> This approach has the drawback of being less ergonomic, though.

I'd maybe also want to see if we can handle this without resorting to `catch_unwind` perhaps (e.g. maybe we can just have a canary that would poison the stack state if dropped during an unwind) but maybe that'd be necessary.

> 
> > @argv-minus-one do we agree now that it's reasonable to have `get_array_elements[_critical]` borrow the array reference mutably as a way to help avoid aliasing - even though it's understood that it's not a water-tight solution?
> 
> It just occurred to me that requiring `&mut JPrimitiveArray` makes it impossible to use `JNIEnv::get_array_elements{,_critical}` if the array reference is a `GlobalRef`. Users who have a global reference to a Java primitive array will have to create a local reference from it before they can use `get_array_elements{,_critical}`. Is that acceptable? If so, should it be documented?
> 
> Other than that, yes.

hmmm, yeah that's kind of an annoying limitation.

trying to make `jni-rs` safer is a game of whack a mole :)

hmmm.

It bugs me that its so easy to create aliases without the mutable borrow but since it's anyway in the `# Safety` notes that this is not allowed maybe it's just simpler for now to go back to not taking the mutable borrow, and this can be something to try and revisit later.

rib:
> Is there a reason _not_ to use `catch_unwind`?

It looks like it makes sense here. Saying I was worried about "resorting to `catch_unwind`" was just a reflection of my default hesitation before using `catch_unwind` to check that its use is in keeping with Rust's model that panic unwinding is optional and not a general substitute for exceptions. For good reasons; it's not _that_ common that code should have to catch a panic while unwinding so in this case "resort" just means that I think it should pass a bit of extra scrutiny to double check that it's the most appropriate solution.

_(I was also trying to avoid getting side tracked thinking about the details of this just yet ;-p)_

rib:
Okey, I've rebased on master, squashed the doc fixes and dropped the patch that made `get_array_elements[_critical]` borrow the array mutably (and updated the CHANGELOG, MIGRATION guide and docs accordingly)

Hopefully this is ok to land now @argv-minus-one

src/wrapper/jnienv.rs
```diff
@@ -1738,7 +1745,11 @@ impl<'local> JNIEnv<'local> {
     }
 
     /// Converts a java byte array to a rust vector of bytes.
-    pub fn convert_byte_array(&self, array: jbyteArray) -> Result<Vec<u8>> {
+    pub fn convert_byte_array<'other_local>(
+        &self,
+        array: &impl AsRef<JByteArray<'other_local>>,
```
argv-minus-one, line 1750: This method, unlike the others, requires `&impl AsRef<JByteArray>` instead of just `impl AsRef<JByteArray>`. Is that intentional?

argv-minus-one:
I've pushed two commits to your branch:

* 1b56c74 adds a missing `unsafe` to the `JNIEnv::get_array_elements` example in the migration guide.
* de909b1 changes the changelog and migration guide to avoid implying that `AutoElements` can be used without `unsafe`, which is unfortunately no longer true.

I've also added a comment at `src/wrapper/jnienv.rs` line 1750.

Everything else looks good to me.

src/wrapper/jnienv.rs
rib: I don't think so - thanks for pointing out. I can tweak this and re-run CI before landing.

argv-minus-one (approved):

PR #404, opened 2023-01-25
Generalize with_local_frame closure return value

rib:
This updates with_local_frame to take a closure that may return any type of `Result`, so long as the error type implements
`From<jni::errors::Error>`

It also updates `Executor::with_attached_capacity` and `Executor::with_attached` (which are wrappers for `with_local_frame`) in the same way.

This removes the special-case ability to pass back a single local reference to the parent frame. (A new `with_local_frame_returning_local()` supports this case)

The main motivations here are:
- Consider that the closure is for user code that shouldn't be required to return a jni::errors::Result since the code may be doing a lot more than just calling a few JNI functions in crates that have their own Result types.
- Consider that it's often not necessary to return a local reference from a local frame and that feature is only really a modest optimization for one case that complicates other cases.

The fact that the closure was returning a `jni::errors::Result` before was convenient for in-tree tests but not convenient in crates that might wrap significant portions of code within a JNI local frame, and need to propagate crate-specific errors and arbitrary Ok() values.

One alternative that was considered was to make the closure return value completely generic and return that to the caller via a `jni::errors::Result`. This would typically imply returning nested `Result`s  - one outer `Result` for the Push/PopLocalFrame JNI calls, and then and inner `Result` for the closure.

The reason we chose this `E: From<jni::errors::Error>` approach is that it avoids the need to unwrap the return value twice in common cases.

The added use of generics may require callers to specify type parameters if they can't be inferred but it's also possible for applications to create their own ergonomic wrappers for crate-specific Result/Error types.

The documentation has been updated to highlight that it's possible to use a GlobalRef in situations you want to pass back an object reference to the calling scope.

This also adds a `with_local_frame_returning_local()` alternative where the closure returns a `std::result::Result<JObject, E>` (with the error being generic) that can return a local reference to the calling frame. Similar to `with_local_frame` the error must implement `From<jni::errors::Error>` which avoids the need to return nested `Result`s.

Micro benchmarks have been added to quantify the difference between:
- returning a `GlobalRef` from `with_local_frame` and then creating a local reference before dropping the global
- returning a local reference via `with_local_frame_returning_local`

On my machine I saw results like:
```
test tests::jni_with_local_frame_returning_global_to_local    ... bench:       2,104 ns/iter (+/- 181)
test tests::jni_with_local_frame_returning_local              ... bench:       1,669 ns/iter (+/- 99)
```
(So it was about %26 slower to return the global reference and create a local reference manually - but overall unlikely to be a notable difference for most code)

Co-authored-by: argv-minus-one <argv-minus-one@users.noreply.github.com>

### Definition of Done

Closes: #399

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ This release makes extensive breaking changes in order to improve safety. Most p
 - `WeakRef` and `JNIEnv#new_weak_ref`. ([#304](https://github.com/jni-rs/jni-rs/pull/304))
 - `define_class_bytearray` method that takes an `AutoArray<jbyte>` rather than a `&[u8]`
 - `JObject` now has an `as_raw` method that borrows the `JObject` instead of taking ownership like `into_raw`. Needed because `JObject` no longer has the `Copy` trait. ([#392](https://github.com/jni-rs/jni-rs/issues/392))
+- `JNIEnv::with_local_frame_returning_local` supports returning a single local ref from a new local frame, since this feature was removed from `with_local_frame` ([#399](https://github.com/jni-rs/jni-rs/issues/399))
 
 ### Changed
 - `JNIEnv::get_string` checks that the given object is a `java.lang.String` instance to avoid undefined behaviour from the JNI implementation potentially aborting the program. ([#328](https://github.com/jni-rs/jni-rs/issues/328))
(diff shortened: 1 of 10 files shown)


rib:
@argv-minus-one I hope you don't mind that I have squashed the changes you made while discussing #399 into my first patch (with a `Co-authored-by` attribution) and also made a further minor change to make `Executor::with_attached` also follow the same pattern.

I've also added CHANGELOG and MIGRATION notes

argv-minus-one (approved):
Yes, that's fine. Everything looks good to me.

Issue #391, opened 2022-11-20
Add convenience method for calling `DetroyJavaVM`


keastrid:
This is fairly simple to workaround, but having a convenient way to call it would be nice.

[Calling this method unloads the JVM](https://docs.oracle.com/en/java/javase/17/docs/specs/jni/invocation.html#unloading-the-vm), and functionally it is also important for GUI applications using AWT/Swing in that it waits for the application to finish before terminating. Without this call, such applications immediately shutdown after opening the main window.

rib:
Yeah this sort of seems like it should be done within a `Drop` implementation but even though `JavaVM` may load the vm via 
`JavaVM::new()` or `JavaVM::with_libjvm()` it's conceptually just a transparent pointer that doesn't really own anything.

(It's e.g. possible to create any number of `JavaVM`s via JNIEnv::get_java_vm`)

I guess the simplest (`unsafe`) approach could just be to directly expose an api like `JavaVM::destroy()` which would be very non-Rusty and it would just need to be documented that any attempt to access the vm afterwards would likely result in a crash.

Maybe it would be better if all the `invocation` functionality for loading a vm were actually moved out into some kind of `OwnedJavaVM` which could then implement `Drop`. There would still be the danger of something holding on to a `JavaVM` pointer though that would be an invalid pointer after the `OwnedJavaVM` gets dropped.

keastrid:
An `OwnedJavaVM` automatically handling this would be good, but even just a `JavaVM::destroy()` would be good.

rib:
Okey, I've taken a look at adding a `JavaVM::destroy()` method as discussed here.

Please see: https://github.com/jni-rs/jni-rs/pull/406

There are some significant safety issues related to the use of daemon threads that I've tried to describe in the `# Safety` section of the API, but with this addition it would at least be possible to try and unload a jvm if you're feeling lucky :-)

PR #406, opened 2023-01-25
Adds `JavaVM::destroy()` API

rib:
This adds an (unsafe) `JavaVM::destroy()` API that can be used to try and unload the JavaVM on supported platforms.

This adds a `dev-dependency` on the `rusty-fork` crate to make it easy to test `JavaVM::destroy()` in a separate process, which is needed to avoid breaking other tests.

It's notable that the way in which `jni-rs` uses an `InternalAttachGuard` to automatically detach threads makes it particularly difficult to use `JavaVM::destroy()` safely due to the fact that daemon threads will automatically try and access an invalid `JavaVM` pointer if they exit after `JavaVM::destroy()` has been called.

This also exposes `JavaVM::detach_current_thread` (it was previously public but hidden from docs) since that provides a mechanism for daemon threads to potentially detach themselves from a `JavaVM` explicitly if they need to be able to continue running after `JavaVM::destroy()` has been called.

Closes: #391

### Definition of Done

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,8 @@ This release makes extensive breaking changes in order to improve safety. Most p
 - `WeakRef` and `JNIEnv#new_weak_ref`. ([#304](https://github.com/jni-rs/jni-rs/pull/304))
 - `define_class_bytearray` method that takes an `AutoArray<jbyte>` rather than a `&[u8]`
 - `JObject` now has an `as_raw` method that borrows the `JObject` instead of taking ownership like `into_raw`. Needed because `JObject` no longer has the `Copy` trait. ([#392](https://github.com/jni-rs/jni-rs/issues/392))
-- `JNIEnv::with_local_frame_returning_local` supports returning a single local ref from a new local frame, since this feature was removed from `with_local_frame` ([#399](https://github.com/jni-rs/jni-rs/issues/399))
+- `JavaVM::destroy()` (unsafe) as a way to try and unload a `JavaVM` on supported platforms ([#391](https://github.com/jni-rs/jni-rs/issues/391))
+- `JavaVM::detach_current_thread()` (unsafe) as a way to explicitly detach a thread (normally this is automatic on thread exit). Needed to detach daemon threads manually if using `JavaVM::destroy()` ([#391](https://github.com/jni-rs/jni-rs/issues/391))
 
 ### Changed
 - `JNIEnv::get_string` checks that the given object is a `java.lang.String` instance to avoid undefined behaviour from the JNI implementation potentially aborting the program. ([#328](https://github.com/jni-rs/jni-rs/issues/328))
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,7 +30,7 @@ walkdir = "2"
 
 [dev-dependencies]
 lazy_static = "1"
-
+rusty-fork = "0.3.0"
 
 [features]
 invocation = ["java-locator", "libloading"]
--- a/src/wrapper/java_vm/vm.rs
+++ b/src/wrapper/java_vm/vm.rs
@@ -286,11 +286,30 @@ impl JavaVM {
         }
     }
 
-    /// Detaches current thread from the JVM. This operation is _rarely_ appropriate to use,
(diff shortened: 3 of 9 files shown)


rib:
@keastrid I wonder if you might be able to take a quick look at this if you get a chance, since it sounded like you were interested in having an API for this

argv-minus-one:
The documentation says the `JavaVM` must be dropped promptly after a successful return from `JavaVM::destroy`. This should also apply to any remaining `JNIEnv` belonging to the same JVM, and to anything that contains a `JNIEnv` and/or `JavaVM` belonging to the same JVM, namely:

* `AttachGuard`
* `AutoElements`
* `AutoElementsCritical`
* `AutoLocal`
* `GlobalRef`
* `JavaStr`
* `JMap`
* `WeakRef`

(Hopefully I didn't miss any. :sweat_smile:)

Several of these also perform a JNI call on drop, so it's especially important that they be dropped before `JavaVM::destroy`, or at least that their destructors don't run afterward.

The above implies that there must not be any calls in progress on or involving a `JNIEnv` or `JavaVM` belonging to the same JVM, such as `JNIEnv::with_local_frame`.

Also, the JNI spec doesn't say what will happen if `DestroyJavaVM` is called from inside a Java native method call. My guess is this will cause undefined behavior as soon as it tries to return to Java, since there is no longer any Java to return to, but that's only a guess. Not sure if this should be documented or not.

rib:
Yeah about all the other types that also need to be dropped ahead of time maybe I can find a different way of making it clear that all this state needs to be dropped / cleared up before trying to delete the `JavaVM`. The only reason I had the special note about dropping the JavaVM itself was just because of the oddity where the `destroy` API can't take ownership of the `JavaVM` and cause the drop internally and so I wanted to make sure that the caller would be responsible for that final drop (I suppose I was was assuming implicitly that everything else would have been dropped)

It's quite a notable thing to clarify though because it's very easy to not realize you have something like an `AutoLocal`/`GlobalRef` in scope which might end up getting dropped after calling `destroy()` (I actually hit this myself when writting the test and initially got a segfault as a result)

> Also, the JNI spec doesn't say what will happen if DestroyJavaVM is called from inside a Java native method call

I have a feeling I saw a recent spec clarification about this which effectively said "no you definitely can't do that" but will have to poke around to try and dig up where I saw that.

rib:
> > Also, the JNI spec doesn't say what will happen if DestroyJavaVM is called from inside a Java native method call
> 
> I have a feeling I saw a recent spec clarification about this which effectively said "no you definitely can't do that" but will have to poke around to try and dig up where I saw that.

see https://bugs.openjdk.org/browse/JDK-8292897 and https://bugs.openjdk.org/browse/JDK-8290482 and http://cr.openjdk.java.net/~dholmes/8290482/8290482-jni-spec/diff.html

keastrid:
This looks good!

Not sure how to handle the safety issues.

rib:
> This looks good!
> 
> Not sure how to handle the safety issues.

Certainly in the short term, with the current design of jni-rs I don't think there will be a way to make `::destroy()` safe, so it's more just for completeness I feel.

Tbh it's hard to imagine it ever being safe. It looks like it's not a very well specified API and also not necessarily supported by all JVMs either.

The way in which it doesn't wait for daemon threads actually seems like a fairly terrible design to me - that specifically means you can destroy a JVM while there _are_ threads attached to the JVM with no details specified about what happens to any of the attachment state for those daemon threads.

It looks like Openjdk/hotspot doesn't actually support unloading the JVM, I guess in part because it would just be too dangerous to actually free everything and leave dangling JNI pointers.

For a safer design I could imagine removing / hiding the concept of "daemon" threads which just seem like a recipe for use after free to me. There would probably also need to be clearer ownership relationships between the `JavaVM` and all the other JNI resources which could be hard to achieve while all these things are just transparent wrappers over JNI pointers.

rib:
For reference, the latest iteration updated the docs, considering the comments from @argv-minus-one. The safety notes now specifically make it clear it can't be called with any Java methods on the call stack, and makes it clearer that its your responsibility to release JNI resources beforehand that could lead to any use of the VM after it's destroyed (including the above list of auto-release types that would result in automatic JNI usage when dropped)

PR #414, opened 2023-02-03
Correctly encode JVM option strings

### LINKED ISSUE #410: Invocation API uses wrong encoding for options
The JVM Invocation API requires options to be in the platform default encoding, while Rust's CString always uses UTF-8 as far as I can tell (unless the code is really non-obvious). While this is (usually) correct on Linux and macOS, it isn't at all on Windows.
`local_encoding_ng` might help here.

argv-minus-one:
## Overview

This draft PR fixes #410. It adds code to transcode JVM option strings, as passed to `InitArgsBuilder::option`, into the character encoding that HotSpot expects. It also adds `InitArgsBuilder::option_encoded` to skip transcoding and pass a raw `CStr` to the JVM.

This is a draft because it's a work in progress. Only Windows is handled so far. Next I will see about doing the same for POSIX platforms.

### Definition of Done

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ This release makes extensive breaking changes in order to improve safety. Most p
 - `JObjectArray` provides a reference wrapper for a `jobjectArray` with a lifetime like `JObject`. ([#400](https://github.com/jni-rs/jni-rs/pull/400))
 - `AutoElements` and `AutoElementsCritical` (previously `AutoArray`/`AutoPrimitiveArray`) implement `Deref<Target=[T]>` and `DerefMut` so array elements can be accessed via slices without needing additional `unsafe` code. ([#400](https://github.com/jni-rs/jni-rs/pull/400))
 - `AsJArrayRaw` trait which enables `JNIEnv::get_array_length()` to work with `JPrimitiveArray` or `JObjectArray` types ([#400](https://github.com/jni-rs/jni-rs/pull/400))
+- `InitArgsBuilder` now has `try_option` and `option_encoded` methods. ([#414](https://github.com/jni-rs/jni-rs/pull/414))
 
 ### Changed
 - `JNIEnv::get_string` checks that the given object is a `java.lang.String` instance to avoid undefined behaviour from the JNI implementation potentially aborting the program. ([#328](https://github.com/jni-rs/jni-rs/issues/328))
@@ -62,11 +63,13 @@ This release makes extensive breaking changes in order to improve safety. Most p
 - `get_array_elements` is now also `unsafe` (for many of the same reasons as `get_array_elements_critical`) and has detailed safety documentation ([#400](https://github.com/jni-rs/jni-rs/pull/400))
 - `AutoArray/AutoArrayCritical::size()` has been replaced with `.len()` which can't fail and returns a `usize` ([#400](https://github.com/jni-rs/jni-rs/pull/400))
 - The `TypeArray` trait is now a private / sealed trait, that is considered to be an implementation detail for the `AutoArray` API.
-
+- `JvmError` has several more variants and is now `non_exhaustive`. ([#414](https://github.com/jni-rs/jni-rs/pull/414))
+- `InitArgsBuilder::option` raises an error on Windows if the string is too long. The limit is currently 1048576 bytes. ([#414](https://github.com/jni-rs/jni-rs/pull/414))
 
 ### Fixed
 - Trying to use an object reference after it has been deleted now causes a compile error instead of undefined behavior. As a result, it is now safe to use `AutoLocal`, `JNIEnv::delete_local_ref`, and `JNIEnv::with_local_frame`. (Most of the limitations added in #392, listed above, were needed to make this work.) ([#381](https://github.com/jni-rs/jni-rs/issues/381), [#392](https://github.com/jni-rs/jni-rs/issues/392))
 - Class lookups via the `Desc` trait now return `AutoLocal`s, which prevents them from leaking. ([#109](https://github.com/jni-rs/jni-rs/issues/109), [#392](https://github.com/jni-rs/jni-rs/issues/392))
+- `InitArgsBuilder::option` properly encodes non-ASCII characters on Windows. ([#414](https://github.com/jni-rs/jni-rs/pull/414))
 
 ### Removed
 - `get_string_utf_chars` and `release_string_utf_chars` from `JNIEnv` (See `JavaStr::into_raw()` and `JavaStr::from_raw()` instead) ([#372](https://github.com/jni-rs/jni-rs/pull/372))
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,7 @@ version = "0.21.0-alpha.0"
 edition = "2018"
 
 [dependencies]
+cfg-if = "1.0.0"
 cesu8 = "1.1.0"
 combine = "4.1.0"
 java-locator = { version = "0.1", optional = true }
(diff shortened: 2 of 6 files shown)


rib:
Thanks for looking at this.

Please see my comment here: https://github.com/jni-rs/jni-rs/issues/410#issuecomment-1415945899 where I try to clarify my concerns around something similar for POSIX.

Although I was roughly expecting that the Windows support was going to be a lot simpler than this I can still see that there's real-world value in adding some character mapping support on Windows.

I really can't say the same for POSIX - I can't see how we can really justify adding code, complexity and dependence on legacy character maps etc when there's no longer any reason for POSIX systems to use those legacy non-utf8 locales - we'd be adding code and complexity that we'll have to maintain but no one is going to use - which isn't the same situation as with Windows.

argv-minus-one:
> Although I was roughly expecting that the Windows support was going to be a lot simpler than this

I was too, but Windows' unfortunate choice to use `int` instead of `SIZE_T` as the string length type created the possibility of overflow, which added a bunch of complexity that I wasn't expecting. I decided against using local-encoding-ng because it doesn't check for or handle overflow.

Right now, this code splits the input string into chunks small enough that they shouldn't overflow, with the maximum chunk length being just small enough to avoid overflow if the code page is UTF-7 (which I believe is the least-byte-efficient character encoding Windows supports).

I could simplify this code by only checking for overflow and reporting an error if there is one. This would only save about 15 lines of code, though.

> I really can't say the same for POSIX - I can't see how we can really justify adding code, complexity and dependence on legacy character maps etc when there's no longer any reason for POSIX systems to use those legacy non-utf8 locales - we'd be adding code and complexity that we'll have to maintain but no one is going to use - which isn't the same situation as with Windows.

OK, I'll leave POSIX assuming UTF-8 like it does now.

rib:
> > Although I was roughly expecting that the Windows support was going to be a lot simpler than this
> 
> I was too, but Windows' unfortunate choice to use `int` instead of `SIZE_T` as the string length type created the possibility of overflow, which added a bunch of complexity that I wasn't expecting. I decided against using local-encoding-ng because it doesn't check for or handle overflow.

I had roughly thought it was going to be possible to just go from Rust uft8 -> utf16 and then there was a Windows API that took a complete utf16 string and could return an 8-bit encoded string.

Having to faff around manually checking for utf16 surrogate pairs was very surprising to see here.

Why can't we either:
1. pass `cchWideChar` as -1 after converting the utf16 string to a `CString` (that will ensure it's NUL terminated)
2. query the utf16 length in Rust and throw our own error if it's greater than `i32::MAX` before passing the length to `WideCharToMultiByte`

I don't think efficiency should be concern at all if it means we can have a drastically simpler solution.

> OK, I'll leave POSIX assuming UTF-8 like it does now.

ah, okey, _phew_

rib:
It looks like it could be helpful to utilize this [widestring](https://docs.rs/widestring/1.0.2/widestring/) crate (licensing is compatible and it looks like a well considered, well maintained crate with some utilities that could be handy here)

In particular this utf16 equivalent of `CString` looks useful: https://docs.rs/widestring/1.0.2/widestring/ucstring/struct.U16CString.html

argv-minus-one:
> > OK, I'll leave POSIX assuming UTF-8 like it does now.
> 
> ah, okey, _phew_

Sorry, didn't mean to worry you. 😅

> I had roughly thought it was going to be possible to just go from Rust uft8 -> utf16 and then there was a Windows API that took a complete utf16 string and could return an 8-bit encoded string.

Well, yes, there is, but it has overflow problems.

> Having to faff around manually checking for utf16 surrogate pairs was very surprising to see here.

That's a consequence of splitting the string in order to avoid overflow. It's safe to split a UTF-16 string to transcode it in chunks, if and *only* if it isn't split in the middle of a surrogate pair.

The widestring crate doesn't check for this, by the way. According to [an example in the documentation](https://docs.rs/widestring/1.0.2/widestring/ustring/struct.U16String.html#examples), it considers unpaired surrogates valid.

> Why can't we either:
> 
> 1. pass `cchWideChar` as -1 after converting the utf16 string to a `CString` (that will ensure it's NUL terminated)

[Here it is](https://github.com/jni-rs/jni-rs/pull/414/commits/2b3b9f102f3f55a091c558e46aecf22234261131) (I left in some debug statements for now), but this breaks on overflow. Turns out that Win32 `WideCharToMultiByte` doesn't check for overflow if the input string is more than `i32::MAX` UTF-16 units long, and behaves similarly to local-encoding-ng.

I've added [a test that checks what happens when the string length is around `i32::MAX`](https://github.com/jni-rs/jni-rs/blob/2b3b9f102f3f55a091c558e46aecf22234261131/src/wrapper/java_vm/init_args/char_encoding_windows.rs#L217), and it fails [here](https://github.com/jni-rs/jni-rs/blob/2b3b9f102f3f55a091c558e46aecf22234261131/src/wrapper/java_vm/init_args/char_encoding_windows.rs#L317). `WideCharToMultiByte` returns zero, which normally indicates error, but in this case it's the result of its string length counter wrapping around, resulting in an amusing error message of “The operation completed successfully.”

Conclusion: feeding `WideCharToMultiByte` a null-terminated string is not overflow-safe.

> 2. query the utf16 length in Rust and throw our own error if it's greater than `i32::MAX` before passing the length to `WideCharToMultiByte`

[Here it is](https://github.com/jni-rs/jni-rs/pull/414/commits/c501e88d73687d7d465f83fdd6de9f09b1ad661f). This sort of works, but `WideCharToMultiByte` raises `ERROR_INVALID_PARAMETER` (instead of `ERROR_INSUFFICIENT_BUFFER`) if the input UTF-16 string isn't too large (≤ `i32::MAX` UTF-16 units) but the output string is (> `i32::MAX` bytes).

Maybe that's okay. The only Windows code page I know of where this could happen is UTF-7, and I seriously doubt anyone will ever set that as their default code page. (UTF-8 can also exceed the limit this way, as demonstrated by the test, but [transcoding is skipped if the code page is UTF-8](https://github.com/jni-rs/jni-rs/blob/c501e88d73687d7d465f83fdd6de9f09b1ad661f/src/wrapper/java_vm/init_args/char_encoding_windows.rs#L215), so it doesn't matter.)

That test code sure is ugly, though, and it takes 5 minutes and around 12GB of memory to run. `WideCharToMultiByte` apparently allocates a lot of memory when transcoding a huge UTF-16 string to Windows-1252. It's much more efficient transcoding UTF-16 to UTF-8, interestingly.

So, what do you think? Here are the options I can think of:

1. Check for overflow and raise an error if it happens, like in c501e88.
2. Avoid overflow by splitting the string, like in 8809b61.
3. Don't even bother checking for or handling overflow at all.

Option 3 seems to be the most popular approach. That's what local-encoding-ng and `WideCharToMultiByte` do. I usually don't like ignoring potential errors like that, but if that's what you think is best, then I'll do it.

I'm fond of option 2 because it makes testing simple: just change the chunk size to something small and make sure the string gets split and reassembled correctly.

I don't like option 1. The implementation is a little simpler, but the test is horrible. My computer almost melted running it! 😅 Maybe it can be simplified somehow, but I'm not sure how.

rib:
> I've added [a test that checks what happens when the string length is around i32::MAX](https://github.com/jni-rs/jni-rs/blob/2b3b9f102f3f55a091c558e46aecf22234261131/src/wrapper/java_vm/init_args/char_encoding_windows.rs#L217)

Hehe, kudos. I was wondering if you were going to end up doing that. To be fair I was assuming the implementation was going to detect the overflow fine, so interesting to see that it doesn't.

I think the right solution here is to just to set our own option size limit and just check for that. E.g. 8k would be a ludicrous size surely for an option value. If we adapt your test for that it will also be able to run in a much more reasonable time too.

argv-minus-one:
Okay, I've pushed https://github.com/jni-rs/jni-rs/commit/ee109d3ec1ab98f2fcae7bf95f4afc394f16d0b0 where it limits the string's length to what should very definitely never overflow, even with UTF-7. That's still 858,993,458 bytes of UTF-8, so hopefully no one will complain. 😅

I've kept the big overflow test. I figured we'd better keep it so we know for sure that it really doesn't overflow. It's not so computer-melting any more; it only requires about 4GB of memory and runs in one minute instead of five. Still needs to be run in a `--release` build, though; it relies very heavily on the optimizer to fill those giant buffers efficiently.

Or we could just forget about this and limit to some arbitrary length like 8k. Your choice, but I thought I'd better present this option too.

argv-minus-one:
Okay, all set. I've set the limit to 1MB so the overflow test runs quickly even without `--release`. I've also added changelog entries and added to the documentation. Should be ready to merge.

rib:
looks good, thanks!

Issue #426, opened 2023-03-03
Example request: `call_method` with non-primitive type as parameters.


Renkai:
Hi, community!
We have great examples of how to call a java method in rust if all the parameters a primitive values(int, double, etc...)
But I don't see an example with java objects as parameters. JNI should support it, right?
The below code is what I tried so far(compile error).
Any help is welcomed  😄
```
// use arrow::ffi::{FFI_ArrowArray, FFI_ArrowSchema};
use jni::JNIEnv;

// These objects are what you should use as arguments to your native
// function. They carry extra lifetime information to prevent them escaping
// this context and getting used after being GC'd.
use jni::objects::{JClass, JLongArray, JObject, JString, JValue, JValueGen};

// This is just a pointer. We'll be returning it from our function. We
// can't return one of the objects with lifetime information because the
// lifetime checker won't let us.
use jni::sys::{jclass, jlong, jlongArray, jobjectArray, jstring};

#[no_mangle]
pub extern "system" fn Java_lance_JNI_saveToLance<'local>(
    mut env: JNIEnv<'local>,
    _class: JClass<'local>,
    path: JString<'local>,
    vec: JObject<'local>,
    allocator: JObject<'local>) {
    let path: String = env
        .get_string(&path)
        .expect("Couldn't get java string!")
        .into();
    // let schema = FFI_ArrowSchema::empty();
    // let array = FFI_ArrowArray::empty();
    // 
    // let schema = Box::new(schema);
    // let array = Box::new(array);
    // 
    // let schema_ptr = &*schema as *const FFI_ArrowSchema;
    // let array_ptr = &*array as *const FFI_ArrowArray;
    // 
    // let schema_jlong = schema_ptr as i64;
    // let array_jlong = schema_ptr as i64;
    let schema_jlong = 0i64;
    let arrary_jlong = 0i64;
    let vecO = JValue::from(vec);
    let allocatorO = JValue::from(allocator);

    env.call_static_method(_class, "fillVector", "(I)V", &[schema_jlong.into(), array_jlong.into(), vecO, allocatorO]);

    println!("jenv {:?}", env);
    println!("class {:?}", _class);
    println!("path {:?}", path);
    println!("vec {:?}", vec);
    println!("allocator {:?}", allocator)
}
```

rib:
Yeah passing object references as arguments is supported (e.g. via `JObject`, `JString`, `JClass` or `JPrimitiveArray` types that all wrap a reference to an object)

[...]

PR #429, opened 2023-03-07
Remove `JValueGen`; degenerify `JValue` and `JValueOwned`

argv-minus-one:
## Overview

As discussed in https://github.com/jni-rs/jni-rs/issues/426#issuecomment-1458343035, this PR removes `JValueGen`. `JValue` and `JValueOwned` are now separate, unrelated, non-generic types.

Although this introduces some code duplication, it should result in better compiler error messages when you mistakenly use one when the other is expected.

### Definition of Done

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Changed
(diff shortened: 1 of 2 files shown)


rib:
Yeah, maybe going for the separated types will be the more ergonomic trade off for users of the API. It doesn't seem like it will be too bad of a maintenance burden dealing with this amount of duplication, so if it makes the public API a little simpler it's probably worth it.

Thanks for cooking up a patch for this.

Lets just hold off merging breaking changes for a little while incase any other issues come up that require another patch release.

rib (approved):

PR #472, opened 2023-07-14
Use the  "system-unwind" ABI in examples

koutheir:
Given that the ABI "system-unwind" is [now in stable Rust](https://github.com/rust-lang/rust/issues/74990#issuecomment-1363473645), it should be the new recommended way to define JNI functions in Rust, in order to avoid undefined behavior when an unwind happens, due to a C++ exception, or a Rust panic, etc.

### Definition of Done

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -113,7 +113,7 @@
 //! // This keeps Rust from "mangling" the name and making it unique for this
 //! // crate.
 //! #[no_mangle]
-//! pub extern "system" fn Java_HelloWorld_hello<'local>(mut env: JNIEnv<'local>,
+//! pub extern "system-unwind" fn Java_HelloWorld_hello<'local>(mut env: JNIEnv<'local>,
 //! // This is the class that owns our static method. It's not going to be used,
 //! // but still must be present to match the expected signature of a static
 //! // native method.


rib:
I'm not sure about this.

The prefix for these unwind ABIs define how Rust can continue unwinding past a Rust function, over an ffi boundary into another language's stack ABI.

Logically (from the pov of developers binding Java + Rust code together) these native methods are being called from code running _on_ a JVM. The stack that we'd really be interested in unwinding into lives at higher level of abstraction than I think any of these unwind ABIs will know about.

Technically there is going to have to be a small amount of shim system code but logically I'm not sure we will really want to let Rust unwind into the implementation details of the JVM.

Ideally it seems like we'd want a "java-unwind" that could somehow really follow the stack frames in Java and potentially even pop-out the other side back into Rust code, but I don't imagine that's possible currently.

Have you tried panicking with this change to see what kind of backtrace you get when e.g. running with openjdk? Maybe in some cases, with some JVMs the system backtrace could give some clues about what java code might have been running at that time, but I'd guess you might just see some awkward symbols that represent implementation details of the JVM.

For most code I'd guess that mapping Rust errors or panics into Java exceptions is the best way to effectively unwind across these boundaries.

argv-minus-one:
As far as I know, the JNI specification doesn't say it's safe to throw a C++ exception from a function called by the JVM (i.e. a `native` method implementation), so the result of unwinding into the JVM is unspecified at best. Rust may support it now, but that doesn't mean the JVM does.

HotSpot will probably cleanly abort if it catches a Rust panic, since it's written in C++, but what if you're using some other JVM that's written in C and compiled with a compiler that doesn't know anything about C++ exceptions? What will happen then? Nobody ever said the JVM is guaranteed to be written in C++.

So, no, I don't think telling people to use `C-unwind` with this crate is a good idea.

PR #490, opened 2023-08-30
link the Maven Rust plugin

jerrinot:
I believe this plugin can be useful for jni-rs crate users. 

Full Disclosure: I work for QuestDB - the company that maintains the plugin.

PS: We also wrote about our experience from integrating Java and Rust. I believe it is also very useful, but I wasn't sure whether it's something you would want to link from the crate docs: https://questdb.io/blog/leveraging-rust-in-our-high-performance-java-database/


rib (approved):
Thanks for the pointer, as well as to your blog post, this looks good to me

PR #498, opened 2023-09-26
Display Java exception messages in `Error::JavaException` error message

### LINKED ISSUE #496: `Error::JavaException`, when `Display`ed, doesn't show the Java exception message
[Some people on Discord got confused](https://meta.jorun.dev/notice/AY214MjkpCsMPe5Vmi) when a `Display`ed `Error::JavaException` showed only the words `Java exception` and the Java method that threw it, and not the Java exception's class or message.

They also got confused when this happened on what appears to be a Macintosh 128k, except it has a color display and runs what appears to be Windows XP, with the taskbar at the top of the screen. 🤨

Anyway, when I have some time, I'll see if I can change `<Error as Display>::fmt` to try to get the currently pending Java exception and display the exception message. Obviously that is not guaranteed to succeed, or to display the *correct* exception's message, but it's better than nothing.

argv-minus-one:
## Overview

(Currently a draft PR because this is based on #478, which isn't yet merged.)

This PR fixes #496 by capturing information about Java exceptions in `Error::JavaException`, and displaying that information as part of the error message.

It accomplishes this by storing a `GlobalRef` to the exception object, then calling its `toString()` method (or, if that also throws an exception, `getClass().getName()`) to produce an exception message.

## Other changes

In addition to the above, this PR makes two other changes, in separate commits.

This PR adds the method `AttachGuard::was_already_attached`, which returns true iff the thread was already attached when `JavaVM::attach_current_thread` was called. This information is used to decide whether to log a warning about the thread not being attached when the `impl Display for JavaException` is invoked. The `GlobalRef` destructor is changed to use this new method as well.

This PR makes `JNIEnv::exception_occurred` infallibly return `Option<JThrowable>`. It wasn't made infallible in #478 like the other exception-checking `JNIEnv` methods, and my exception-capturing code needed it to be infallible.

## Breaking changes

`Error::JavaException` is now a tuple variant instead of a unit variant, which is a breaking change.

## Open questions

Should we do this at all? On my machine, this adds an overhead of at least 2µs every time a `JNIEnv` call (other than `throw` or `throw_new`) results in a Java exception being thrown.

Should the exception message be captured eagerly or lazily? This code is currently lazy: it creates a `GlobalRef` to the exception object, and only calls its `toString()` if `Display::fmt` is called. It could be changed to eagerly call `toString()` when the exception is thrown instead. Here's how they compare:

* Lazy mode saves 2µs if `Display::fmt` is called, but costs 2µs if `Display::fmt` is *not* called (i.e. the exception is thrown, but the exception message is never displayed, at least not by Rust code).
* Lazy mode costs an additional 86µs if `Display::fmt` and `Drop::drop` are called on a thread that's not currently attached to the JVM. This is the same problem as dropping a `GlobalRef` on an unattached thread, but doubled: once for `Display::fmt`, once for `Drop::drop`.

Since I expect most Java exceptions will never be displayed by Rust code, only passed back to Java, lazy mode seems like the best approach.

Should users be allowed to suppress exception capture? This removes the benefit of seeing exception messages, but reduces overhead to almost zero. If the answer to this question is yes, should it be by a function call, an environment variable, or both?

Am I overthinking this? Wouldn't be the first time I overthought a performance edge case… 😅

### Benchmark results

```
test tests::exception_capture_eager_nostringify_nosuppress    ... bench:       4,097 ns/iter (+/- 60)
test tests::exception_capture_eager_nostringify_suppress      ... bench:          20 ns/iter (+/- 0)
test tests::exception_capture_eager_stringify_nosuppress      ... bench:       4,135 ns/iter (+/- 516)
test tests::exception_capture_eager_stringify_suppress        ... bench:          48 ns/iter (+/- 3)
test tests::exception_capture_lazy_nostringify_nosuppress     ... bench:       2,195 ns/iter (+/- 117)
test tests::exception_capture_lazy_nostringify_suppress       ... bench:          20 ns/iter (+/- 0)
test tests::exception_capture_lazy_stringify_nosuppress       ... bench:       6,410 ns/iter (+/- 116)
test tests::exception_capture_lazy_stringify_suppress         ... bench:          48 ns/iter (+/- 1)
test tests::exception_capture_send_baseline                   ... bench:      29,852 ns/iter (+/- 10,037)
test tests::exception_capture_send_withattach                 ... bench:     116,000 ns/iter (+/- 26,477)
```

Benchmark function names contain the following keywords:

* `eager` vs `lazy`:
  * `eager`: Calls the exception's `toString()` immediately when it is thrown.
  * `lazy`: Calls the exception's `toString()` only if `Display::fmt` is called.
* `stringify` vs `nostringify`:
  * `stringify`: `Display::fmt` is called.
  * `nostringify`: The `JavaException` is silently discarded without ever calling `Display::fmt`.
* `suppress` vs `nosuppress`:
  * `suppress`: `toString()` is not called and `GlobalRef` is not created. `Display::fmt` produces a dummy exception message instead of the real one.
  * `nosuppress`: `toString()` and/or `GlobalRef` is called/created. `Display::fmt` produces the real exception message.
* `send_baseline` and `send_withattach`:
  * `send_baseline`: Sends the `JavaException` to another thread using an MPSC channel, receives it back with another MPSC channel, and then stringifies it as in `lazy_stringify_nosuppress`. This benchmark is the baseline for `send_withattach`.
  * `send_withattach`: Sends the `JavaException` to another thread using an MPSC channel, which stringifies and drops it, then sends back the resulting string using another MPSC channel. This demonstrates the severe overhead of stringifying and dropping a `JavaException` on an unattached thread.

## Definition of Done


src/wrapper/macros.rs
```diff
@@ -25,7 +25,7 @@ macro_rules! jni_call_check_ex {
     ( $jnienv:expr, $version:tt, $name:tt $(, $args:expr )* ) => ({
         let ret = jni_call_unchecked!($jnienv, $version, $name $(, $args)*);
         if $jnienv.exception_check() {
-            Err(crate::errors::Error::JavaException)
```
rib, line 28: I'm not confident this is the way to go...

It feels a bit funky to be automatically clearing + re-throwing all exceptions for the sake of creating a global ref that may or may not be used afterwards.

Should we instead be providing utilities that make it ergonomic to capture an exception explicitly, but leave it up to `jni` users to decide if they want to do that from Rust or else let them decide to return back to the JVM so it will be thrown within the JVM instead?

For example in the android-activity crate I have a utility that maps a `JavaException` into an Error: https://github.com/rust-mobile/android-activity/blob/967882f3d904d9bdc660dee96855c874a8968ce3/android-activity/src/jni_utils.rs#L57

src/wrapper/jnienv.rs
```diff
@@ -591,9 +591,13 @@ impl<'local> JNIEnv<'local> {
...
     /// not caught in a java function until `exception_clear` is called.
-    pub fn exception_occurred(&mut self) -> Result<JThrowable<'local>> {
+    pub fn exception_occurred(&mut self) -> Option<JThrowable<'local>> {
         let throwable = unsafe { jni_call_unchecked!(self, v1_1, ExceptionOccurred) };
-        Ok(unsafe { JThrowable::from_raw(throwable) })
+        if throwable.is_null() {
+            None
+        } else {
+            Some(unsafe { JThrowable::from_raw(throwable) })
+        }
```
rib, line 600: Yeah, it could be good to pull this out into a separate PR, since this definitely looks good to land to me.

src/wrapper/objects/global_ref.rs
```diff
@@ -165,25 +170,19 @@ impl Drop for GlobalRefGuard {
     fn drop(&mut self) {
         let raw: sys::jobject = mem::take(&mut self.obj).into_raw();
 
-        let drop_impl = |env: &JNIEnv| -> Result<()> {
+        let res: Result<()> = (|| {
+            let env = self.vm.attach_current_thread()?;
```
rib, line 174: My slight reservation here is that I'm half expecting that when we look more closely at the safety of the attachment APIs, we're going to conclude that we should remove the RAII/guard based attachments, at least until we can create a proper singleton crate that can guarantee we only have a single thread-local for tracking the attachments. (currently multiple versions of the jni crate or mixing with alternative Rust jni crates could lead to unreliable attachment checks)

The end result is more readable here but it looks like the previous code has the same behaviour effectively, except it relies on the jni library to do the thread-local look up. Is there maybe a significant performance difference here?

src/wrapper/jnienv.rs
rib: I ended up cherry picking and merging this via https://github.com/jni-rs/jni-rs/pull/517

rib:
For reference here; I cherry picked the change to make `exception_occurred` infallible and merged that via https://github.com/jni-rs/jni-rs/pull/517 while I'm not currently 100% sure about some of the changes here.

src/wrapper/macros.rs
argv-minus-one: As I recall, the reason I did this was to guarantee that the *correct* exception is displayed. As far as I know, the only way to do that is to catch it immediately when it's thrown, and retain a reference to it in case it is displayed later.

If the exception is caught only when it's to be displayed, then it's possible that the exception was already caught, cleared, and some *other* exception (or none at all) is now pending instead.

But yeah, I was wondering (in the original PR comment) whether such a guarantee is worth the performance overhead and code complexity. I take it you're thinking not?

src/wrapper/objects/global_ref.rs
argv-minus-one: Yes, it has the same behavior as before.

As you know, the `GlobalRef` destructor has to delete the global reference on a potentially unattached thread, so it tries attaching the current thread and, if the current thread is not already attached, temporarily attaches it and logs a warning.

Getting a Java exception message from a `Display` implementation has the same problem, so I refactored the attach-with-warning code into its own function, which is called from the `GlobalRef` destructor as well as from `impl Display for JavaException` [here](https://github.com/jni-rs/jni-rs/blob/138006d14a8cac865b4a8d40130b03e8e3422fde/src/wrapper/errors/exception_capture.rs#L244).

> I'm half expecting that when we look more closely at the safety of the attachment APIs, we're going to conclude that we should remove the RAII/guard based attachments, at least until we can create a proper singleton crate that can guarantee we only have a single thread-local for tracking the attachments.

Probably. The safety issue I'm aware of is that two `AttachGuards` could be acquired at the same time, and then the first one could be dropped before the second one is dropped, resulting in an invalid `JNIEnv`. This could be solved with a thread-local variable that counts how many times the thread has been attached, so that it is only actually detached when the counter drops to zero, but yeah, there has to be exactly one such counter that everyone uses. Pity the JVM doesn't do that…

I don't think all of that applies here, though, because this function only acquires a single `AttachGuard`, and doesn't call anything that could potentially acquire a second one.

brandonros:
Any update on this?

PR #508, opened 2023-11-08
[proposed feature] Allow AutoElements to store an owned local reference

jrose-signal:
## Overview

Currently, AutoElements borrows a local reference to a JPrimitiveArray. However, this makes it harder to pass the AutoElements around and make collections of them, since they depend not only on the context but on the local variable that held the array they were borrowed from. This PR changes AutoElements to depend on anything that can AsRef to a JPrimitiveArray, which in practice means either a reference or an owned local reference, or even an AutoLocal.

Another option would be for AutoElements to *always* create a new local reference, but that could break programs that aren't expecting to allocate more local references just by using AutoElements.

This is a breaking change, since it changes the generic signature of AutoElements.

Is this useful enough to take?

### Definition of Done

- [ ] **Change is covered by [automated tests](https://github.com/jni-rs/jni-rs/blob/master/CONTRIBUTING.md#tests)**: I manually messed with the existing AutoElements test to make sure this worked, but I don't want to add a ton of duplication to the tests.
- [ ] **Public API has documentation**: Haven't updated this yet.
- [ ] **User-visible changes are mentioned in the Changelog**: Or this.


jrose-signal:
This is not the same thing as #302 but it might be enough of a usability improvement anyway.

jrose-signal:
Ran into a need for this again today, can rebase if there's interest.

argv-minus-one:
Seems reasonable to me. @rib, how do you feel about this?

rib:
Sorry for not being active in the last year or so. This also seems good to me. I'm hoping to take a pass at some of the safety issues around thread attachments which seem like they really need to be addressed before a release and can also look at rebasing this if you've moved on to other things @jrose-signal.

jrose-signal:
Yeah, it's been a bit too long, please take over!

rib:
Okey, I ended up re-implementing this from scratch (while quite a lot has changed since #570) but I added a ` Co-authored-by` line while I borrowed bits from your original change @jrose-signal 

In the end, a number of simplifications were possible:
- `AutoElements` is now only parameterized by one `'array_local` lifetime that represents the JNI stack frame that owns the array reference. (Considering AutoElements needs to keep the reference to be able to Release it later).
- Since `JNIEnv::get_array_elements` and `AutoElements::new` don't need to create a new local reference (the pointer that we `Get` from JNI isn't a local reference that's tied to the top JNI stack frame) that means we didn't actually need a mutable JNIEnv reference and we don't need to be tied to the `JNIEnv` lifetime.

Maybe this is enough to even say that #302 has been fully addressed?

I haven't looked at `AutoElementsCritical` but I guess some similar simplifications should be possible there too.

rib:
I've tweaked the tests to redundantly call `get_array_elements` from a new JNI frame created using `env.with_local_frame()` and then returning the `AutoElements` back to the previous frame, to show that it's not tied to the JNIEnv lifetime of the new frame (only tied to the lifetime of the array reference).

I've also added a `trybuild` test to double check that even though `AutoElements` is not tied to lifetime of the current JNI stack frame, it is tied to the JNI stack frame that owns the array reference.

PR #512, opened 2023-11-23
Removes the `CStr` coercion from `strings::*`

### LINKED ISSUE #510: `JNIStr` should implement `to_str()` and not `Deref` directly to a `CStr`
An issue was recently hit in the [btleplug ](https://github.com/deviceplug/btleplug) and [bluey ](https://github.com/rib/bluey) Bluetooth LE crates hitting errors converting the `JStrings` device name from the Android `BluetoothDevice` API into a Rust String.

ref: https://github.com/deviceplug/btleplug/issues/349 and https://github.com/rib/bluey/issues/1

and this has highlighted that the current `JNIStr` type can be dereferenced directly into a `CStr` that can trivially bypass the 'Modified UTF8' decoding that JNI strings need to go through when converting into a Rust String.

In particular both crates were inadvertently trying to call a `JavaStr::to_str()` method that doesn't exist and were instead calling through to a dereferenced `CStr::to_str()` that is implemented as `str::from_utf8(self.to_bytes())`.

The only way that a `JNIStr` can currently be decoded as modified UTF8 correctly is via this `From` implementation:

```rust
impl<'str_ref> From<&'str_ref JNIStr> for Cow<'str_ref, str> {
    fn from(other: &'str_ref JNIStr) -> Cow<'str_ref, str> {
        let bytes = other.to_bytes();
        match from_java_cesu8(bytes) {
            Ok(s) => s,
            Err(e) => {
                debug!("error decoding java cesu8: {:#?}", e);
                String::from_utf8_lossy(bytes)
            }
        }
    }
}
```

It looks like we should almost certainly:
1. Remove the `Deref` + `Target = CStr` implementation for `JNIStr`
2. Implement `to_str()` for `JNIStr`

argv-minus-one:
## Overview

This PR fixes #510. It removes the `CStr` coercion from `strings::*`, replacing it with an `as_cstr` method, and adds a `to_str` method that properly transcodes modified UTF-8 into standard UTF-8.

The new `as_cstr` method's documentation contains a warning not to use `to_str` or `to_string_lossy` on the returned `CStr`, and to use `JNIStr::to_str` instead.

This PR also adds several new methods and trait implementations to `strings::*`, makes `JNIString` and `JavaStr` to coerce to `JNIStr` instead, and expands the documentation. Hopefully, this will make the `strings` module easier to use.

Fixes: #510

### Definition of Done

### DIFF
--- a/src/wrapper/strings/ffi_str.rs
+++ b/src/wrapper/strings/ffi_str.rs
@@ -7,54 +7,44 @@ use std::{
 use cesu8::{from_java_cesu8, to_java_cesu8};
 use log::debug;
 
-use crate::wrapper::strings::ffi_str;
-
 #[cfg(doc)]
-use std::ops::Deref;
+use crate::wrapper::strings::JavaStr;
 
-/// An owned, null-terminated string, encoded in Java's [Modified UTF-8].
+/// An owned null-terminated string (like [`CString`]) encoded in Java's
+/// [modified UTF-8].
 ///
-/// Most JNI functions that accept or return strings, such as [`NewStringUTF`],
-/// expect or produce strings encoded this way.
+/// This type is intended for constructing Java strings from Rust code. To use
+/// it, first construct an ordinary Rust [`str`] or [`String`], then use
+/// [`JNIString::new`] to convert it to the encoding that Java expects.
 ///
-/// This type plays a similar role as [`CString`]. Its borrowed counterpart is
-/// [`JNIStr`], which this type [dereferences][Deref] to.
+/// As with `CString`, this type has a borrowed counterpart, [`JNIStr`], that
+/// it coerces to.
 ///
-/// Ordinary Rust strings ([`String`], <code>&amp;[str]</code>, or any other
-/// type implementing <code>[AsRef]&lt;str&gt;</code>) can be converted to this
-/// type using `.into()`. Specifically, this type implements
-/// <code>[From]&lt;T&gt; where T: AsRef&lt;str&gt;</code>.
-///
-/// [Modified UTF-8]: https://docs.oracle.com/en/java/javase/11/docs/specs/jni/types.html#modified-utf-8-strings
-/// [`NewStringUTF`]: https://docs.oracle.com/en/java/javase/11/docs/specs/jni/functions.html#newstringutf
+/// [modified UTF-8]: https://en.wikipedia.org/wiki/UTF-8#Modified_UTF-8
 pub struct JNIString {
     internal: CString,
 }
 
-/// A borrowed, null-terminated string, encoded in Java's [Modified UTF-8].
+/// A borrowed null-terminated string (like [`CStr`]) encoded in Java's
+/// [modified UTF-8].
 ///
-/// Most JNI functions that accept or return strings, such as [`NewStringUTF`],
-/// expect or produce strings encoded this way.
+/// As with `CStr` and [`str`], instances of `JNIStr` cannot be created
+/// directly, but can be borrowed from either [`JNIString`] (for constructing a
+/// Java string in Rust code) or [`JavaStr`] (which refers to an existing Java
+/// string).
 ///
-/// This type plays a similar role as (and [dereferences][Deref] to) [`CStr`].
-/// Its owned counterpart is [`JNIString`].
+/// To convert a `JNIStr` into an ordinary Rust string, use the
+/// [`to_str`][Self::to_str] method. To get a view of the modified UTF-8
+/// encoding of the `JNIStr`, use the [`as_cstr`][Self::as_cstr] method to get
+/// a `CStr`, then call its [`to_bytes`][CStr::to_bytes] method.
 ///
+/// Note that, as with `CStr`, this type is **not** `repr(C)`. See [the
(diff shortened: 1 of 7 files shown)


src/wrapper/strings/ffi_str.rs
```diff
@@ -1,40 +1,59 @@
...
+/// encoding of the `JNIStr`, use the [`as_cstr`][Self::as_cstr] method to get
+/// a `CStr`, then call its [`to_bytes`][CStr::to_bytes] method.
+///
+/// Note that, as with `CStr`, this type is **not** `repr(C)`. See [the
+/// `CStr` documentation][CStr] for an explanation of what that means. (This
+/// type is `repr(transparent)`, but it wraps around a `CStr`, not a raw
+/// pointer.)
+///
+/// [modified UTF-8]: https://en.wikipedia.org/wiki/UTF-8#Modified_UTF-8
+#[repr(transparent)] // needed because `JNIStr` gets pointer-punned from `CStr`.
```
rib, line 47: is this strictly true - I'm not sure it would technically be a problem if it wasn't transparent, considering that for ffi you anyway need to pull out the pointer explicitly. It seems like `transparent` repr would be more important if indeed `CStr` where considered to be ffi safe.

We aren't directly transmuting the types in a way that assumes / requires them to have the same layout, we explicitly construct by pairing up `from_ptr()` and `.as_ptr()`, so hypothetically if there were some spurious extra padding added to the `JNIStr` struct that wouldn't really be a problem?

Having it be explicitly transparent could _maybe_ be misconstrued as suggesting that it's ok to transmute the types which would be another way of unwittingly bypassing the modified-utf8 decode step?

src/wrapper/strings/java_str.rs
```diff
@@ -79,36 +100,35 @@ impl<'local, 'other_local: 'obj_ref, 'obj_ref> JavaStr<'local, 'other_local, 'ob
         Ok(())
     }
 
-    /// Get a [JavaStr] from a [JNIEnv] and a [JString].
-    /// You probably want [JNIEnv::get_string] instead of this method.
-    pub fn from_env(env: &JNIEnv<'local>, obj: &'obj_ref JString<'other_local>) -> Result<Self> {
-        Ok(unsafe {
+    pub(crate) unsafe fn from_env_totally_unchecked(
```
rib, line 103: it's internal api, so it's ok either way, but maybe don't need to say that it's 'totally' unchecked, if there's no naming ambiguity or another api that is also unsafe but with more checks.

rib:
Cool, thanks for looking at this!

As an aside; I half wonder if we could clarify the relationship between `JavaStr` and `JNIString` with a better name - maybe something like `JNIStringData` perhaps? but not sure atm.

This looks good to land to me, esp thanks for the rustdoc additions!

I'm not entirely sure we need the `repr(transparent)`, as noted above, but we could remove that before 0.22 if you agree it's perhaps not needed?

src/wrapper/strings/ffi_str.rs
rib: I realized later that we rely on casting the type references as pointers and casting those. Currently it's supposed to be an implementation detail that a `CStr` is a DST based on a `[c_char]` slice which will map to fat pointers that include the length of the string. I guess hypothetically this could be changed to just be a thin `c_char` pointer in the future and the `repr(transparent)` should ensure we maintain compatibility in that case.

rib (approved):

src/wrapper/strings/ffi_str.rs
argv-minus-one: [A comment in the standard library source code](https://github.com/rust-lang/rust/blob/190f4c96116a3b59b7de4881cfec544be0246d84/library/core/src/ffi/c_str.rs#L92) says `&CStr` probably will be changed to a thin pointer in the future.

src/wrapper/strings/java_str.rs
argv-minus-one: Well, it's called from `JNIEnv::get_string_unchecked`, which is actually not entirely unchecked ([it does a null check](https://github.com/jni-rs/jni-rs/blob/dccd7f94bfff2d804cd23064276fe6ad0e848841/src/wrapper/jnienv.rs#L1877)).

But `JavaStr::get_env_totally_unchecked` *also* isn't entirely unchecked ([it calls](https://github.com/jni-rs/jni-rs/blob/dccd7f94bfff2d804cd23064276fe6ad0e848841/src/wrapper/strings/java_str.rs#L69) `GetStringUTFChars`, which [does unspecified checks of its own](https://docs.oracle.com/en/java/javase/11/docs/specs/jni/functions.html#getstringutfchars)), so…yeah, this is a misnomer too. 🤷‍♂️

I didn't want to leave the name as `from_env` because it would give people confusing compiler errors (“method `from_env` is private”, not “no method named `from_env` found”) when they migrate from an old version of this library. For the same reason, I didn't want to rename it to `new`. I guess I could have named it `from_env_private`?

argv-minus-one:
Yeah, I tried pretty hard to explain the differences between `JNIStr`/`JNIString`/`JavaStr` because I myself was terribly confused by them. I'm still very unsure as to what they could be renamed to that would be any clearer.

PR #528, opened 2024-02-29
Fix local reference leak in `JNIEnv::get_string`

### LINKED ISSUE #527: JNIEnv::get_string should delete the local ref of string_class
When the JVM is created with option "-Xcheck:jni", I get many logs like:
`WARNING: JNI local refs: 3597, exceeds capacity: 3596`

It is caused by calling JNIEnv::get_string, since if I remove them, the warning log would disappear.
As the following code shown, I think the root cause is that the string_class returned by find_class() is not deleted:
```
pub fn get_string<'other_local: 'obj_ref, 'obj_ref>(
        &mut self,
        obj: &'obj_ref JString<'other_local>,
    ) -> Result<JavaStr<'local, 'other_local, 'obj_ref>> {
        let string_class = self.find_class("java/lang/String")?;
        if !self.is_assignable_from(string_class, self.get_object_class(obj)?)? {
            return Err(JniCall(JniError::InvalidArguments));
        }

        // SAFETY: We check that the passed in Object is actually a java.lang.String
        unsafe { self.get_string_unchecked(obj) }
    }
```
Please help check, thanks.

argv-minus-one:
## Overview

This PR fixes #527.

Note: Because there is no JNI function to get a count of how many local references are in the current frame, this change is, as far as I know, impossible to test in an automated fashion.

### Definition of Done

- ~~Change is covered by [automated tests](https://github.com/jni-rs/jni-rs/blob/master/CONTRIBUTING.md#tests)~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - `JavaStr::get_raw` has been renamed to `as_ptr`. ([#510](https://github.com/jni-rs/jni-rs/issues/510) / [#512](https://github.com/jni-rs/jni-rs/pull/512))
 - `JavaStr`, `JNIStr`, and `JNIString` no longer coerce to `CStr`, because using `CStr::to_str` will often have incorrect results. You can still get a `CStr`, but must use the new `as_cstr` method to do so. ([#510](https://github.com/jni-rs/jni-rs/issues/510) / [#512](https://github.com/jni-rs/jni-rs/pull/512))
 
+### Fixed
+- `JNIEnv::get_string` no longer leaks local references. ([#527](https://github.com/jni-rs/jni-rs/issues/527))
+
 ## [0.21.1] — 2023-03-08
 
 ### Fixes
--- a/src/wrapper/jnienv.rs
+++ b/src/wrapper/jnienv.rs
@@ -1919,7 +1919,7 @@ impl<'local> JNIEnv<'local> {
         &mut self,
         obj: &'obj_ref JString<'other_local>,
     ) -> Result<JavaStr<'local, 'other_local, 'obj_ref>> {
-        let string_class = self.find_class("java/lang/String")?;
+        let string_class = AutoLocal::new(self.find_class("java/lang/String")?, self);
         let obj_class = self.get_object_class(obj)?;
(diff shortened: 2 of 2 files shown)


olivergillespie:
I am working on a performance optimization for get_string which might sidestep this issue. Should I open that separately or does it make sense to look at it here?

We can cache the lookup of the String jclass, so we don't need to call find_class every time.

```
jni_get_string          time:   [445.86 ns 448.89 ns 454.23 ns]
                        change: [-53.617% -45.104% -39.388%] (p = 0.00 < 0.05)
                        Performance has improved.
```

```
     pub fn get_string<'other_local: 'obj_ref, 'obj_ref>(
         &mut self,
         obj: &'obj_ref JString<'other_local>,
     ) -> Result<JavaStr<'local, 'other_local, 'obj_ref>> {
-        let string_class = self.find_class("java/lang/String")?;
+        static STRING_CLASS: OnceLock<JClass> = OnceLock::new();
+        let string_class = STRING_CLASS.get_or_init(|| self.find_class("java/lang/String").unwrap());

         let obj_class = self.get_object_class(obj)?;
         if !self.is_assignable_from(string_class, obj_class)? {
             return Err(JniCall(JniError::InvalidArguments));
```             

I'm working on a better implementation, but that shows the performance at least.

Side note, I think we can also get a further improvement knowing that String is a final class:

```
-         let obj_class = self.get_object_class(obj)?;
-         if !self.is_assignable_from(string_class, obj_class)? {
+         if !self.is_instance_of(obj, string_class)? {
             return Err(JniCall(JniError::InvalidArguments));
```

```
jni_get_string          time:   [406.30 ns 406.37 ns 406.45 ns]
                        change: [-9.6565% -9.1695% -8.8758%] (p = 0.00 < 0.05)
                        Performance has improved.
```

rib:
> I am working on a performance optimization for get_string which might sidestep this issue - see #531.

Just to mention here that although I also agree with the direction of the optimization in #531, I'm going to first look at merging this and #557 so we at least land fixes for the `get_string()` leaks, and even if those become redundant with #531 we can then look at enabling global caching of class references.

rib (approved):
Thanks, this fix looks good to me.

Technically the changelog comment isn't entirely accurate since there's also the leak from the line below that's addressed in #557, but it seems good to land this and, and the changelog can be tweaked later to also reference #557

PR #534, opened 2024-04-18
refactor!: JavaType::Object holds shared pointer

### LINKED ISSUE #499: Reduce clone overhead for JavaType
Currently, we have:

```rust
#[allow(missing_docs)]
#[derive(Eq, PartialEq, Debug, Clone)]
pub enum JavaType {
    Primitive(Primitive),
    Object(String),
    Array(Box<JavaType>),
    Method(Box<TypeSignature>),
}
```

When trying to build a signature programmatically, `JavaType::Object(String)` can cause identical types copy the internal `String`.

For example:

```rust
let ty = TypeSignature {
// even we define the args once and do .clone(), it creates two string instances and thus copy.
            args: vec![JavaType::Object("java/lang/Boolean".to_string()), JavaType::Object("java/lang/Boolean".to_string())],
            ret: ReturnType::Object,
        };
```

Here is a patch for use:

```diff
diff --git a/src/wrapper/signature.rs b/src/wrapper/signature.rs
index 6f86f1b..994e80c 100644
--- a/src/wrapper/signature.rs
+++ b/src/wrapper/signature.rs
@@ -1,4 +1,5 @@
 use std::{fmt, str::FromStr};
+use std::sync::Arc;
 
 use combine::{
     between, many, many1, parser, satisfy, token, ParseError, Parser, StdParseResult, Stream,
@@ -43,7 +44,7 @@ impl fmt::Display for Primitive {
 #[derive(Eq, PartialEq, Debug, Clone)]
 pub enum JavaType {
     Primitive(Primitive),
-    Object(String),
+    Object(Arc<String>),
     Array(Box<JavaType>),
     Method(Box<TypeSignature>),
 }
@@ -185,7 +186,7 @@ where
     let end = token(';');
     let obj = between(marker, end, many1(satisfy(|c| c != ';')));
 
-    obj.map(JavaType::Object).parse_stream(input).into()
+    obj.map(|o| JavaType::Object(Arc::new(o))).parse_stream(input).into()
 }
 
 fn parse_type<S: Stream<Token = char>>(input: &mut S) -> StdParseResult<JavaType, S>
```

tisonkun:
## Overview

This closes #499 

### Definition of Done

### DIFF
--- a/src/wrapper/signature.rs
+++ b/src/wrapper/signature.rs
@@ -1,4 +1,5 @@
 use std::{fmt, str::FromStr};
+use std::sync::Arc;
 
 use combine::{
     between, many, many1, parser, satisfy, token, ParseError, Parser, StdParseResult, Stream,
@@ -43,7 +44,7 @@ impl fmt::Display for Primitive {
 #[derive(Eq, PartialEq, Debug, Clone)]
 pub enum JavaType {
     Primitive(Primitive),
-    Object(String),
+    Object(Arc<String>),
     Array(Box<JavaType>),
     Method(Box<TypeSignature>),
 }
@@ -185,7 +186,7 @@ where
     let end = token(';');
     let obj = between(marker, end, many1(satisfy(|c| c != ';')));
 
-    obj.map(JavaType::Object).parse_stream(input).into()
+    obj.map(Arc::new).map(JavaType::Object).parse_stream(input).into()
 }
 
 fn parse_type<S: Stream<Token = char>>(input: &mut S) -> StdParseResult<JavaType, S>


tisonkun:
@rib any thoughts?

knopp:
Not commenting whether this is a good idea or not, but `Arc<String>` needs two allocations, while `Arc<str>` would only need one.

rib:
Hey, sorry for not being able to look at this until now.

The representation of `JavaType` with owned `Strings` has always bugged me, so I'd love to see this be improved.

One thing I did recently was remove `JavaType::Method`: https://github.com/jni-rs/jni-rs/pull/602 which I was hoping might be a stepping stone for further simplification.

Today I was investigating the `lasso` and `string-interner` crates as possible ways in which we could intern object names and exchange them for u32 handles so we would effectively have `JavaType::Object(u32)` but still be able to get back to the string names.

This still doesn't avoid the the `Box` that's in `JavaType::Array` which I also really dislike, but is currently required for allowing recursive `Array` definitions.

I also experimented with making the signature parser so it borrowed from its input to avoid owning the object name Strings but that doesn't really solve much, it just moves the ownership and copying somewhere else (and makes `JavaType/TypeSignature` impractical).

Taking a step back and reviewing how `JavaType` is used though (at least internally) it doesn't actually look like anything needs to preserve these object names. Everything that processes a `JavaType` (except when formatting/printing) actually discards these object names).

For example, when checking arguments in `call_method` (and similar `call_` APIs) we do this:

```rust
        // check arguments types
        let base_types_match = parsed
            .args
            .iter()
            .zip(args.iter())
            .all(|(exp, act)| match exp {
                JavaType::Primitive(p) => act.primitive_type() == Some(*p),
                JavaType::Object(_) | JavaType::Array(_) => act.primitive_type().is_none(),
            });
```

and in `set_field` we match like:

```rust
    match field_ty {
        JavaType::Object(_) | JavaType::Array(_) if val_primitive.is_some() => wrong_type,
        JavaType::Primitive(p) if val_primitive != Some(p) => wrong_type,
        JavaType::Primitive(_) if val_primitive.is_none() => wrong_type,
        _ => {
            let class = self.get_object_class(obj)?.auto();

            // Safety: We have explicitly checked that the field type matches
            // the value type
            unsafe { self.set_field_unchecked(obj, (&class, name, ty), val) }
        }
    }
```

I'm starting to think we should experiment with just keeping the `ReturnType` enum which can be renamed `JavaType` and stop trying to track a fully-qualified AST for JNI method signatures. All we really want is a way to differentiate primitive types and reference types (all objects and arrays).

I'm tending to think that `TypeSignature` should be renamed to `MethodSignature` or maybe just `Signature` which should use a simplified `JavaType` enum to represent the args and return type that is equivalent to the current `ReturnType` enum.

The `Signature` should then probably also own an internal `JNIString` that's a pre-encoded copy of the signature (needed for being able to use a `Signature` for method ID lookups) and that should also be able to help amortise some of the cost of using a signature to repeatedly lookup method IDs in some cases.

I'm going to take a pass this and see how it looks.

PR #552, opened 2024-11-18
Port from windows-sys to windows-bindgen

complexspaces:
## Overview

This PR implements an alternative to https://github.com/jni-rs/jni-rs/pull/513 that requires less maintenance overall and has a lower chance of bitrotting in various user's dependency trees/lockfiles. 

In summary, the changes proposed here move away from using pre-generated `windows-sys` crates to using the `windows-bindgen` tool to make the minimal Windows binding code required for `jni`, which only uses a very small surface area. This is done by introducing a small, manually run CLI, which uses `windows-bindgen` to create the required file and updating `jni`'s import paths to use that instead.

### Definition of Done


Cargo.toml
```diff
@@ -49,7 +49,7 @@ lazy_static = "1"
 rusty-fork = "0.3.0"
 
 [target.'cfg(windows)'.dependencies]
-windows-sys = { version = "0.48.0", features = ["Win32_Globalization"] }
+windows-targets = { version = "0.52.0" }
```
complexspaces, line 52: Note that while this still needs something from the `windows-*` ecosystem, they've made `windows-targets` pretty stable and it hasn't had a breaking change in over a year.

rib (approved):
Thanks for looking at this - this looks good to me :+1:

rib:
ah, for running `cargo test` it just looks like `windows_sys.rs` also needs a binding for `MultiByteToWideChar`

complexspaces:
If you don't mind an additional change coming in here (or I can open a followup PR), I was going to investigate expanding the `windows-targets = { version = "0.52.0" }` requirement to allow a range in order to support the new `0.53` version series of the crate that came out in January, making 1.3 years between releases. This would be done by trying to compile with both `0.52` and `0.53` in the lockfile to see if there were breaking changes between.

complexspaces:
> Thanks, the follow up sounds good to me and I can wait a bit to merge.

Sorry, to clarify: Do you want that additional change as...
- another commit on this branch?
- a followup PR targeting this branch, which will be merged into this branch and then the combined result into `main`?
- a followup PR targeting this branch will merge after this one does into `main`?

> I also wondered if it'd be worth looking at bumping to windows-bindgen 0.59...

IMO its not worth doing right now and could be investigated at a later point. The types used in the bindings here are not complex (they are all aliases or constants) so I don't think the generated output for `jni` will change enough to be worthwhile today.

ssrlive:
I have created a PR #653 to alternative PR. do you think it's OK?

complexspaces:
The PR you created kind of defeats the point of this one, respectfully. 

My goal with this PR was to remove the need to constantly bump `windows-sys` to the latest version and instead only use a crate with a much longer version lifetime so that the `jni` maintainers need to do less work and so the time between `jni` releases is less fraught for the lockfiles of downstream users.

ssrlive:
I think it's not a good idea to add a lot of code in unimportant places.

complexspaces:
This PR adds ~13 lines of actual code used in production builds, and a few for development tooling. The rest is boilerplate or will never be touched again.

ssrlive:
Well. you can change [windows-targets to `v0.53.0`](https://crates.io/crates/windows-targets)  and  [windows-bindgen
v0.59.0](https://crates.io/crates/windows-bindgen) and add `Windows.Win32.Globalization.MultiByteToWideChar` to the config file.

ssrlive:
Well. you can change [windows-targets to `v0.53.0`](https://crates.io/crates/windows-targets)  and  [windows-bindgen
v0.59.0](https://crates.io/crates/windows-bindgen) and add `Windows.Win32.Globalization.MultiByteToWideChar` to the config file.

As you can see, you still have to maintain the versions of `windows-targets` and `windows-bindgen`.

complexspaces:
Please see my comments about about how long `windows-targets 0.52` lasted and how I am trying to test a cargo version spec that would let `jni` use _either_ 0.52 or 0.53 for optimal compatibility.

> ... add Windows.Win32.Globalization.MultiByteToWideChar to the config file.

This is just part of the initial transition away from `windows-sys` and something I would have taken care of in my followup work once I hear back where it should go. Its not nearly as big of a deal as you observe it to be.

> ...As you can see, you still have to maintain the versions of windows-targets and windows-bindgen.

You are halfway correct about `windows-targets` but missing the point about `windows-bindgen`. It spits out code to be checked into this repository and is never touched between uses. It doesn't need updated ever technically because it doesn't end up in downstream user's lockfiles and the `jni` doesn't use new Windows functions almost ever. `0.59` is completely optional for this crate to consider.

ssrlive:
However, the version of `windows-targets` must be updated anyway. Otherwise the new compiler may no longer support the old `windows-targets` version. I have encountered this before. Another project of mine needed to use the latest compiler to generate an executable file for win7, but it relied on `windows-sys` version `0.48`-- Yes, it's `0.48`. In the end, the compilation failed and I had to castrate some functions under windows.

complexspaces:
If you'd like I can investigate just dropping the `windows-target` dependency in favor of an `extern "system" {}` block? All three functions used come from `kernel32.dll`, which already needs to be properly linked by `std` to support Rust itself on Windows so maybe `windows-targets` isn't gaining us much here.

ssrlive:
If you can accomplish this, I'd be grateful to you for doing so.

complexspaces:
I managed to pull this off in https://github.com/complexspaces/jni-rs/commit/773a1f61914c5a5a5d9d97028f258a0121b61fc1, which currently lives in a temporary branch until I know where to put it for real. I confirmed it links + runs via `cargo test --all-features` on a Windows system. It also covers the `windows-bindgen` update and fixing `MultiByteToWideChar`.

The only downside is that it vendors a little bit of raw-dylib code from `windows-link` (which is not yet released) because we still need to slot into the expected structure `windows-bindgen` looks for. I don't think its too bad though, it would just be nice to drop in `windows-link` at a later date because that crate has no dependencies and should never have its minor version bumped.

rib:
> I managed to pull this off in [complexspaces@773a1f6](https://github.com/complexspaces/jni-rs/commit/773a1f61914c5a5a5d9d97028f258a0121b61fc1), which currently lives in a temporary branch until I know where to put it for real. I confirmed it links + runs via `cargo test --all-features` on a Windows system. It also covers the `windows-bindgen` update and fixing `MultiByteToWideChar`.
> 
> The only downside is that it vendors a little bit of raw-dylib code from `windows-link` (which is not yet released) because we still need to slot into the expected structure `windows-bindgen` looks for. I don't think its too bad though, it would just be nice to drop in `windows-link` at a later date because that crate has no dependencies and should never have its minor version bumped.

Nice! this seems like the way to go - the vendored `win_link` macro seems reasonable.

I think it'd be fine to force push that to the branch for this PR and we can land that instead.

complexspaces:
@rib Thanks for the positive feedback! I've gone ahead and force pushed that work here as requested. This should be ready for another CI run, and hopefully, merge afterwards.

complexspaces:
Hey again @rib, sorry to bother,

I was wondering if this has an estimated release timeline to make it onto crates.io.

gaoqiangz:
When will it be released to crates.io? I also encountered Win7 compatibility issues.

PR #554, opened 2024-11-21
In `JavaVM::{new,with_libjvm}`, prevent libjvm from being unloaded.

### LINKED ISSUE #550: After loading the JVM library, the Library instance was not retained, causing methods like get_env to fail to execute properly.
After loading the JVM library, the Library instance was not retained, causing methods like get_env to fail to execute properly.
image

## Env

os: windows 10
rust: v1.82.0
jni: v0.21.1

## Code

```rust
// ...
let jvm = JavaVM::new(init_args).unwrap();
let mut env = jvm.get_env().unwrap();    // Error msg: called `Result::unwrap()` on an `Err` value: JniCall(ThreadDetached)
// ...
```

> After testing and verification, retaining the Library value can resolve the issue.

argv-minus-one:
## Overview

If libjvm is unloaded while it's running, the process will crash as soon as the JVM tries to do anything, since all of its code has been unloaded. `JavaVM::{new,with_libjvm}` does indeed unload libjvm, by dropping the `libloading::Library` that was used to load it. This PR solves that problem by using `std::mem::forget` on `libloading::Library`, which prevents libjvm from ever being unloaded.

Somehow, HotSpot seems to prevent itself from being unloaded. That's why our tests weren't failing. But I have no idea how it's doing that. See discussion in #550. In any case, other JVMs might not prevent themselves from being unloaded, so they could still crash.

It would be ideal if we could unload libjvm in `JavaVM::destroy`, but we can't. The `JavaVM` structure is `repr(transparent)`, so we can't store the `libloading::Library` there, nor does JNI offer any other place to store it. So, this PR plays it safe by simply preventing libjvm from being unloaded at all.

**Question:** Does `JavaVM` actually need to be `repr(transparent)`? If not, then we could store the `libloading::Library` inside it, and unload it in `JavaVM::destroy`.

Fixes #550

### Definition of Done

### DIFF
--- a/src/wrapper/java_vm/vm.rs
+++ b/src/wrapper/java_vm/vm.rs
@@ -196,7 +196,7 @@ impl JavaVM {
             Err(error) => return Err(StartJvmError::LoadError(libjvm_path_string, error)),
         };
 
-        unsafe {
+        let result = unsafe {
             // Try to find the `JNI_CreateJavaVM` function in the loaded library.
             let create_fn = libjvm
                 .get(b"JNI_CreateJavaVM\0")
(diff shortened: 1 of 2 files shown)


src/wrapper/java_vm/vm.rs
```diff
@@ -196,15 +196,36 @@ impl JavaVM {
...
+        //
+        // For reasons unknown, HotSpot seems to somehow prevent itself from being unloaded, so it
+        // will work even if this `forget` call isn't here, but there's no guarantee that other JVM
+        // implementations will also prevent themselves from being unloaded.
+        //
+        // Note: `jni-rs` makes the assumption that there can only ever be a single `JavaVM`
+        // per-process and it's never possible to full destroy and unload a JVM once it's been
+        // created. Calling `DestroyJavaVM` is only expected to release some resources and
+        // leave the JVM in a poorly-defined limbo state that doesn't allow unloading.
+        // Ref: https://github.com/jni-rs/jni-rs/issues/567
```
rib, line 223: I took the liberty of tweaking this comment to refer to the issue that discusses making the assumption that the JVM can't be unloaded - and removed the notes about potentially storing the handle so it could hypothetically be dropped.

rib (approved):
Good stuff, this looks good to me! Thanks :+1:

Issue #469, opened 2023-07-06
`attach_current_thread_as_daemon` (and any notion that we support 'daemon' threads) should probably be removed


rib:
The low-level JNI API supports the idea that you can attach a thread to a JVM in such a way that it won't automatically get detached from the JVM when destroying the VM!

The high-level idea behind Java daemon threads is that applications could use them for work that they don't necessarily want to block the application from exiting (things that run continuously like watchdogs/loggers).

The problem with mapping that idea to JNI thread attachment though is that it leaves a dangerous race condition between destroying the VM and exiting the application + terminating the daemon threads. If any of those daemon threads do anything to interact with JNI / the VM they are going to hit undefined behaviour.

In the context of this crate which is trying to provide a (mostly) safe JNI API I really struggle to see any way in which daemon thread attachments can be exposed safely.

Since it's relevant here I'll copy docs I wrote for `detach_current_thread` which highlight how we currently require applications to explicitly detach daemon threads if they call `JavaVM::destroy()` to avoid this risk of undefined behaviour:

```rust
/// Explicitly detaches the current thread from the JVM.
///
/// _**Note**: This operation is _rarely_ appropriate to use, because the
/// attachment methods [ensure](#attaching-native-threads) that the thread
/// is automatically detached._
///
/// Detaching a non-attached thread is a no-op.
///
/// To support the use of `JavaVM::destroy()` it may be necessary to use this API to
/// explicitly detach daemon threads before `JavaVM::destroy()` is called because
/// `JavaVM::destroy()` does not synchronize and wait for daemon threads.
///
/// Any daemon thread that is still "attached" after `JavaVM::destroy()` returns would
/// cause undefined behaviour if it then tries to make any JNI calls or tries
/// to detach itself.
///
/// Normally `jni-rs` will automatically detach threads from the `JavaVM` by storing
/// a guard in thread-local-storage that will detach on `Drop` but this will cause
/// undefined behaviour if `JavaVM::destroy()` has been called.
///
/// Calling this will clear the thread-local-storage guard and detach the thread
/// early to avoid any attempt to automatically detach when the thread exits.
///
/// # Safety
///
/// __Any existing `JNIEnv`s and `AttachGuard`s created in the calling thread
/// will be invalidated after this method completes. It is the__ caller’s __responsibility
/// to ensure that no JNI calls are subsequently performed on these objects.__
/// Failure to do so will result in unspecified errors, possibly, the process crash.
///
[...]

PR #555, opened 2024-11-22
Make `JavaVM` thread-attaching methods fail if the current thread is already attached

### LINKED ISSUE #548: Segmentation fault might be caused by mixed usages of `JavaVM::attach_current_thread()` and `JavaVM::attach_current_thread_permanently()`
It is unlikely that it would occur in real-world use cases, though.

Minimum program for producing the fault on PC with Java installed:

```
[package]
name = "jni-env-test"
version = "0.1.0"
edition = "2021"

[dependencies]
jni = { version = "0.21.1", features = ["invocation"] }
```

```rust
fn main() {
    let args = jni::InitArgsBuilder::new()
        .version(jni::JNIVersion::V8)
        .build()
        .unwrap();
    let vm = jni::JavaVM::new(args).unwrap();
    let env = vm.attach_current_thread().unwrap();
    let env_perm = vm.attach_current_thread_permanently().unwrap();
    drop(env);
    let _ = env_perm.new_string("").unwrap();
}
```

Fortunately, placing initialization of `env_perm` before `env` makes it work, thanks to `AttachGuard::new_nested(env)` (`src/wrapper/java_vm/vm.rs`) which doesn't detach the current thread from JVM on dropping. This current behavior makes the machanism below work (multiple Android support crates being used):

```rust
// bold emulation of obtaining a jni::JavaVM on Android (see crate `ndk-context`)
fn java_vm() -> &'static jni::JavaVM {
    use std::sync::OnceLock;
    static JAVA_VM: OnceLock<jni::JavaVM> = OnceLock::new();
    JAVA_VM.get_or_init(|| {
        let args = jni::InitArgsBuilder::new()
            .version(jni::JNIVersion::V8)
            .build()
            .unwrap();
        jni::JavaVM::new(args).unwrap()
    })
}

fn lib1_do_something() {
    with_jni_env(|env| {
        let s = env.new_string("lib1_do_something()").unwrap();
        let s: String = env.get_string(&s).unwrap().into();
        println!("{s}");
    });
}

fn with_jni_env<R>(f: impl FnOnce(&mut jni::JNIEnv) -> R) -> R {
    let mut env = java_vm().attach_current_thread().unwrap();
    f(&mut env)
}

fn lib2_do_something() {
    with_jni_env_perm_attach(|env| {
        lib1_do_something();
        let s = env.new_string("lib2_do_something()").unwrap();
        let s: String = env.get_string(&s).unwrap().into();
        println!("{s}");
    });
}

fn with_jni_env_perm_attach<R>(f: impl FnOnce(&mut jni::JNIEnv) -> R) -> R {
    let mut env = java_vm().attach_current_thread_permanently().unwrap();
    f(&mut env)
}

fn main() {
    lib2_do_something();
}
```

Do you think this issue should be closed?

argv-minus-one:
## Overview

This PR changes the `JavaVM` methods that attach the current thread such that they fail if the current thread is already attached.

The reason for this is that, if they return the `JNIEnv` for the already-attached current thread, then they are equivalent to `JNIEnv::get_env`, which is `unsafe`. This is no longer a merely theoretical concern; @wuwbobo2021 got segfaulted as a result of this unsoundness and reported it in #548.

Closes #548

## Outstanding Issue

`Executor` currently doesn't work, and the tests for it currently fail, because `Executor` relies on being able to repeatedly attach the current thread, which is no longer allowed.

Other than removing `Executor` entirely, I'm not sure how to fix this. `Executor` could call `JavaVM::get_env` if the current thread is already attached, but what if the passed closure drops an `AttachGuard`?

### Definition of Done


rib:
Every time I try and think through closing up the soundness holes around JNI thread attachment I become convinced that the notion of supporting RAII auto-detachment guards is practically impossible to support safely, and I think we should seriously consider removing `AttachGuard` entirely and even making thread attachment / detachment `unsafe`.

One of the traps with trying to think about solutions here is to forget that the `jni-rs` Rust crate is never the source of truth for global JNI state because we just provide one set of bindings for one language. Even when we think of Rust being the centre of the world, this crate has no guarantee that an application does link in multiple versions of `jni-rs` into one application which can e.g. result in disjoint  TLS state for attempting to track thread attachment counts. There are also alternative JNI bindings that Rust crates can use that can warp any attempt to track global state locally (i.e. in TLS).

So although there are a few things we can try and do to plug safety issues that arguably come directly from `jni-rs` we can't take full ownership of global JNI state, which effectively belongs to the JVM implementation and may be manipulated via other non-jni-rs bindings or even different versions of jni-rs linked into the same binary.

I think currently I would lean towards us removing `AttachGuard`, and effectively rename `attach_current_thread_permanently` to `attach_current_thread` but make it `unsafe`. `detach_current_thread` should be a more direct (`unsafe`) binding for `DetachCurrentThread` but the docs should still generally warn that most applications probably shouldn't ever explicitly detach threads. I think in general JNI is not designed to support accurate micro-management of thread attachments and you just have to err on the side of asserting an attachment at entrypoints for your library and leave attachments in tact in case code outside of your control depends on the attachment. I guess there are probably relatively few cases where you can be _really_ confident that JNI thread detachment is both necessary and safe.

Although it's unfortunate for ergonomics to have an important API like this be `unsafe`, I think it's probably still the right thing to do, and then encourage that libraries would only handle attachments at the frontends of their APIs before then passing around JNIEnv references internally that they can use safely, until their API returns.

On a semi-related note, I think it could also be a good time to remove `attach_current_thread_as_daemon` which is especially fraught with safety issues (re: https://github.com/jni-rs/jni-rs/issues/469)

rib:
As a caveat it's notable that since we are strictly required to explicitly detach from native threads before they terminate then we probably do at least need to keep the `InternalAttachGuard` functionality for being able to ensure an attached thread gets detached automatically (only as long as we _really_ did attach the thread). I.e. preserving the existing semantics for `attach_current_thread_permanently`.

rib:
Besides considering whether we could remove `AttachGuard`, as above, I think my inclination here would also be to make the `attach_current_thread` APIs themselves `unsafe` instead of having them return an Error when they don't attach, and not make it the caller's responsibility to effectively manually implement the fallback to calling `JavaVM::get_env()`.

`jni-rs` can be a dependency of numerous, orthogonal, crates and so it's reasonable to expect that each of those crates won't necessarily seamlessly pass `JNIEnv` references into their frontend APIs and separate crates may each make (potentially redundant) thread attachments and they can each have a valid need to acquire a `JNIEnv` which makes sense to acquire though a standard "attach current thread" operation - without really caring about whether they were _really_ responsible for the attachment.

Thread attachment being `unsafe` would seem quite reasonable to me - while I would expect that in many cases where you're creating a native thread and manually attaching JNI there's a good chance you're needing to use `unsafe` code to access a `VM` pointer anyway.

Using the "Safety" docs to stipulate that you're never allowed to use thread attachment to effectively materialise access to a `JNIEnv` if you already have some other `JNIEnv` in scope might also be a pragmatic way to ensure references can't escape a local stack frame.

(and having thread attachment be `unsafe` seems better than e.g. making `with_local_frame()` `unsafe` while I would expect thread attachment to happen at the periphery/frontend of an API - likely coupled with other FFI details that are `unsafe`)

rib:
I've opened this alternative draft PR that overlaps with the changes proposed here - it would be great to get your thoughts @argv-minus-one : https://github.com/jni-rs/jni-rs/pull/570

rib:
The issues with thread attachment safety have hopefully been addressed in #570 :crossed_fingers:

PR #561, opened 2025-02-02
Add `try_into_raw` for `GlobalRef`

elichai:
I'm storing these global refs in a static, so they won't actually be dropped unless the whole shared library will be unloaded (this can be handled in `JVM_Unload` if someone really wants), So I have no need for their "complex" type.

### DIFF
--- a/src/wrapper/objects/global_ref.rs
+++ b/src/wrapper/objects/global_ref.rs
@@ -1,4 +1,9 @@
-use std::{mem, ops::Deref, sync::Arc};
+use std::{
+    mem::{self, ManuallyDrop},
+    ops::Deref,
+    ptr,
+    sync::Arc,
+};
 
 use log::{debug, warn};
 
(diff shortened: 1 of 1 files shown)


rib (approved):
Thanks, this looks reasonable to me.

Incidentally, with `jni-rs` now making the assumption that a JVM can't be unloaded, such that it can cache `java.lang.` class references (ref: https://github.com/jni-rs/jni-rs/issues/567) it should also be possible to make some simplifications to the `GlobalRef` implementation itself.

We should be able to remove the vm pointer from each GlobalRef, and instead of creating a separate Arc heap allocation for ref-counting the global references it might be nicer if the type were just `!Copy` and `!Clone` and only wrapped the jni_sys reference pointer - so then it would be the user's responsibility to choose to wrap into a shareable Arc if that's useful.

I'd guess the `Arc` is redundant in lots of cases where applications have some natural state struct for holding on to their global references.

elichai:
@rib Thanks!
In my code I did very similar caching to a dozen classes, (although I initialize them on JVM_ONLoad instead of first use) so would love to see any kinds of improvements on that

(I also use GlobalRef in the "regular" way to allocate CompletableFuture and complete+dealloc the global ref at a later time)

PR #563, opened 2025-02-11
bump to windows-sys 0.59

ssrlive:


rib:
This has hopefully been addressed now by https://github.com/jni-rs/jni-rs/pull/552

Issue #558, opened 2024-12-05
Lifetime restriction failure in methods that return `AutoLocal`


wuwbobo2021:
Simplified test:

```
use jni::objects::JObject;
type AutoLocal<'a> = jni::objects::AutoLocal<'a, JObject<'a>>;

fn main() {
    let t = Test {};
    let obj = t.test();
    let env = jni_get_vm().attach_current_thread().unwrap();
    let _ = env.get_object_class(&obj); // Segmentation fault (core dumped)
}

struct Test;

impl Test {
	fn test(&self) -> AutoLocal<'_> {
		let env = jni_get_vm().attach_current_thread().unwrap();
		let obj = env.new_string("a").unwrap();
		env.auto_local::<JObject<'_>>(obj.into())
	}
}

pub fn jni_get_vm() -> &'static jni::JavaVM {
    use std::sync::OnceLock;
    static JAVA_VM: OnceLock<jni::JavaVM> = OnceLock::new();
    JAVA_VM.get_or_init(|| {
        let args = jni::InitArgsBuilder::new().build().unwrap();
        jni::JavaVM::new(args).unwrap()
    })
}
```

wuwbobo2021:
Trying to make it a bit clearer:

```rust
impl Test {
    fn test<'local>(&self) -> AutoLocal<'local> {
        use std::ops::Deref;
        let env = jni_get_vm().attach_current_thread().unwrap();
        // What's the lifetime of `env` "'_" ? Setting it manually will eliminate the problem
        let env: &'_ jni::JNIEnv<'local> = env.deref();
        let obj: JObject<'_> = env.new_string("a").unwrap().into();
        env.auto_local(obj)
    }
}
```

wuwbobo2021:
Actually, I found this bug while testing my `jni-min-helper` which uses `AutoLocal` frequently: a function may return a `AutoLocal` which can be used wrongly! It should return `GlobalRef` instead in such cases.

Further modification of the test case posted above:

```rust
impl Test {
    // Wrong: 'local should be considered valid *inside* the function,
    // it does not outlive &'a self, it shouldn't be compiled!
    fn test<'a, 'local: 'a>(&'a self) -> AutoLocal<'a> {
        use std::ops::Deref;
        let env: jni::AttachGuard<'local> = jni_get_vm().attach_current_thread().unwrap();
        // What's the lifetime of `env` "'_" ? Setting it manually eliminates the unsound compilation
[...]

Issue #477, opened 2023-07-27
`push/pop_local_frame` should be internal-only APIs or removed altogether


rib:
Not only are these really unsafe APIs to access directly, they are also basically just tiny wrappers around the raw JNI function call and so there's really very little value in having them. Even internally we could perhaps just call the JNI functions directly where we need to.

ivanbuper:
Hi @rib, are you still looking for someone to take a look at this? I'd be happy to give it a try with some guidance.

Issue #470, opened 2023-07-06
Remove/deprecate Executor API?


rib:
The `Executor` API is an elaborate wrapper over:

```rust
let mut jni_env = self.vm.attach_current_thread_as_daemon()?;
jni_env.with_local_frame(capacity, |jni_env| f(jni_env))
```

Considering https://github.com/jni-rs/jni-rs/issues/469 it seems like we should at least change this to not attach as a daemon thread but maybe we could instead expose an API like `vm.attach_with_local_frame()` and deprecate or remove the `Executor` API?

PR #570, opened 2025-03-14
Re-work JNI thread attachments and borrow `JNIEnv` lifetimes from `AttachGuard`s

rib:
The main changes here are that:

1. All thread attachments are now explicitly and consistently represented by an `AttachGuard`

    - `JavaVM::attach_current_thread` requests a permanent thread attachment (the shorter, more-natural, API name is now used to request permanent attachments as a way to discourage the use of scoped attachments by default - since scoped attachments make it more likely that you'll repeatedly incur the higher cost of attaching/detaching a thread)
    - `JavaVM::attach_current_thread_for_scope` requests an RAII thread attachment
    - `AttachGuard::from_unowned` explicitly represents the thread attachment that's implied for native Java method implementations.

2. The `JNIEnv` API is now only accessible by borrowing an `env` reference from a thread `AttachGuard`

   It should no longer be possible to get a `JNIEnv` with a static lifetime (assuming adherence to the Safety rules when attaching the thread that require you to keep the `AttachGuard` on the stack)

3. Thread attachment is considered `unsafe` - including some rules about how `AttachGuard`s need to be kept on the stack.
   
   (I'm not convinced atm that we can create a water-tight safe abstraction for JNI thread attachment, but the implementation does go some way towards at least trying, so there is still wiggle room here)

4. The existence of any `AttachGuard`s on the stack acts as a blocker for detaching the current thread.

   I.e. a permanently attached thread can't be detached (even `unsafe`ly) while there is an `AttachGuard` that would become an invalid pointer.

   `JavaVM::detach_current_thread` is only applicable to threads attached permanently with `JavaVM::attach_current_thread` and the new constraints meant the API doesn't need to be `unsafe`.

The implementation for `AttachGuard` tracks a nesting level such that we can assert that a `JNIEnv` is only ever borrowed from the top-most guard

TODO: we should now also be able to associate the `JNIEnv` with a level that can be checked within any method taking a mutable reference that may create new local references.

We no longer try to be clever with how scoped and permanent attachments are integrated. A permanent attachment will use thread local storage to install a `TLSAttachGuard` similar to before, but scoped attachments never install a `TLSAttachGuard` since it's assumed to be adequate that dropping the `AttachGuard` will detach the thread before any `TLSAttachGuard` would do anything.

TODO: allow passing optional extra info when attaching threads, including a thread name.

TODO: remove `JNIEnv::unsafe_clone`! Initially it should be possible to replace with `JavaVM::get_env_attachment()` (similarly `unsafe`) - since it's implied that any code cloning a `JNIEnv` depends on an existing attachment.

TODO: Implement something like `JavaVM::fast_get_env_attachment()` that will only return an `AttachGuard` if the current thread has been attached already using the `jni-rs` crate - so it doesn't need to call into `GetEnv()` (or rather this should be how `get_env_attachment` works and the version that calls `GetEnv()` can potentially be a private API)

Fixes: #441
Fixes: #469
Fixes: #470
Fixes: #477
Fixes: #548
Fixes: #558
Addresses: #422
Addresses: #560


rib:
Heya @wuwbobo2021 and @argv-minus-one; sorry that this is quite a bit dump of draft changes but it would be great to get some more eyes here to see what you think about this approach to addressing some of the thread attachment and JNIEnv lifetime issues that we have currently.

wuwbobo2021:
Thanks for the invitation, I am an amateur and I can't take the responsibility of being a regular member of the `jni-rs` team maintaining an important library. I have looked at these main changes, it looks great. These soundness issues seems resolved, and the API remains "usable". Here are a few notes:

- I guess you are making `AttachGuard` creation `unsafe` because the introduced attach `level` mechanism cannot really prevent an outer `JNIEnv` from being used in an inner frame. To make it safe, at least every safe JNI method of `JNIEnv` creating a new local reference needs to check the attach level, and I guess you are not doing so because of the possible overhead.

- If local frame management can be made `unsafe`, a Java callback function will probably not need an unsafe `AttachGuard::from_unowned` call in order to use the `JNIEnv`, and `AttachGuard` creation will probably become safe (I'm not sure). This should be considered by other team members.

- A fact that I had missed before: `jni-rs` is probably meant to be used by native libraries that support/accelerate Java libraries/applications directly, and the "invocation" API is probably a secondary feature. However, my use case is about invoking Android Java APIs from cross-platform native applications (calling Java methods from a native thread frequently, in an opposite direction), and `java-spaghetti` is probably the better choice for me: cached method IDs can be used in a safe manner through generated bindings. The `jni-rs` documentation claims that "Calling unchecked methods with invalid arguments and/or invalid class and method descriptors may lead to segmentation fault" (which is even worse than throwing an Error that crashs the JVM), if it's true, it means that we must rely on a bindgen to use cached method IDs in a safe manner. Currently, `java-spaghetti` doesn't rely on `jni-rs`, but uses `jni-sys` directly (it might be changed in the future, though).

Besides, the TODO comment in `AttachGuard::from_owned` code confuses me:

```rust
    unsafe fn from_owned(env: *mut sys::JNIEnv) -> Self {
        Self {
            // TODO: make the JNIEnv non-transparent and read
            // `jvm.thread_attach_guard_level()` for validating that only the
            // environment at the top of the stack is ever usable for creating
            // new local references (in addition to requiring a `&mut` reference)
            env: JNIEnv::from_raw_unchecked(env),
            should_detach: true,
            level: thread_guard_level_inc()
        }
    }
```

The `JNIEnv` is already non-transparent; did you mean checking `jvm.thread_attach_guard_level()` in this method, or in `JNIEnv` methods that creates a new local reference?

wuwbobo2021:
Thanks for the invitation, I am an amateur and I can't take the responsibility of being a regular member of the `jni-rs` team maintaining an important library. I have looked at these main changes, it looks great. These soundness issues seems resolved, and the API remains "usable". Here are a few notes:

- I guess you are making `AttachGuard` creation `unsafe` because the introduced attach `level` mechanism cannot really prevent an outer `JNIEnv` from being used in an inner frame. To make it safe, at least every safe JNI method of `JNIEnv` creating a new local reference needs to check the attach level, and I guess you are not doing so because of the possible overhead.

- If local frame management can be made `unsafe`, a Java callback function will probably not need an unsafe `AttachGuard::from_unowned` call in order to use the `JNIEnv`, and `AttachGuard` creation will probably become safe (I'm not sure). This should be considered by other team members.

- A fact that I had missed before: `jni-rs` is probably meant to be used by native libraries that support/accelerate Java libraries/applications directly, and the "invocation" API is probably a secondary feature. However, my use case is about invoking Android Java APIs from cross-platform native applications (calling Java methods from a native thread frequently, in an opposite direction), and `java-spaghetti` is probably the better choice for me: cached method IDs can be used in a safe manner through generated bindings. The `jni-rs` documentation claims that "Calling unchecked methods with invalid arguments and/or invalid class and method descriptors may lead to segmentation fault" (which is even worse than throwing an Error that crashs the JVM), if it's true, it means that we must rely on a bindgen to use cached method IDs in a safe manner. Currently, `java-spaghetti` doesn't rely on `jni-rs`, but uses `jni-sys` directly (it might be changed in the future, though).

Besides, the TODO comment in `AttachGuard::from_owned` code confuses me:

```rust
    unsafe fn from_owned(env: *mut sys::JNIEnv) -> Self {
        Self {
            // TODO: make the JNIEnv non-transparent and read
            // `jvm.thread_attach_guard_level()` for validating that only the
            // environment at the top of the stack is ever usable for creating
            // new local references (in addition to requiring a `&mut` reference)
            env: JNIEnv::from_raw_unchecked(env),
            should_detach: true,
            level: thread_guard_level_inc()
        }
    }
```

The `JNIEnv` is already non-transparent; did you mean checking `jvm.thread_attach_guard_level()` in this method, or in `JNIEnv` methods that creates a new local reference?

rib:
> Thanks for the invitation, I am an amateur and I can't take the responsibility of being a regular member of the `jni-rs` team maintaining an important library.

no worries - I initially just invited you as a read-only outside contributor because I wanted to be able to add you as a reviewer (which github doesn't otherwise let me do) - I haven't just handed you the keys and made you maintainer ;-p 

> I guess you are making AttachGuard creation unsafe because the introduced attach level mechanism cannot really prevent an outer JNIEnv from being used in an inner frame. To make it safe, at least every safe JNI method of JNIEnv creating a new local reference needs to check the attach level, and I guess you are not doing so because of the possible overhead.

I don't currently see how this design (which hands out guards by-value) can be made safe, even with runtime checks for the "level". I'm not really worried about the overhead that would come from checking the level, and am still interested in adding level checks to JNIEnv APIs that take a mutable ref - I just wanted to share this work-in-progress before getting a chance to look at doing that.

Rust can't force by-value types to be immovable and to only exist on the stack - so if we return an `AttachGuard` we can't stop someone from creating as many as they like in the same scope, and then they can shuffle them around however they like and move some of them into static storage, e.g. boxing them or moving into static variables. If a guard would be moved into static storage that would then given you access to `JNIEnv` references with static lifetimes that would not be safe.

The only way I can think of keeping the attachment guards on the stack is to require an API like:

```rust
vm.attach_current_thread(|guard| {

})
```
which would run a closure that effectively lets us put a guard on the stack and only give a reference to the guard (or just directly give a reference to a `JNIEnv`).

If we would be confident that an API like above that takes a closure can be made safe that would be my preference but otherwise I'm worried about the ergonomics of an `unsafe` API like above that will lead to the closure being considered `unsafe` too.

Maybe if all APIs that take an `&mut JNIEnv` (which can return new local references associated with that env lifetime) would also `assert!` the guard level matches the top level that would be enough for us to be able to make an API that takes a closure be safe? _maybe_?

> A fact that I had missed before: jni-rs is probably meant to be used by native libraries that support/accelerate Java libraries/applications directly, and the "invocation" API is probably a secondary feature. However, my use case is about invoking Android Java APIs from cross-platform native applications (calling Java methods from a native thread frequently, in an opposite direction), and java-spaghetti is probably the better choice for me: cached method IDs can be used in a safe manner through generated bindings. 

FWIW, from my POV I personally also tend to only really care about Android, and I also care about making calls from native threads into the JVM. I'm also the maintainer (well haven't been very active in the last year) of the `android-activity` crate and that's at least one example where native method implementations are also relevant, but also with `bluey` on Android that needs native method implementations so I tend to think that it's important to support JNI in both directions when working on Android.

Also from my POV, I feel like I should point out that I wasn't involved in the early development of `jni-rs` and only became the "maintainer" after the project had been largely abandoned (a bit similar to how I haven't been able to work on the crate in the last year or so).

The more I have worked on the crate then I can't really say that I've been very happy with the design and it has often felt like a game of Whack 'a' Mole to try and make it vaguely safe.

That said - I _think_ `jni-rs` is at least a lot safer now than it was ~ 3 years ago, and I think @argv-minus-one helped a lot with that.

I would understand why you may end up preferring to just use another abstraction on `jni-sys` instead of use `jni-rs` though (I've wondered before whether `jni-rs` just has too many design issues to bother trying to fix).

> The jni-rs documentation claims that "Calling unchecked methods with invalid arguments and/or invalid class and method descriptors may lead to segmentation fault" (which is even worse than throwing an Error that crashs the JVM), if it's true, it means that we must rely on a bindgen to use cached method IDs in a safe manner. Currently, java-spaghetti doesn't rely on jni-rs, but uses jni-sys directly (it might be changed in the future, though).

Yeah, all of the _unchecked methods are something I really dislike in this crate. In practice I use them because I think it's also unacceptable that the other methods will repeatedly, re-query the class references and method IDs.

I do think these are something that should also be improved in this crate though.

Even without an external bindgen too I'd rather see a separation of method ID lookups that can be cached with a corresponding class references and then calls to those APIs shouldn't need to involve lookups.

One option here has been to do something to https://github.com/jni-rs/jni-rs/issues/137#issuecomment-1292521322

but also I've also considered a procmacro approach e.g. something like:

```rust
#[jni::class_bindings(my.package.MyClass)]
struct MyClassBindings {
    #[jni::method(myMethodA(boolean, int, java.lang.String, int, int) -> boolean)]
    method_a: jni::Binding,
    #[jni::method("ZILjava/lang/String;II)Z")]
    method_b: jni::Binding,
    #[jni::static_method(myStaticMethodC(boolean, int, java.lang.String, int, int) -> boolean)]
    method_c: jni::Binding,
}
```
(ref: https://github.com/jni-rs/jni-rs/issues/137#issuecomment-1325123080)

rib:
> If local frame management can be made unsafe, a Java callback function will probably not need an unsafe AttachGuard::from_unowned call in order to use the JNIEnv, and AttachGuard creation will probably become safe (I'm not sure). This should be considered by other team members.

With this current design then I'm not sure we can exchange `unsafe` attachments with `unsafe` local frame creation, since this design depends on the 'Safety' rules for ensuring that guards are nested and can't be reordered on the stack or moved to static storage. Without those rules then "safe" code could break the "level" tracking in thread local storage and could retain an `AttachGuard` until the `JNEnv` pointer it holds becomes invalid.

E.g. imagine a native thread that calls into JNI that then calls a native method implemented in `jni-rs`. This code could create as many guards as it likes and safely save them in ('static) thread local storage. Outside of that native method implementation the same thread could then try to use one of those saved guard to get a `JNIEnv` reference based on a `JNIEnv` pointer that could now be invalid.

Assuming one or the other needs to be `unsafe` then I tend to think it's good to keep the need for `unsafe` APIs towards the outside of a library (for example needing some unsafe code within the implementation of a native method to map to and from `jni_sys` types is probably unavoidable anyway - or as a special case when spawning a native thread that needs to use JNI).

My gut feeling is that there are probably also more reasons why thread attachments may need to be considered unsafe.

Another consideration here is that APIs like `with_local_frame` that take a closure are also especially unergonomic in Rust if they are `unsafe` because they naturally lead you to make all of the code within the closure be `unsafe` when you only really want the call itself to be considered unsafe.

Ideally, we could figure out a satisfactory way to make both be safe and _maybe_ that's still possible, but probably not with this current design.

rib:
> Besides, the TODO comment in `AttachGuard::from_owned` code confuses me

I think part of the confusion here is that I wrote that TODO note before changing `JNIEnv` to not be transparent.

But I have not yet implemented any validation of the attach guard level in methods that take a `&mut JNIEnv` yet, which is the main idea being described by this TODO note.

A `&mut JNIEnv` is currently required for any API that can return a local reference that is tied to the lifetime of that environment and so these are the only APIs where we need to assert that the lifetime corresponds to the top of the JNI stack.

If the internal `JNIEnv::from_raw_unchecked()` API took a `level` that would be stored in the `JNIEnv` then we could add a runtime check in any API that takes a `&mut JNIEnv` that `self.level == JavaVM::thread_attach_guard_level()`.

So that `TODO` comment is essentially about passing in the current `level` to `from_raw_unchecked` so we would be able to implement that check.

If we had these runtime checks then maybe we could make it "safe" to attach threads in the same way it's safe for a `RefCell` to panic for a borrow check failure.

With the current design there are also concerns with what can happen if guards would be incorrectly saved in static variables until they become invalid pointers (and the above TODO doesn't help with that) so I think we'd probably _also_ need to change the design so that thread attachments only worked with closures instead of explicit guards - but maybe that could be OK.

wuwbobo2021:
> If a guard would be moved into static storage that would then given you access to JNIEnv references with static lifetimes that would not be safe.

I agree, because of the `pub fn current_frame_env<'local>(&'local mut self) -> &'local mut JNIEnv<'local>` in `AttachGuard`.

> I think we'd probably also need to change the design so that thread attachments only worked with closures instead of explicit guards.

If you are going to move `level` into `JNIEnv` which is already made non-transparent add attach level checkers in `JNIEnv` methods that creates local references, `AttachGuard` will probably become redundant... I guess some of the `AttachGuard` methods can be moved to `JavaVM` or even a new module, then `AttachGuard` can be removed to avoid confusion. Method `current_frame_env` can be changed to `with_current_frame_env`, also taking a closure.

I had considered seperating the frame counter from the nested-attachment counter, it is probably unnecessary; the attachment counter may be increased without pushing a local frame, and the frame counter is just redundant if the nested-attachment counter can be increased while pushing a local frame.

I am busy now, please do not expect me to do important things. @argv-minus-one is a better reviewer. However, I would like to help out with `jni-rs` if he weren't to reply in two months.

rib:
> I had considered seperating the frame counter from the nested-attachment counter, it is probably unnecessary; the attachment counter may be increased without pushing a local frame, and the frame counter is just redundant if the nested-attachment counter can be increased while pushing a local frame.

Right; I'm currently just adding an `AttachGuard` within the implementation of `with_local_frame` so it's based on the same level counter.

>I am busy now, please do not expect me to do important things. @argv-minus-one is a better reviewer. However, I would like to help out with jni-rs if he weren't to reply in two months.

No worries. I'll aim to push updates here and continue iterating. Hopefully you both get some time to comment. Since you've obviously been thinking about this stuff recently, it'll be great if you will be able to keep an eye on this, but I totally understand if you end up being too busy (that's been me for the last year and a half tbh).

I'll probably try and make forward progress either way, while I have a bit of time currently, since a jni-rs release is _so_ long over due, but in any case I'll also be looking to adapt some other crates like `android-activity` and `bluey` to also try and validate any design that comes out of this.

Btw, I've just added runtime assertions to `JNIEnv` apis that return local references, and so I'm next going to experiment with having the attachment APIs be safe and take a closure.

I agree that there's a good chance that `AttachGuard` may become redundant at that point but I'm not 100% sure just yet.

rib:
For the sake of giving an update for where my iteration here has taken me, I've pushed some updates that:

- Adds a getter for the global JavaVM singleton, via `JavaVM::singleton()`
- Adds safe JavaVM APIs for attaching threads or running JNI code against the currently attached thread.

    E.g. attach the current thread and run some JNI code:
    ```rust
    JavaVM::singleton().attach_current_thread(JNIVersion::V1_4, capacity, |env| -> Result<()> {
        // JNI code using `env`
    }).unwrap();
    ```

    Or run some JNI code, assuming you know the current thread must already be attached:
    ```rust
    JavaVM::with_env(|env| -> Result<()> {
        // JNI code using `env`
    }).unwrap();
    ```
- Re-works `GlobalRef` and `WeakRef` so they are generic over `'static` reference types like `GlobalRef<JClass<'static>>` so we don't erase the inner type (the implementation of these are affected by thread attachment API changes, given how they require an attachment within their Drop implementations).

    These types are now thin wrappers around reference types that don't put the reference into a heap-allocated, ref-counted `Arc` and it's assumed that if applications really want to share global references that way they could put the `Arc` on the outside instead. 
- `AutoLocal` is a bit more ergonomic to use now, since you can convert any owned reference type into an AutoLocal with `.auto()`.

    E.g. like
    ```rust
    let klass = env.find_class(MATH_CLASS).unwrap().auto();
    ```

I need to split this work up into more digestible PRs somehow but overall I'm pretty happy with where this is going so far.

The end result should give us safe, ergonomic APIs for attaching threads and also address the safety issues around JNIEnv lifetimes.

There are still open questions about how to support native/foreign method implementations. Currently I've just updated the native methods example to use the unsafe `AttachGuard`s, and one side-effect of these changes currently is that we have to go back to returning native method references via `jni_sys` types since there's no way to get a `JNIEnv` that has an arbitrary lifetime from a native method.

argv-minus-one:
Looks good so far. 👍

One problem stands out: `extern fn`s to be called from Java need better ergonomics and safety. As of this PR in its current state, this involves a raw pointer and two `unsafe` calls.

Is it perhaps possible to define another type to wrap around the `*mut jni::sys::JNIEnv` that such functions “inherit” from the JVM? Something like:

```rust
#[repr(transparent)]
pub struct JNIEnvInherited(*mut sys::JNIEnv);

impl JNIEnvInherited {
	pub fn with_env<T>(self, closure: impl FnOnce(&mut JNIEnv) -> T) -> T {
		unsafe {
			let mut guard = AttachGuard::from_unowned(self.0);
			let mut env = JNIEnv::new(&mut guard);
			closure(&mut env)
		}
	}
}
```

```rust
#[no_mangle]
extern "system" fn Java_com_example_Example_foo(env_i: JNIEnvInherited) {
	env_i.with_env(|env: &mut JNIEnv| {
		// Do stuff with the `JNIEnv` here.
	})
}
```

Also, this method could catch unwinding panics and translate them into Java exceptions. (Otherwise, unwinding out of such a function will abort the whole program.) This is especially important now that every `JNIEnv` method can panic.

If you want to get fancy, this method could also handle regular errors and translate them into Java exceptions as well:

```rust
pub trait ToJavaException {
	fn to_java_exception<'local>(&self, env: &mut JNIEnv<'local>) -> jni::errors::Result<JThrowable<'local>>;
}

impl ToJavaException for jni::errors::Error { … }

impl JNIEnvInherited {
	pub fn with_env<T, E>(self, closure: impl FnOnce(&mut JNIEnv) -> Result<T, E>) -> T
	where
		T: Default,
		E: ToJavaException,
	{
		unsafe {
			let mut guard = AttachGuard::from_unowned(self.0);
			let mut env = JNIEnv::new(&mut guard);

			let maybe_panic = catch_unwind(AssertUnwindSafe(|| {
				let error: E = match closure(&mut env) {
					Ok(ok) => return ok,
					Err(err) => err,
				};

				match error.to_java_exception(&mut env) {
					Ok(ok) => env.throw(ok),
					// If instantiating a Java exception also fails, make a last-ditch attempt to throw a generic `java.lang.Error`.
					Err(_) => env.throw_new("java/lang/Error", error.to_string()),
				}

				T::default()
			}));

			match maybe_panic {
				Ok(ok) => ok,
				Err(panic) => {
					let panic_message: &str = {
						if let Some(s) = panic.downcast_ref::<&'static str>() {
							s
						}
						else if let Some(s) = panic.downcast_ref::<String>() {
							s
						}
						else {
							"panic in Rust code"
						}
					};
					env.throw_new("java/lang/Error", panic_message);
					T::default()
				}
			}
		}
	}
}
```

argv-minus-one:
If `JNIEnv` isn't going to be transparent any more, then there need to be big, conspicuous warnings for people migrating from previous versions, both in the changelog and the crate docs.

This change, as it stands, is a breaking change of the most severe kind: it will cause previously valid code to segfault or exhibit other undefined behavior, and the compiler won't issue so much as a warning, let alone an explanation of what needs to be done. If people don't know about the change, they'll know that their code no longer works, but won't know why.

rib:
> If `JNIEnv` isn't going to be transparent any more, then there need to be big, conspicuous warnings for people migrating from previous versions, both in the changelog and the crate docs.
> 
> This change, as it stands, is a breaking change of the most severe kind: it will cause previously valid code to segfault or exhibit other undefined behavior, and the compiler won't issue so much as a warning, let alone an explanation of what needs to be done. If people don't know about the change, they'll know that their code no longer works, but won't know why.

oh, I hadn't double checked but I was assuming the compiler was going to complain that the non-transparent `JNIEnv<'foo>` isn't FFI safe and not allow it - at least not without a loud warning - but I just tried it out and, like you say, it's a silent ABI break currently which isn't good :/

rib:
> One problem stands out: `extern fn`s to be called from Java need better ergonomics and safety. As of this PR in its current state, this involves a raw pointer and two `unsafe` calls.

Yeah, I worked on this a bit further yesterday and came up with an alternative unsafe `with_env_unowned` API that can be used something like:

```rust
#[no_mangle]
pub extern "system" fn Java_HelloWorld_hello<'caller_frame>(
    env: *mut jni::sys::JNIEnv,
    _class: JClass<'caller_frame>,
    input: JString<'caller_frame>,
) -> JString<'caller_frame> {
    unsafe {
        jni::JavaVM::with_env_unowned::<'caller_frame>(env, |env| -> jni::errors::Result<_> {
            let input: String = env
                .get_string(&input)
                .expect("Couldn't get java string!")
                .into();

            let output = env
                .new_string(format!("Hello, {}!", input))
                .expect("Couldn't create java string!");
            Ok(output)
        })
        .unwrap()
    }
}
```

It's `unsafe` to use because it effectively lets you name the lifetime for the JNI stack frame that you don't own with anything (and e.g. `'static` wouldn't be safe).

The `&JNIEnv` reference lifetime given to the closure is borrowed from an AttachGuard just like any other JNIEnv and that means you can't move the reference itself out of the closure, but it has an externally named lifetime like `&JNIEnv<'caller_frame>` so local reference that are created within the closure _can_ be safely moved out of the closure.

The implementation needs to use `transmute` to cast the lifetime but my current belief is that this isn't breaking any fundamental soundness rules with Rust and so long as the user follows the "Safety" rules such that they don't try and pass a `'static` lifetime then I think it's safe.

The API doesn't currently layer on things like catch_panic or mapping errors to exceptions etc, although yep `jni-rs` sorely needs  something to help with that.

<details><summary>This is the current implementation of `with_env_unowned`</summary>
<p>

```rust
    /// Runs a closure with a [`JNIEnv`] based on a raw [`crate::sys::JNIEnv`]
    /// pointer representing a thread attachment and JNI stack frame that was
    /// created externally.
    ///
    /// This API is specifically intended to be used within native/foreign Java
    /// method implementations in cases where you have named the lifetime for
    /// the current JNI stack frame.
    ///
    /// If you aren't implementing a native/foreign Java method then you almost
    /// certainly should not use this `unsafe` API.
    ///
    /// # Safety
    ///
    /// The given lifetime must not outlive the current JNI stack frame. For
    /// example it would _never_ be safe to use `'static`.
    ///
    pub unsafe fn with_env_unowned<'unowned, F, T, E>(
        env: *mut crate::sys::JNIEnv,
        f: F,
    ) -> std::result::Result<T, E>
    where
        F: FnOnce(&mut JNIEnv<'unowned>) -> std::result::Result<T, E>,
        E: From<Error>,
    {
        // Safety: we trust that the caller has passed a valid, non-null pointer
        let mut guard = unsafe { AttachGuard::from_owned(env) };
        guard.with_env_current_frame(|env| {
            // ☠☠☠☠☠☠☠☠☠☠☠☠☠☠  !WARNING!  ☠☠☠☠☠☠☠☠☠☠☠☠☠☠
            //
            // This is stepping into dangerous waters...
            //
            // We are casting the JNIEnv lifetime name here so that code will be
            // able to create local references that are (correctly) associated
            // with a JNI stack frame lifetime that is owned by the JVM which
            // has called our native method.
            //
            // Note that the reference itself still has the same lifetime (and
            // so the `env` reference itself can't be moved out of the closure).
            //
            // Any local reference types (like `JObject<'local>`) that can then
            // be created with this lifetime will be safely movable out of the
            // closure since they are in fact owned by the outer scope.
            //
            // Reference types like `JObject` don't ever actually borrow any
            // data, they are transparent wrappers over a `jint`, so it's safe
            // for them to technically outlive the `JNIEnv` and `AttachGuard` in
            // this case.
            //
            let unowned_env: &mut JNIEnv<'unowned> = std::mem::transmute(env);
            f(unowned_env)
        })
    }
```

</p>
</details>

rib:
> This change, as it stands, is a breaking change of the most severe kind: it will cause previously valid code to segfault or exhibit other undefined behavior, and the compiler won't issue so much as a warning, let alone an explanation of what needs to be done. If people don't know about the change, they'll know that their code no longer works, but won't know why.

So, one thing we can do here is add an empty `Drop` implementation for `JNIEnv` and then we would at least generate a warning like:
```rust
warning: `extern` fn uses type `JNIEnv<'_>`, which is not FFI-safe
  --> src/lib.rs:44:10
   |
44 |     env: JNIEnv<'caller_frame>,
   |          ^^^^^^^^^^^^^^^^^^^^^ not FFI-safe
   |
   = help: consider adding a `#[repr(C)]` or `#[repr(transparent)]` attribute to this struct
   = note: this struct has unspecified layout
   = note: `#[warn(improper_ctypes_definitions)]` on by default
```

perhaps combined with a big warning in the release notes that code implementing native Java methods should add `#![deny(improper_ctypes_definitions)]` and _maybe_ that could be OK.

otherwise maybe we could give the non-transparent JNIEnv a different path and keep the old `#[transparent]` type with a deprecation warning.

so we would have:

- `jni::sys::JNIEnv` as the raw pointer
- `jni::JNIEnv` as the `#[deprecated]` transparent type which would no longer have any associated API
- `jni::something::JNIEnv` for the new `JNIEnv`.

I guess a lot of the time the change in path wouldn't be a big ergonomic issue, for code that just uses a given `JNIEnv` reference and doesn't need to `use` anything explicitly.

rib:
One thing I'm not sure of atm is whether it's really beneficial to have a dedicated wrapper type like `JNIEnvInherited` vs just using the `jni_sys` pointer type when it comes to the `extern "system"` function signatures, where it seems semi-expected that you're in `unsafe` territory here anyway?

wuwbobo2021:
Excuse me, I looked at the previous `AttachGuard::current_frame_env()` and get confused again about why it is impossible to be safe.

> Rust can't force by-value types to be immovable and to only exist on the stack - so if we return an `AttachGuard` we can't stop someone from creating as many as they like in the same scope, and then they can shuffle them around however they like and move some of them into static storage, e.g. boxing them or moving into static variables. If a guard would be moved into static storage that would then given you access to `JNIEnv` references with static lifetimes that would not be safe.

I wonder how you could do it without `unsafe` operations. Because `AttachGuard` is `!Send` and `!Sync`, one cannot put it into some global storage like `OnceLock`; to store it "statically", I have to put it into a local key. According to <https://doc.rust-lang.org/stable/std/thread/struct.LocalKey.html>: "The `with` method yields a reference to the contained value which cannot outlive the current thread or escape the given closure." How could you get a reference to a `JNIEnv<'static>` via `pub fn current_frame_env<'local>(&'local mut self) -> &'local mut JNIEnv<'local>`?

> then they can shuffle them around however they like and move some of them into static storage, e.g. boxing them or moving into static variables.

These are definitely wrong usages, but I can't get access to a `JNIEnv` reference with the `'static` lifetime parameter through local keys, and I guess that a wrong drop order of `AttachGuard`s stored in local keys may trigger `panic`s instead of UBs.

Of course, these words probably make no sense if `AttachGuard` itself is redundant.

wuwbobo2021:
@arv-minus-one My understanding of your `JNIEnvInherited`: you've thought about the jni_sys pointer type, which may be constructed from an invalid pointer "safely" in Rust; however, such a pointer passed to an `extern fn` transparently should be considered as a valid pointer (?), and it would be safe to use this pointer instead of calling `GetEnv` again. If it is true, then something like your `JNIEnvInherited` will make sense.

rib:
> Excuse me, I looked at the previous `AttachGuard::current_frame_env()` and get confused again about why it is impossible to be safe.
> 
> > Rust can't force by-value types to be immovable and to only exist on the stack - so if we return an `AttachGuard` we can't stop someone from creating as many as they like in the same scope, and then they can shuffle them around however they like and move some of them into static storage, e.g. boxing them or moving into static variables. If a guard would be moved into static storage that would then given you access to `JNIEnv` references with static lifetimes that would not be safe.
> 
> I wonder how you could do it without `unsafe` operations. Because `AttachGuard` is `!Send` and `!Sync`, one cannot put it into some global storage like `OnceLock`; to store it "statically", I have to put it into a local key. According to https://doc.rust-lang.org/stable/std/thread/struct.LocalKey.html: "The `with` method yields a reference to the contained value which cannot outlive the current thread or escape the given closure." How could you get a reference to a `JNIEnv<'static>` via `pub fn current_frame_env<'local>(&'local mut self) -> &'local mut JNIEnv<'local>`?
> 
> > then they can shuffle them around however they like and move some of them into static storage, e.g. boxing them or moving into static variables.
> 
> These are definitely wrong usages, but I can't get access to a `JNIEnv` reference with the `'static` lifetime parameter through local keys, and I guess that a wrong drop order of `AttachGuard`s stored in local keys may trigger `panic`s instead of UBs.
> 
> Of course, these words probably make no sense if `AttachGuard` itself is redundant.

The constraints are that the guard needs to be pinned to the stack for the thread, so it can't be reordered on the stack and so it can't outlive the stack frame being unwound.

I don't think can automatically enforce either of those with the Rust type system if the guard is exposed as a value, but we can if the the guard is a hidden local variable (on the stack) that we only expose as a shared reference (which can't be used to move the value).

The guard doesn't need to be `Send/Sync` to still be possible to put onto the heap instead of the stack, and be given a `'static` lifetime.

For example, this code compiles:

```rust
use jni::{AttachGuard, JNIVersion};

pub fn jvm() -> &'static jni::JavaVM {
    use std::sync::OnceLock;
    static JAVA_VM: OnceLock<jni::JavaVM> = OnceLock::new();
    JAVA_VM.get_or_init(|| {
        let args = jni::InitArgsBuilder::new()
            .option("-Djavax.net.debug=all")
            .option("-verbose:jni")
            .build()
            .unwrap();
        jni::JavaVM::new(args).unwrap()
    })
}

fn main() {
    let env_ptr = jvm().attach_current_thread(JNIVersion::V1_4, |env| -> jni::errors::Result<_> {
        Ok(env.get_raw())
    }).unwrap();

    let guard = unsafe { AttachGuard::from_unowned(env_ptr) };
    let static_guard = Box::new(guard);
    let static_guard: &'static mut AttachGuard = Box::leak(static_guard);

    let _static_env: jni::JNIEnv<'static>  = unsafe { jni::JNIEnv::new(static_guard) };
}
```

This is using `unsafe` code based on this branch, but the bit of interest that shows how you could move the guard off the stack doesn't involve `unsafe` code, you can Box and leak and give yourself an arbitrary (including `static`) lifetime for a value.

argv-minus-one:
@rib @wuwbobo2021 Yes, the idea of `JNIEnvInherited` is that the `*mut JNIEnv` pointer provided by the JVM is guaranteed to be valid, so using it should not be `unsafe`.

Of course, there is no guarantee that the parameters to the `extern fn` are correct. Besides the `*mut JNIEnv`, you can also get undefined behavior if a `JObject` or the like is in the wrong place. This is why the 2024 edition of Rust requires `no_mangle` to be written as `unsafe(no_mangle)`.

Although I should have included a lifetime parameter on `JNIEnvInherited`, or else it is still `unsafe` to use because the local frame lifetime could be set incorrectly, as mentioned in https://github.com/jni-rs/jni-rs/pull/570#issuecomment-2743164291. Something like:

```rust
#[repr(transparent)]
pub struct JNIEnvInherited {
	ptr: *mut sys::JNIEnv,
	_lifetime: PhantomData<&'local ()>,
}

impl<'local> JNIEnvInherited<'local> {
	pub fn with_env<T>(self, closure: impl FnOnce(&mut JNIEnv<'local>) -> T) -> T {
		unsafe {
			let mut guard = AttachGuard::from_unowned(self.0);
			let mut env = JNIEnv::new(&mut guard);
			closure(transmute(&mut env))
		}
	}
}
```

```rust
#[unsafe(no_mangle)]
extern "system" fn Java_com_example_Example_foo<'local>(
	env_i: JNIEnvInherited<'local>,
	obj: JObject<'local>,
) -> JObject<'local> {
	env_i.with_env(|env: &mut JNIEnv<'local>| -> jni::errors::Result<JObject<'local>> {
		todo!()
	})
}
```

rib:
Okey, I've just come back to take another pass at this...

I've added a `JNIEnvUnowned` type that can be used with native methods (I went with "Unowned" instead of "Inherited" since I felt that aligned with other references to "owned" and "unowned" with these changes).

`JNIEnvUnowned::with_env()` replaces `JavaVM::with_env_unowned()` and is considered safe, based on the reasoning that:
1. You can't safely construct a `JNIEnvUnowned` manually
2. If you acquire a `JNIEnvUnowned` value as a native method argument, you are separately responsible for ensuring your FFI function signature is correct but by the time you call `.with_env()` we will assume it contains a valid, non-null pointer.

`JNIEnvUnowned::with_env()` also wraps the given closure in a `catch_unwind` to help ensure that native method implementations don't allow Rust code to unwind over the FFI boundary. There is an alternative `JNIEnvUnowned::with_env_no_catch()` in case you want to handle the `catch_unwind` yourself.

Since there's basically no way to make it a compiler error to use an existing Rust type in an C/system ABI FFI interface, I've instead opted to move the real `JNIEnv` to `jni::env::JNIEnv` and have made `jni::JNIEnv` into a deprecated alias for `JNIEnvUnowned` that will show a very verbose warning like this:

```
warning: use of deprecated type alias `jni::JNIEnv`: Since 0.22, `JNIEnv` is not an FFI safe pointer wrapper any more.

         For safety, `jni::JNIEnv` is now an alias for `JNIEnvUnowned` (which is FFI safe).

         The real `JNIEnv` is now exported as `jni::env::JNIEnv` _BUT_ it's rare that you should need to refer to it directly.

         Use `JNIEnvUnowned` to capture a `JNIEnv` pointer in native methods, like:
             `pub extern "system" fn Java_HelloWorld_hello<'frame>(unowned_env: JNIEnvUnowned<'frame>, ...)`
         Then use `unowned_env.with_env()` to upgrade it to a `JNIEnv` reference.

         Most of the time you should temporarily acquire a `JNIEnv` reference using:
           - `JavaVM::attach_current_thread` (preferred) or `JavaVM::attach_current_thread_for_scope`
           - `JavaVM::with_env` (if certain that the thread is already attached)
```

It's tricky to know the best way to handle this but this way it should be hard to avoid seeing this message and it errs on the side of leaving `jni::JNIEnv` as being FFI safe.

Since you will have to explicitly modify your code to avoid the deprecation warning and either refer to `JNIEnvUnowned` or `jni::env::JNIEnv` then _hopefully_ that's enough to help avoid crashes from code trying to use the non-transparent `JNIEnv` to capture `jni::sys::JNIEnv` pointers in native methods. :crossed_fingers:

rib:
Okey, I've done a bunch of clean up, split a few things out into separate PRs (#595 and #596) and done a big re-write of the PR description with lots of details about the design.

It's still a pretty big change but it's hopefully in a good enough state to land. Since quite a lot of the boilerplate changes are just indentations, it can help to hide whitespace changes when looking at the diffs.

To test the changes, I've locally updated `android-activity` and my Bluetooth LE crate `bluey` to work with the new APIs and didn't hit any problems.

I appreciate you may not have time to look at this but it would be great if you get a chance @argv-minus-one.

Also @wuwbobo2021, since you looked at a lot of the details here, it would be great if you get a chance to take a look. 

Since you last looked, the main difference I think is that there is now an FFI-safe `JNIEnvUnowned`, so a native method can now look like:

```rust
pub extern "system" fn Java_HelloWorld_hello<'caller>(unowned_env: JNIEnvUnowned<'caller>,
                                                      class: JClass<'caller>,
                                                      input: JString<'caller>)
-> JString<'caller> {
    unowned_env.with_env(|env| -> Result<_, jni::errors::Error> {
        let input: String =
            env.get_string(&input).expect("Couldn't get java string!").into();
        let output = env.new_string(format!("Hello, {}!", input))
            .expect("Couldn't create java string!");
        Ok(output)
    }).unwrap_or_else(|err| {
        eprintln!("Error occurred: {}", err);
        Default::default()
    })
}
```
(which will also handle wrapping your code in `catch_unwind` so a panic won't unwind over an FFI boundary)

wuwbobo2021:
> `JNIEnvUnowned::with_env()` also wraps the given closure in a `catch_unwind` to help ensure that native method implementations don't allow Rust code to unwind over the FFI boundary.

I'm a bit confused with this issue. Documentation of `std::panic::catch_unwind` says:

> Rust functions that are expected to be called from foreign code that does not support unwinding (such as C compiled with `-fno-exceptions`) should be defined using `extern "C"`, which ensures that if the Rust code panics, it is automatically caught and the process is aborted. If this is the desired behavior, it is not necessary to use `catch_unwind` explicitly. This function should instead be used when more graceful error-handling is needed.

(Note: `extern "system"` usually means `extern "C"`, see <https://doc.rust-lang.org/nomicon/ffi.html#foreign-calling-conventions>. probably `extern "stdcall"` also prevents unwinding into the underlying JVM stack...)

I can't make clear of the extra safety introduced by doing `catch_unwind` and `resume_unwind` manually here. If I had missed something, it would have meant that the `jni` 0.21 example, `java-spaghetti` proxies and `jni-min-helper` were all promoting unsound usages.

Anyway, thanks for your efforts of building such a crate.

Added:

I did a test on JDK 21 (Linux), running a java program calling a `jni-rs` (release 0.21.1) based native method function marked with `no_mangle` and `extern "C"`.

`panic("rust panic")` and `let e = catch_unwind(AssertUnwindSafe(|| panic!("rust panic") )).unwrap_err(); resume_unwind(e);` produced the same output:

```
thread '<unnamed>' panicked at src/lib.rs:24:46:
rust panic
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
fatal runtime error: failed to initiate panic, error 5, aborting
Aborted                    (core dumped) java JniTest

```

rib:
> > `JNIEnvUnowned::with_env()` also wraps the given closure in a `catch_unwind` to help ensure that native method implementations don't allow Rust code to unwind over the FFI boundary.
> 
> I'm a bit confused with this issue. Documentation of `std::panic::catch_unwind` says:
> 
> > Rust functions that are expected to be called from foreign code that does not support unwinding (such as C compiled with `-fno-exceptions`) should be defined using `extern "C"`, which ensures that if the Rust code panics, it is automatically caught and the process is aborted. If this is the desired behavior, it is not necessary to use `catch_unwind` explicitly. This function should instead be used when more graceful error-handling is needed.
> 
> (Note: `extern "system"` usually means `extern "C"`, see https://doc.rust-lang.org/nomicon/ffi.html#foreign-calling-conventions. probably `extern "stdcall"` also prevents unwinding into the underlying JVM stack...)
> 

I think historically it was considered "undefined behaviour" to allow Rust code to unwind across an FFI boundary but think perhaps the rustonomicon was updated when ABIs like "C-unwind" were added. Even though it may have been considered 'undefined' I think in practice Rust has triggered an abort and I think now that has become the defined behaviour, which you can now control by using ABIs like "C-unwind". I might not be quite correct, but that's my vague recollection atm.

> I can't make clear of the extra safety introduced by doing `catch_unwind` and `resume_unwind` manually here. If I had missed something, it would have meant that the `jni` 0.21 example, `java-spaghetti` proxies and `jni-min-helper` were all promoting unsound usages.

Unfortunately `jni-rs` 0.21 currently does very little to stop you shooting yourself in the foot while implementing native methods. It's left up to you to make sure you use catch_unwind if you don't want your program to crash/abort on panic.

This has been quite a long standing pain point. E.g. here's some older discussion around handling catch_unwind manually: https://github.com/jni-rs/jni-rs/issues/432

I guess since it has now become well defined (since 1.71 I think) that Rust aborts your program, it can be a valid choice to just accept that behaviour. I think I still have a mindset that says this should preferably be handled intentionally.

> I did a test on JDK 21 (Linux), running a java program calling a `jni-rs` (release 0.21.1) based native method function marked with `no_mangle` and `extern "C"`.
> 
> `panic("rust panic")` and `let e = catch_unwind(AssertUnwindSafe(|| panic!("rust panic") )).unwrap_err(); resume_unwind(e);` produced the same output:

I'm not entirely sure what you were aiming to check here because it looks like you're testing the behaviour of panic unwinding in both cases. If you call resume_unwind with the payload from `catch_unwind()` then that's essentially the same as if you didn't do the `catch_unwind()`? In this context the expectation would be that if you catch a panic then you would not want to allow that to unwind into the JVM and you would instead either log an error or return an error status to Java or throw an exception for Java to represent the panic.

wuwbobo2021:
Sorry for my stupid test posted above, I used to think that `catch_unwind` has to be paired with `resume_unwind`. I've remembered it wrongly; however, `std::panic::catch_unwind` tells "be careful in how you drop the result of this function. If it is Err, it contains the panic payload, and dropping that may in turn panic!". In the current `JNIEnvUnowned::with_env`, the panicking payload is probably dropped at the line `Err(_) => "non-string panic payload".to_string(),` without calling `std::mem::forget` if it's not a `String`. I'm not sure if it always prevent the aborting.

rib:
> If it is Err, it contains the panic payload, and dropping that may in turn panic!"

hmmm, yeah that's a good point - it'd probably be good to ensure that any jni-rs examples for how to implement native methods are careful to avoid dropping any payload that's not resumed.

rib:
> If it is Err, it contains the panic payload, and dropping that may in turn panic!". In the current `JNIEnvUnowned::with_env`, the panicking payload is probably dropped at the line `Err(_) => "non-string panic payload".to_string(),` without calling `std::mem::forget` if it's not a `String`. I'm not sure if it always prevent the aborting.

To be even more cautious here, this has been updated to use `catch_unwind` around an explicit `drop(payload)` (just for non-string payloads) with a fallback that will `forget()` the payload instead.

It will also log an error if the fallback path is hit which could technically itself cause a panic but I think it's reasonable enough to try and log an error here and if that does panic I guess you're just out of luck and will hit an abort.

rib:
Okey, I've gone ahead and landed these changes while I'm happy enough with the smoke testing I've done and am convinced this is a worthwhile re-design of the thread attachment APIs that should make `jni-rs` safer.

If anyone does end up looking through the details here, I'd still be more than happy to hear any retrospective review feedback or comments, which can still be addressed in follow up PRs.

Given the significance of these changes it will probably make sense to have a `docs/0.22-MIGRATION.md` guide like we had for the 0.21 release but that can hopefully be done in a follow up closer to the 0.22 release.

Just for reference here: There are a notable number of `FIXMEs` added due to JNIEnv APIs that take a `&mut self` when they shouldn't need to (they don't return any local references). These are currently due to to `Desc::lookup` needing a `&mut JNIEnv` combined with the fact that `Desc` is not flexible enough to be usable within `JavaVM::with_env_current_frame` in these methods. My current preference here is to investigate removing the `Desc` trait since I don't personally like its complexity and the hidden costs associated with `Desc::lookups`: https://github.com/jni-rs/jni-rs/issues/594

As follow ups to this I'm tempted to consider some API/type renames before release:
E.g.
- Maybe `JavaVM::with_env_top_frame()` would be clearer than `JavaVM::with_env_current_frame()`?
- Renaming `JNIEnv` -> `Env` and `JNIEnvUnowned` -> `EnvUnowned` would further help avoid any risk of mistakes with using the new (non-transparent) `JNIEnv/Env` type as an FFI argument for capturing `jni_sys::JNIEnv` pointers for native methods. (I also think the current `JNI` prefix feels redundant in a "jni" Rust crate.)

but think it's OK to handle those in follow ups.

PR #573, opened 2025-03-20
Improve JNIEnv lifetime in with_local_frame

elichai:
This tied the `JNIEnv` lifetime with the lifetime of the `JniEnv` itself instead of the `&self` lifetime

### DIFF
--- a/src/wrapper/jnienv.rs
+++ b/src/wrapper/jnienv.rs
@@ -1033,7 +1033,7 @@ impl<'local> JNIEnv<'local> {
     /// [`GlobalRef`] / [`Self::make_global`].
     pub fn with_local_frame<F, T, E>(&mut self, capacity: i32, f: F) -> std::result::Result<T, E>
     where
-        F: FnOnce(&mut JNIEnv) -> std::result::Result<T, E>,
+        F: FnOnce(&mut JNIEnv<'local>) -> std::result::Result<T, E>,
         E: From<Error>,
     {
         unsafe {


rib:
For reference here I'm working on some fairly significant changes related to thread attachment safety in `jn-rs`, also relating to JNIEnv lifetime management: https://github.com/jni-rs/jni-rs/pull/570

src/wrapper/jnienv.rs
```diff
@@ -1033,7 +1033,7 @@ impl<'local> JNIEnv<'local> {
     /// [`GlobalRef`] / [`Self::make_global`].
     pub fn with_local_frame<F, T, E>(&mut self, capacity: i32, f: F) -> std::result::Result<T, E>
     where
-        F: FnOnce(&mut JNIEnv) -> std::result::Result<T, E>,
+        F: FnOnce(&mut JNIEnv<'local>) -> std::result::Result<T, E>,
```
rib, line 1036: It's very important that the `JNIEnv` that is passed to `f` must not have the same lifetime as the outer `JNIEnv` so I don't think this change is safe.

The purpose of the `JNIEnv` lifetime is to represent the JNI local stack frame that local references belong to and since this API creates a new JNI stack frame (via `PushLocalFrame`) then the `JNIEnv` needs to have a new, unique lifetime, otherwise it would be possible to move a local reference (such as `JObject<'local>` or `JString<'local>` out of the given closure back to the caller, where they would effectively become invalid pointers.

rib: E.g. I guess this change would let you do something like:

```rust
let mut outer_ref: Option<JString<'outer>> = None;
env.with_local_frame(5, |env| {
    let local_ref = env.new_string("foo").unwrap();
    outer_ref.replace(local_ref);
    Ok(())
}).unwrap();
```

and then when `with_local_frame` returns if you would try and access `outer_ref` it would contain an invalid pointer for a local reference that was freed when `PopLocalFrame` was called within `with_local_frame()`.

rib:
Can you maybe clarify what problem you were trying to solve with this change?

If you are trying to return a local reference from inside the `with_local_frame` closure to the caller, you may be able to use `with_local_frame_returning_local`.

rib:
Oh and coincidentally, considering the brief discussion in https://github.com/jni-rs/jni-rs/pull/561, you might be interested in the `GlobalRef` changes I've drafted in #570

elichai:
@rib Thanks for the quick replies!
I'll look at #570 as it looks very exciting and very relevant to my code :) (if you release an RC with this at some point I'll give it a try to integrate it into our codebase and give some feedback if you want :) )
The reason I wanted it is because it gave me a hard time specifying the correct lifetimes in wrapper traits calling generic functions, I'll try to make some minimal reproducible, but it might be that I need to do some more HRBTs here and there :)

If it's a fully fresh env, shouldn't it's lifetime be defined via `for<'other>  F: FnOnce(&mut JNIEnv<'other>) -> std::result::Result<T, E>,`? that way it won't be inferred as related to `&self`?

rib:
> If it's a fully fresh env, shouldn't it's lifetime be defined via `for<'other> F: FnOnce(&mut JNIEnv<'other>) -> std::result::Result<T, E>,`? that way it won't be inferred as related to `&self`?

I _think_ that's effectively what we have, even though we elide the explicit `for<'foo>` in the `where` bounds.

E.g. looking at it through the lense of rust-analyzer in vscode, I currently see:

image

rib:
> > If it's a fully fresh env, shouldn't it's lifetime be defined via `for<'other> F: FnOnce(&mut JNIEnv<'other>) -> std::result::Result<T, E>,`? that way it won't be inferred as related to `&self`?
> 
> I _think_ that's effectively what we have, even though we elide the explicit `for<'foo>` in the `where` bounds.
> 
> E.g. looking at it through the lense of rust-analyzer in vscode, I currently see:
> 
> image

well actually that doesn't exactly clarify the implied lifetime for the `JNIEnv` lifetime parameter, only that the lifetime of the reference will invariant with respect to `&self`.

I think this becomes clearer in #570 where a `JNIEnv` explicitly borrows from an `AttachGuard` that represents a JNI stack frame, and in the case of `with_local_frame` it's clear the implementation allocates a new AttachGuard when it calls PushLocalFrame and then that guard value is borrowed to create a new JNIEnv that's completely decoupled from the `&self` env.

maybe it would be good to explicitly specify that the closure gets a reference like `&'new_local JNIEnv<'new_local>`

elichai:
Very interesting, I did not know HRBTs could be elided 😅
I'll dive into #570 I love the lifetime improvements :), I already have a bunch of abstraction in our own code to make jni building "safer" and any way that allows me to improve it is great :)
(which is both the nice high level types and proper lifetime tracking)

I'll close this PR as I now understand it is indeed wrong, I did not know about the stack VS JNIEnv differences before

PR #578, opened 2025-05-25
Use SPDX license format

atouchet:
## Overview

Use SPDX license expression for license field in Cargo.toml.


rib (approved):
Thanks, this looks good to me

PR #593, opened 2025-08-18
Remove JavaVM::attach_current_thread_as_daemon

rib:
The thing that sets "daemon" thread attachments apart from permanent
thread attachments is that `JavaVM:destroy()` doesn't wait for them to
finish.

Or in other words, daemon threads are allowed to continue running even
though the VM's state is undefined.

Using JavaVM::destroy() is very unsafe, owing to how poorly defined the
VM and program state become after being called.

There's nothing practical we can do make it safe to allow jni-rs to be
used in any way after the VM state has become undefined.

If an application were to have any special case where it could somehow
guarantee that the jni-rs API won't be touched after the VM has been
destroyed (including leaking all jni types to ensure no Drop
implementation could run) then this would be `unsafe` enough to justify
using `jni_sys` directly.

This simply removes `JavaVM::attach_current_thread_as_daemon` so jni-rs
applications can't unwittingly get themselves into this very unsafe
state.

Fixes: #469


tisonkun:
Hi @rib!

In OpenDAL, we currently rely on this method to set up thread-local `JNIEnv` so that async rust code can callback properly (ref https://github.com/apache/opendal/pull/7000)

After `attach_current_thread_as_daemon` being removed, is there alternative to achieve the same functionality?

tisonkun:
Perhaps

```rust
        .on_thread_start({
            let vm = vm.clone();
            move || {
                ENV.with(|cell| {
                      vm.attach_current_thread(|| {
                         // further simplify with AttachConfig to set the thread name directly
                         set_current_thread_name(&mut env).expect("current thread name has been set above");

                         *cell.borrow_mut() = Some(env.get_raw());
                       })
                       .expect("attach thread must succeed");
                })
            }
        })
        .on_thread_stop(move || {
            ENV.with(|cell| {
                *cell.borrow_mut() = None;
            });
            
            vm.detach_current_thread().expect("how to handle the error here?");
        })
```

rib:
> Hi @rib!
> 
> In OpenDAL, we currently rely on this method to set up thread-local `JNIEnv` so that async rust code can callback properly (ref [apache/opendal#7000](https://github.com/apache/opendal/pull/7000))
> 
> After `attach_current_thread_as_daemon` being removed, is there alternative to achieve the same functionality?

As far as I can tell, I dont think your use case specifically needed a daemon thread attachment.

With the 0.21 api then I guess you would be best off creating a "permanent" attachment, that's not a daemon attachment. The semantics of daemon threads only matter in use cases where you need to be able to call JavaVM::destroy() without waiting for these threads to finish (which is very unlikely to be a requirement you have in the context of async + tokio). In this very general purpose case then it would probably be very unsafe to allow these Tokio threads to continue running if JavaVM::destroy were called. 

With jni 0.22 it still supports permanent attachments its just that the API doesn't specifically call them "permanent".

In 0.22 it really tries to constrain your JNI usage to closures that run with an AttachGuard on the stack (hidden) as part of a design that allows thread attachment to be safe, but it does also expose lower- level APIs so you have good control for tricky use cases.

Since attach_current_thread() in 0.22 is similar to creating a "permanent" attachment via 0.21 (which will use TLS to auto detach) then your snippet of code looks like one plausible way you could adapt. You could perhaps skip the explicit detach though and just rely the thread automatically getting detached via the TLS guard.

The use of a cell to try and save and pass around an Env pointer does worry me a bit and I would definitely want to try and avoid if possible to stick with safe APIs. 

The attach_current_thread api in master will pass your closure an env reference and attach_current_thread is cheap if the thread is already attached, so it would generally be recommended to wrap any use of JNI within an attach_current_thread closure.

In one project I have that uses jni and async with tokio, I used to do somthing very similar to OpenDAL with hooking into on_thread_start/stop but with 0.22 i just removed those hooks and now I just make sure to wrap code using JNI with an attach_current_thread closure. 

The first call to attach_current_thread will create a permanent attachments that will automatically detach when the thread exits and subsequent calls on the same thread will be cheap if the thread is still attached. It's arguable safer too because it can be more resilient to situations where some other jni crate or library detaches the thread while its not being used for JNI. This approach also only lazily attaches tokio threads that are actually used for JNI, which might not necessarily end up being all of them. It also doesn't rely on special-case Tokio runtime integration so code can can be written in a way that doesn't care whether its running in a tokio thread pool or some other manually spawned thread.

If you do want to stick with the on_thread_start hook, then in jni 0.22 you could look at the lower level (but unsafe) attach_current_thread_guard() API which can set up a permanent attachment (auto detaches via TLS) and you could extract the raw env pointer from the returned guard and assume that nothing else will detach the thread from JNI. I would still probably not recommend saving the raw pointer anywhere though and you might instead want to use get_env_attachment when you need it, which will at least recognise if your thread has become detached. If all your JNI code is going to assume the current thread is attached, you could use with_top_local_frame to avoid the unsafe get_env_attachment API. Essentially it should be redundant to ever save and pass around the raw env pointer because the only places where its safe to use the pointer can also query it when they need it.

Issue #571, opened 2025-03-15
`GlobalRef` shouldn't erase the reference type


rib:
`GlobalRef` should be generic over types implementing `Into<JObject>` like `JObject`, `JString` or `JClass` so that it can implement `Deref` and `AsRef` similar to `AutoLocal`.

argv-minus-one:
Note that this issue also affects `WeakRef`.

rib:
> Note that this issue also affects `WeakRef`.

Thanks, yup. I've just updated the title to reflect that.

For reference here, draft PR #571 currently includes draft changes for `GlobalRef` and `WeakRef` to address this issue.

PR #571 adds an `ObjectRef`trait with a generic `GlobalKind` associated type that makes it possible to map a reference type like `JString<'local>` into `JString<'static>`. This can then be used to construct a `GlobalRef<T>` (or `WeakRef`) with a type that's equivalent to some local-reference type except it has a static lifetime and is `Send + Sync`.

PR #596, opened 2025-08-22
Global/WeakRef: simpler reference wrappers that don't erase types

rib:
A `Global/WeakRef` is now a transparent wrapper over a 'static reference type like `JOBject`, `JClass` or `JString` that is itself a transparent wrapper over a `jni_sys::jobject` pointer.

Both are now parameterized, like `GlobalRef<JClass<'static>>` to preserve the type of reference that a Global or Weak reference represents.

E.g. a `GlobalRef<JClass<'static>>` will now deref to a `&JClass` and can be more easily used with APIs that require a `JClass` reference, or some other specific type of reference.

This introduces a `JObjectRef` trait that deals with mapping the Rust type for a local reference (like `JObject<'local>`) into a corresponding `'static` type that can be used when creating a corresponding Global/Weak reference.

Since we can guarantee that `JavaVM::singleton()` has been initialized at the point that `new_global_ref` or `new_weak_ref` is called we don't need to save a vm pointer internally and can assume it's safe to rely on `JavaVM::singleton()` in the `Drop` implementations (in case a reference is dropped from detached thread).

These types no longer implement `Clone`, since they no longer wrap a heap-allocated Arc around one reference and so copying the reference would involve a JNI call. Although we _could_ potentially use JNI from Clone (Since we can use JavaVM::singleton()) it didn't seem good that this could lead to unpredictable/hidden thread attachments, or potentially trigger some other JNI error that can't be propagated.

Both types now implement `Default`, which correspond to `null()` references.

### Definition of Done

Fixes: #571

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - `JavaStr`, `JNIStr`, and `JNIString` have several new methods and traits, most notably a `to_str` method that converts to a regular Rust string. ([#510](https://github.com/jni-rs/jni-rs/issues/510) / [#512](https://github.com/jni-rs/jni-rs/pull/512))
 - Added dependency on `once_cell` version `1.19.0` for lazy initialization in `JNIEnv::get_string`.
 - `JavaVM::singleton()` lets you acquire the `JavaVM` for the process when you know that the `JavaVM` singleton has been initialized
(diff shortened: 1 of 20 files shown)


rib:
Heya - on the off chance that you might get a moment to skim this @argv-minus-one, I've added you as a reviewer, but apologies if I have already gone ahead and merged by the time you see the request, while I'm hoping to make progress towards landing #570.

In either case though please feel free to give feedback retrospectively if merged (since it can still be addressed in a follow up if needs be)

PR #698, opened 2025-11-27
jni-macros: add `native_method!` macro

rib:
This makes it possible to construct an `env::NativeMethod` with a guarantee that the function pointer matches the JNI signature.

The macro has a long-form `prop = <val>,` config syntax and a short-form signature syntax for common cases.

The signature parsing is based on the syntax introduced for `jni_sig!`

For example:

```
let method = native_method! {
    static extern fn MyRustType::my_native_method(arg: jint) -> bool
};
```

In this case `my_native_method` is mapped to "myNativeMethod" as the Java method name. The signature would be "(I)Z" and the implementation function would be `MyRustType::my_native_method`. If it was an instance method then `MyRustType` would also be used for the `this` argument.

**Note** Although this initially looks like a pretty large PR, the majority of the additions are documentation, examples and tests and it wouldn't really make much sense to split those out into separate PRs.

### Definition of Done

    (actually there are some TODO comments for examples/tests that should use `jni_sig!()` instead of `jni_str!()` after landing planned changes for method call APIs to require a pre-parsed signature instead of a string literal - but these aren't TODOs for _this_ feature)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -101,8 +101,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 #### Macros
 
 - The `#[jni_mangle()]` attribute proc macro can export an `extern "system"` native method with a mangled name like "Java_com_example_myMethod" so it can be automatically resolved within a shared library by the JVM ([#693](https://github.com/jni-rs/jni-rs/pull/693))
-- The `jni_str!` and `jni_cstr!` macros can encode a MUTF-8 `&'static JNIStr` or `&' static CStr` at compile time with full unicode support.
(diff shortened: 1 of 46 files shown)


rib:
Note: CI has been getting stuck on Windows tests quite often recently and I gave up waiting for this PR to pass tests on Windows (https://github.com/jni-rs/jni-rs/actions/runs/20099832737/job/57667475308?pr=698)

I believe now that the Windows CI gets stuck due to this deadlock: https://github.com/jni-rs/jni-rs/issues/701

My suspicion was somewhat confirmed after cherry-picking a workaround for #701 onto this branch which passes Windows CI: https://github.com/jni-rs/jni-rs/actions/runs/20100736639

PR #712, opened 2025-12-10
jni-macros: add bind_java_type macro

rib:
This macro generates a complete `Reference` type binding for a Java class, providing:

- A Rust wrapper type that implements the `Reference` trait
- An API struct that caches the class reference and all method/field IDs
- Type-safe constructor, method, and field accessors
- A trait-based interface for implementing native methods
- Automatic native method registration with the JVM
- Optional JNI export symbols for native methods

Shorthand usage of the macro looks like:

```
bind_java_type! { RustType => test.JavaType }
```

Typical usage of the macro looks like:

```
    bind_java_type! {
        rust_type = RustType,
        java_type = test.JavaType,
        is_instance_of {
            collection: JCollection,
        },
        type_map {
            OtherRustType => test.OtherJavaType
        },
        constructors {
            fn new(),
            fn new_with_value(value: jint),
        },
        methods {
            static fn get_name() -> JString,
            fn set_value(val: jint) -> void,
            fn do_something(arg: OtherRustType) -> jint[],
        },
        fields {
            static val: jint
            count: jint
        },
        native_methods {
            fn my_native_method(arg: JString) -> bool
        }
    }

    impl RustTypeNativeInterface for RustTypeAPI {
        type Error = jni::errors::Error;
        fn my_native_method<'local>(env: &mut Env, arg: JString<'local>)
            -> Result<bool, Self::Error>
        {
            Ok(true)
        }
    }
```

**Note:** Although this _is_ a pretty huge PR, it's worth noting that the majority of the added code is documentation, tests and examples, and not the macro implementation itself. Splitting those out wouldn't really be good because we want to test the implementation thoroughly via CI, and the documentation is tightly coupled to the implementation too (including inline examples).

### Definition of Done


rib:
Although there are some follow ups I might like to make to continue improving the documentation and keep adding more tests, I think this is in a pretty good shape already so I'll go ahead and land so that I can unblock follow up PRs to start using this for the built-in Reference type bindings.

While there aren't currently any other active contributors to the `jni` crate who I can ask to review this i'd certainly welcome async feedback from anyone that starts to experiment with using this!

Issue #701, opened 2025-12-05
Deadlock due to DetachCurrentThread @ Windows


jerrinot:
Hello, there appears to be a design(?) issue with the thread-local destructor/Drop-driven `DetachCurrentThread`.

On Windows, this sequence of events can trigger a deadlock:
1. A native thread, previously attached to a JVM, is stopped
2. Windows kernel calls the TLS destructor - This happens while [holding Windows Loader Lock](https://doc.rust-lang.org/stable/std/thread/struct.LocalKey.html#synchronization-in-thread-local-destructors)
3. jni-rs `InternalAttachGuard` calls `DetachCurrentThread`. This involves a thread state transition from Native to VM
4. Concurrently, there is a new Java thread being started by some other thread
5. During the Native -> VM transition, the native thread is trapped in a JVM safepoint. While holding the loader lock!
6. The Java thread created at step no. 4 is already runnable from JVM's point of view, and the JVM expects this thread to arrive at the safepoint. But Windows won't execute user code on this thread until it manages to acquire the loader lock. Which is held by the Rust thread that is trapped at safepoint. This means the Java thread cannot make any progress and certainly cannot reach the safepoint. JVM won't release the native thread until all threads are at safepoint -> Deadlock

Evidence. 

1. Stacktrace showing a thread trapped at safepoint while running a TLS destructor:
<img width="3951" height="799" alt="Image" src="https://github.com/user-attachments/assets/46235356-d069-4cd5-aac1-1adb2ecb35d9" />

2. Safepoint timeout:
```
[319.638s][warning][safepoint] 
[319.638s][warning][safepoint] # SafepointSynchronize::begin: Timeout detected:
[319.638s][warning][safepoint] # SafepointSynchronize::begin: Timed out while spinning to reach a safepoint.
[319.638s][warning][safepoint] # SafepointSynchronize::begin: Threads which did not reach the safepoint:
[319.638s][warning][safepoint] # "Thread-9951" #10533 daemon prio=5 os_prio=0 cpu=0.00ms elapsed=60.01s tid=0x00000147a271d5b0 nid=0x1bb8 runnable  [0x0000000000000000]
[319.638s][warning][safepoint]    java.lang.Thread.State: RUNNABLE
[319.638s][warning][safepoint] 
[319.638s][warning][safepoint] # SafepointSynchronize::begin: (End of list)
```

Notice how the thread which failed to arrive at the safepoint has:
1. no CPU time
[...]

PR #718, opened 2025-12-16
Windows: Use Fiber Local Storage to avoid loader lock deadlock

rib:
This re-works how we track permanent thread attachments on Windows to use fiber local storage, instead of thread local storage, so that the callback we get when the thread exits does not hold the system loader lock.

This avoids deadlocks that can happen when Java threads are being spawned at the same time that native threads are being detached via TLS destructors because DetachCurrentThread can't progress until all threads are at a safepoint but new Java threads can't run, and reach a safepoint, until the system loader lock is acquired.

It's worth noting that this does _not_ attempt to generally support the use of Windows Fibers with JNI. JNI is fundamentally designed around thread-local state and there would be too many issues with attempting to support fibers + JNI generally (e.g. how to manage the local frame stack or pending exceptions).

From a safety pov the implementation assumes:
- Fibers may never be switched while there are active AttachGuards
- Fibers my never be deleted while there are active AttachGuards.

So hypothetically a Windows application can could utilize fibers and not break those invariants, but in practice we expect that no application using Rust + JNI ever touches the win32 fibers API so they never have to consider the possibility that there is ever more than one fiber per OS thread.

Since the changed depends on additional windows_sys APIs, this updates `jni-gen-windows-bindings/bindings.config` so we continue to avoid having `windows_sys` as a crate dependency. This also tweaks how the generated code is imported as a `crate::windows_sys` module (affecting char_encoding_windows.rs).

For debugging, this PR adds a build.rs that makes it possible to forcibly revert to using thread local storage on Windows by setting the `_JNI_WINDOWS_FORCE_USE_TLS` environment variable.

If I test this against the reproduction repo here: https://github.com/rib/jni-rs-issue-701-repo/tree/rib/wip/windows-fls-alloc then I can reliably reproduce a deadlock when using TLS via `_JNI_WINDOWS_FORCE_USE_TLS=1` and I get no deadlock with the fiber local storage backend.

Fixes: #701

### Definition of Done

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -193,6 +193,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - `Env::get_string` no longer leaks local references. ([#528](https://github.com/jni-rs/jni-rs/pull/528), [#557](https://github.com/jni-rs/jni-rs/pull/557))
+- Deadlocks on Windows when automatically detaching JNI when threads exit. Attachments are now tracked with fiber local storage so detachment happens without holding OS loader lock. ([#701](https://github.com/jni-rs/jni-rs/issues/701))
 
 ### Removed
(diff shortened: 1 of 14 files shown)


rib:
@jerrinot It would be great if you're able to take a quick look over this and confirm whether it fixes the #701 deadlock issue for you

rib:
I'll go ahead and merge this since I did at least test it quite rigorously myself, so hopefully it does also address #701 for @jerrinot

PR #725, opened 2026-02-16
Move jni-macros tests/examples/docs to jni crate

rib:
This removes the jni crate dev-dependency from the jni-macros crate so that we will be able to publish jni-macros to crates.io before publishing the jni crate.

All of the tests and examples that depend on the jni crate have been moved to the jni crate itself.

The jni-macros crate is tagged with `doctest = false` and the macros re-exports in the jni crate are annotated with `#[doc(inline)]` so that the rustdoc examples will instead be tested in the context of the jni crate.


rib:
For reference, I'm ignoring the android CI issues here which will be fixed in https://github.com/jni-rs/jni-rs/pull/724

Issue #731, opened 2026-02-19
Don't allow allow JNI calls with pending exceptions (undefined behaviour)


rib:
As realised / discussed in this comment: https://github.com/jni-rs/jni-rs/issues/730#issuecomment-3928359964 the JNI spec explicitly only allows the following functions to be called safely while there may be a pending exception:

- ExceptionOccurred()
- ExceptionDescribe()
- ExceptionClear()
- ExceptionCheck()
- ReleaseStringChars()
- ReleaseStringUTFChars()
- ReleaseStringCritical()
- Release<Type>ArrayElements()
- ReleasePrimitiveArrayCritical()
- DeleteLocalRef()
- DeleteGlobalRef()
- DeleteWeakGlobalRef()
- MonitorExit()
- PushLocalFrame()
- PopLocalFrame()
- DetachCurrentThread()

Currently `jni-rs` has been making the assumption that calling JNI methods with a pending exception will simply result in getting an error back and not treating it as undefined behaviour.

We need to update `jni/src/macros.rs` so we can differentiate how we call exception-safe functions.

Whenever we call a non-exception-safe function we will need to precede it with an `ExceptionCheck()` call so we can safely return an error in this condition and avoid triggering undefined behaviour.

Although we could potentially try to track pending exceptions via thread-local-storage to try and avoid calling `ExceptionCheck()` this would notably mean we could miss exceptions that might be triggered by external JNI bindings (such as via jni <= 0.21 or non-Rust bindings etc).

rib:
cc: @wuwbobo2021

rib:
Hmm, this may require an API break because it will mean that some of the functions that were made infallible for 0.22 will become fallible if they aren't exception-safe functions.

If we can't land a fix quickly, maybe we can get a way with a 0.22.1 release followed by a yank of 0.22.0

rib:
In particular, these functions were made infallible for 0.22, but they are not all exception-safe:

  - `Env::version`
  - `Env::get_java_vm`
  - `Env::exception_check`
  - `Env::exception_clear`
  - `Env::exception_describe`
  - `Env::exception_occurred`
  - `Env::is_same_object`
  - `Env::delete_local_ref`
  - `WeakRef::is_same_object`
  - `WeakRef::is_weak_ref_to_same_object`
  - `WeakRef::is_garbage_collected`

off the top of my head I'm not sure what internal wrappers also exist that assume they are calling infallible functions and so they also don't return errors.

[...]

PR #733, opened 2026-02-19
Call `ExceptionCheck` before any non-exception-safe JNI calls

rib:
The JNI spec lists a small number of functions that are considered to be safe to call if there is a pending exception and they are:

- ExceptionOccurred()
- ExceptionDescribe()
- ExceptionClear()
- ExceptionCheck()
- ReleaseStringChars()
- ReleaseStringUTFChars()
- ReleaseStringCritical()
- Release<Type>ArrayElements()
- ReleasePrimitiveArrayCritical()
- DeleteLocalRef()
- DeleteGlobalRef()
- DeleteWeakGlobalRef()
- MonitorExit()
- PushLocalFrame()
- PopLocalFrame()
- DetachCurrentThread()

For all other functions we would trigger undefined behaviour if we call them while there is a pending Java exception.

This updates the `jni/src/macros.rs` so:
- We have a separate macro for calling the above exception-safe macros, with no extra checks
- All other JNI calls are preceded with an `ExceptionCheck` call to ensure we can't trigger undefined behaviour. If any pending exception is found then we simply return `JavaException` as we do when a call first triggers an exception.

In the case of `DetachCurrentThread`, which is technically an invocation API, this updates `sys_detach_current_thread` so it always calls `env.exception_clear()` before calling `DetachCurrentThread` and the plan is to follow up with some changes to the `JavaVM::attach_current_thread*` APIs to allow them to capture pending exceptions well before `sys_detach_current_thread` is reached. (See #736)

Unfortunately this has forced us to break a few APIs that had been made infallible for jni 0.22.0, namely:

- `Env::get_java_vm` (`GetJavaVM` is not exception safe)
- `Env::version` (`GetVersion` is not exception safe)
- `Env::is_same_object` (`IsSameObject` is not exception safe)
- `Weak::is_garbage_collected` (based on `Env::is_same_object`)
- `Weak::is_same_object` (deprecated) and (based on `Env::is_same_object`)
- `Weak::is_weak_ref_to_same_object (deprecated)` (based on `Env::is_same_object`)

The current hope is that we land this change fairly quickly and release as jni 0.22.1 and then yank jni 0.22.0 before anyone is depending on it.

Fixes: #731

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,26 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+#### Changed
+
+The following APIs have had to be made fallible again, in order to safely check for pending
+exceptions before calling JNI functions that are not documented as being safe to call with a pending
+exception:
+
+- `Env::get_java_vm` (`GetJavaVM` is not exception safe)
+- `Env::version` (`GetVersion` is not exception safe)
+- `Env::is_same_object` (`IsSameObject` is not exception safe)
+- `Weak::is_garbage_collected` (based on `Env::is_same_object`)
+- `Weak::is_same_object` (deprecated) and (based on
+  `Env::is_same_object`)
+- `Weak::is_weak_ref_to_same_object (deprecated)` (based on
+  `Env::is_same_object`)
+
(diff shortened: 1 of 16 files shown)


rib:
@wuwbobo2021 it would be great if you could review this change if you get a moment please

Issue #732, opened 2026-02-19
JavaVM::attach_current_thread() should catch + clear exceptions


rib:
As a general rule `jni-rs` only checks for exceptions after method calls so that it can return a `JavaException` error that allows developers decide whether they want to let the exception propagate back to the JVM or let them clear the exception and map it to a custom rust error.

If you're using JNI from a native thread though (where you would use `JavaVM::attach_current_thread()` to explicitly attach the thread to the JVM) then there is no JVM at the bottom of the stack that can catch an exception and if you let a `JavaException` error unravel all the way back to an `attach_current_thread()` closure then nothing is going to automatically happen with that exception.

Since it's also an error to try and use JNI while you have pending exceptions then this behaviour currently makes it very easy to ignore Java exceptions in native threads and then unwittingly trigger errors in future attempts to use `attach_current_thread()` + JNI on the same thread that could still have a pending exception.

Until #731 is addressed this is especially bad because this can lead to undefined behaviour.

`attach_current_thread()` (and the lower-level `attach_current_thread_with_config`) should automatically make sure to catch + clear exceptions if they see the closure return a `JavaException` error.

One fiddly detail here though is that these closures don't directly return a `jni:errors::Error` (they can return anything that implements `Into<jni::errors::Error>`) so we will need to use `ExceptionCheck` instead of trying to look at the error itself.

If we catch an exception then we should return an owned representation of the exception as a Rust error (at least including a `String` with a formatted backtrace)

rib:
cc: @wuwbobo2021

wuwbobo2021:
This problem is more significant when using `JavaVM::attach_current_thread_for_scope`. In my past experience of using `jni` 0.21, detaching from JVM while there is a pending exception may cause a FATAL EXCEPTION found in the Android adb logcat, which looks more serious than calling an unallowed JNI function while there is a pending exception; both situation crashes the Android application, though. Note that in Java 8 `DetachCurrentThread` is not safe to call when a pending exception exists.

rib:
To be clear here, when I refer to `attach_current_thread` I also mean `attach_current_thread_for_scope` and `attach_current_thread_with_config`

Even more generally, we will probably do something to update the internal `sys_detach_current_thread` wrapper that calls `DetachCurrentThread` - which will probably also affect detachments that happen via `AttachGuard::Drop`

I expect that only the attachment APIs that take a closure will be able to get back an Error that contains a full exception description though.

[...]

PR #736, opened 2026-02-20
Catch exceptions after running attachment closures

rib:
Firstly this adds a utility `Env::exception_catch` that can clear any pending exception and map it to a `Error::CaughtJavaException`

Then it adds an `AttachGuard::detach_with_catch` that can explicitly drop an attach guard (like `AttachGuard::detach()`) but also ask for any pending exceptions to be returned as `Error::CaughtJavaException`.

And finally; all `JavaVM::attach_current_thread*` APIs finish by calling `AttachGuard::detach_with_catch` after they have run the given closure - ensuring that they don't leave any exceptions pending before logically detaching from the thread.

Note: that these APIs will check for exceptions before they return, regardless of whether they _actually_ triggered a thread attachment, so exception handling is consistent in case the attachment calls are nested.

Considering that the JNI spec for Java 8 doesn't list DetachCurrentThread as one of the APIs that may be called while exceptions are pending; this also updates `sys_detach_current_thread` so that it calls `env.exception_clear()` before calling `DetachCurrentThread` (it doesn't fetch the details of any exception, since it's assumed that `AttachGuard::detach_with_catch` will be used if the details are needed).

Fixes: #732

### Definition of Done

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,11 +17,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- `Env::exception_catch` provides a convenient way of catching pending Java exceptions and mapping them to `Error::CaughtJavaException` ([#736](https://github.com/jni-rs/jni-rs/pull/736))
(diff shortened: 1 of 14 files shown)


rib:
cc: @wuwbobo2021

PR #742, opened 2026-02-22
docs: fix duplicate word typos in docs/comments

adityagiri3600:
## Overview                                                                                             
                                                                                                          
  Fix duplicate word typos in docs and comments.                                                                    

  ### Definition of Done                                                                                  
                                                                                                          
  - [x] Change is covered by [automated tests](https://github.com/jni-rs/jni-rs/blob/master/              
  CONTRIBUTING.md#tests)                                                                                  
  - [x] The [coding guidelines](https://github.com/jni-rs/jni-rs/blob/master/CONTRIBUTING.md#the-code-    
  style) are followed                                                                                     

### DIFF
--- a/crates/jni/docs/0.22-MIGRATION.md
+++ b/crates/jni/docs/0.22-MIGRATION.md
@@ -751,7 +751,7 @@ of RAII guard types in `jni-rs` that may automatically call JNI methods in their
 ### My application freezes after migrating away from daemon thread attachments
 
 If you find that your application freezes or blocks when exiting, after migrating
-away from daemon thread attachments this is a warning sign that you you have an
+away from daemon thread attachments this is a warning sign that you have an
 unsound exit path where there may be threads running that have an invalid JNI
 environment pointer and the potential to try calling JNI methods after
 the JVM is destroyed.
--- a/crates/jni/docs/macros/bind_java_type_3_advanced.md
+++ b/crates/jni/docs/macros/bind_java_type_3_advanced.md
@@ -267,7 +267,7 @@ fn wrapper<'local>(mut unowned_env: EnvUnowned<'local>, this: MyType<'local>) ->
 
 ### Disabling `catch_unwind`
 
-If you you want to allow your native method to abort on panic or you plan to handle panics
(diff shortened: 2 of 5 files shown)


wuwbobo2021:
In `crates/jni/src/macros.rs`, there are comments of "When calling any function added after JNI 1.1 you must know that it's valid for the current JNI version.".

However, some other `Env` methods that requires JNI 1.4 includes comments of "Safety: jni-rs requires JNI >= 1.4 and this is checked in from_raw"; this has confused me, and I can't find the exact JNI version checker in any `from_raw` method, I guess the actual JNI version checking is done by `GetEnv` and `AttachCurrentThread` wrappers in `java_vm.rs`.

rib:
@wuwbobo2021 Essentially those comments are saying that if it's a JNI 1.4 API or less then we can "safely" assume they exist - not because we have done a runtime check but because we consider it inherently unsafe to attempt to use the `jni` crate against a JVM that doesn't at _least_ support JNI 1.4.

In other words its a minimum bar that is imposed on any user of the `jni` crate - you can't expect the implementation to be safe if you try and run it with such an out-dated version of JNI that it doesn't include all of the JNI 1.4 functions at least.

There are technical chicken and egg issues with very early versions of JNI and `GetVersion` is also not exception safe (so we wouldn't know if we can use `ExceptionCheck`)

See these for a bit more context:
- https://github.com/jni-rs/jni-rs/pull/478/changes/450a1641b69137a859860bb239b3ca90e1d52bc0
- https://github.com/jni-rs/jni-rs/pull/478/changes/0d9fed6af2996a020b2d134a50dd4e1b481181b9

This constraint is documented in a few places I believe but it could probably be documented better still.

Note: Java 1.4 / JNI 1.4 was released in 2002 and this is a _very_ conservative minimum requirement.

E.g. considering that Rust itself only supports Windows 10+ it seems reasonable to not need to specifically check versioning details that far back.

I imaging it's pretty hard to find a java runtime these days that doesn't support JNI 1.4 so I doubt anyone really needs to worry about this.

PR #745, opened 2026-02-23
Upgrade to libloading 0.9

djc:


rib:
Annoyingly this is technically a breaking change because `StartJvmError::LoadError` exposes the `libloading::Error` :/

I'm not even sure why I didn't update this before the 0.22 release. I think at some point I intentionally skipped it since the rust-version looked a bit too recent but then by the time 0.22 was released then 1.88 is fine.

hmm - could repeat the pattern again of making a small breaking change in 0.22.2 and yanking 0.22.1 - still assuming that it's highly unlikely anyone has locked in a 0.22.x dependency yet

djc:
Ahh, sorry -- if this has a 1.88 MSRV than I'd probably prefer to hold off on this...

Also highly recommend `Box<dyn Error>`-wrapping any errors from external crates.

rib:
> `Box<dyn Error>`

yup for sure, probably would have been a better choice.

> Ahh, sorry -- if this has a 1.88 MSRV than I'd probably prefer to hold off on this...

yeah it looks like it.

I don't think we've ever codified a MSRV policy for this crate so far but since I see it as being a fairly low-level plumbing API my instinct is to stay pretty conservative. We should probably document that we aim for >= 1yr backwards compatibility and so currently 1.85 (our current rust-version) is already the latest version by that standard.

Issue #741, opened 2026-02-21
Don't report `get_method_id/get_field_id` exceptions as `NullPtr` errors without clearing exceptions


rib:
> Side note: `Env::get_method_id` or `Env::get_field_id` may produce `NoSuchMethodError` or `NoSuchFieldError` that needs to be cleared manually. Currently both `Env` methods do not return `Error::JavaException` on error, and one might not expect to produce a pending exception in a failure of finding a field/method ID.
> 
> The `hiddenapi-flags.csv` filtering feature of your great `jbindgen` (plus <https://github.com/Dirbaio/java-spaghetti/issues/13#issuecomment-3085099935>) have reminded me of this problem. I did a test and found that the problem exists in `jni` 0.22.1. 

 _Originally posted by @wuwbobo2021 in [#730](https://github.com/jni-rs/jni-rs/issues/730#issuecomment-3939303019)_

It's really important that pending exceptions are consistently reported as `JavaException` errors so that code knows it can't continue to call into JNI until it has handled the pending exception.

By returning `Error::NullPtr` from `get_method_id` and `get_field_id` then code may not realize that this is another special case that requires them to clear a pending exception.

For compatibility, we should continue to return `Error::NullPtr` in response to `NoSuchMethodError` and `NoSuchFieldError` exceptions but we should automatically clear the exception internally before returning.

Initially it may also be acceptable to also continue to report `ExceptionInInitializerError` and `OutOfMemoryError` exceptions as `Error::NullPtr` (and automatically clearing the exception) but ideally at some point these exceptions should probably return `JavaException` errors instead.

wuwbobo2021:
(edited)

> It's really important that pending exceptions are consistently reported as JavaException errors so that code knows it can't continue to call into JNI until it has handled the pending exception.

As described in <https://github.com/jni-rs/jni-rs/issues/732#issuecomment-3941400063> by @catgirlconspiracy, <https://github.com/jni-rs/jni-rs/issues/743#issuecomment-3941719762> shows the side effect of clearing the previously produced exception found in the drop glue of `Global`. This might be fixed by not to do exception clearing when doing a "false" detach that doesn't involve a real thread detachment JNI call. However, there might be more problems of similar side affects occurring elsewhere.

[...]

PR #748, opened 2026-02-23
`get_[static_]method/field_id`: squash exceptions into `Method/FieldNotFound` errors

rib:
The way that the `Get[Static]Method/FieldID` JNI calls work is that they will throw an exception on failure, in addition to returning a null ID.

Since these JNI calls were wrapped with the
`jni_call_post_check_ex_and_null_ret` macro that meant that we would immediately return a `JavaException` error to the caller without having the chance to try and map these exceptions to a `Method/FieldNotFound` error instead.

Additionally the previous attempt to map `NullPtr` errors into `Method/FieldNotFound` errors did not consider the need to clear pending exceptions in case we are not returning a `JavaException` error.

The `get_[static]_method/field_id` implementations now make sure to:
- check for pending exceptions before making any JNI call so that they will avoid any side effects from potentially clearing exceptions that they didn't cause.
- check for null method/field IDs and explicitly clear exceptions before returning a `Method/FieldNotFound` error.
- as an extra pedantic check we also assert that there must be no pending exception in case we get back a non-null/valid method/field ID.

This adds unit tests to ensure we get `Method/FieldNotFound` errors with no pending exception side effects.

Fixes: #741

### Definition of Done

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,9 +17,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-#### Changed
+### Changed
 
 - Removed dependency on `paste` crate ([#752](https://github.com/jni-rs/jni-rs/pull/752))
+
+### Fixed
+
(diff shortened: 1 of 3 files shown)


rib:
Note: there are other methods besides these `get_method/field_id` APIs that should probably be updated to squash certain key exceptions internally (instead of always reporting them as `JavaException` errors) but those can be addressed in separate PRs.

Issue #743, opened 2026-02-22
Dropping a `Global` with a pending exception in the `EnvUnowned::with_env` closure may produce segmentation fault


wuwbobo2021:
This issue is first discovered by <https://github.com/jni-rs/jni-rs/issues/732#issuecomment-3941400063>.

`jni` version: 0.22.1

## Test case

`Makefile`:

```
java_run: lib
	javac HelloWorld.java && java -Djava.library.path=mylib/target/debug/ HelloWorld

.PHONY: lib

javah:
	javah HelloWorld

lib:
	cd mylib && cargo build
```

`HelloWorld.java`:

```java
class HelloWorld {
    private static native String hello(String input);

    static {
        System.loadLibrary("mylib");
    }

    public static void main(String[] args) {
        String output = HelloWorld.hello("josh");
        System.out.println(output);
    }
}

```

`mylib/Cargo.toml`:

```toml
[package]
name = "mylib"
version = "0.1.0"
edition = "2021"

[dependencies]
jni = { version = "0.22.1", features = ["invocation"] }

[lib]
crate-type = ["cdylib"]

[workspace]
```

`mylib/src/lib.rs`:

```rust
#[no_mangle]
pub extern "system" fn Java_HelloWorld_hello<'local>(
    mut unowned_env: jni::EnvUnowned<'local>,
    _class: jni::objects::JClass<'local>,
    input: jni::objects::JString<'local>,
) -> jni::objects::JString<'local> {
    let outcome = unowned_env.with_env(|env| -> jni::errors::Result<_> {
        {
            let jstring_not_numeric = env.new_string("s").unwrap();
            let global_string = env.new_global_ref(&jstring_not_numeric).unwrap();

            let err_result = env.call_static_method(
                jni::jni_str!("java/lang/Integer"),
                jni::jni_str!("parseInt"),
                jni::jni_sig!((str: JString) -> jint),
                &[(&jstring_not_numeric).into()],
            );
            assert!(matches!(err_result, Err(jni::errors::Error::JavaException)));
            //assert!(!env.exception_check()); // this will be true if the string is numeric
            //env.exception_clear(); // uncomment this to eliminate the segfault

            eprintln!("dropping global_string...");
            drop(global_string);
            eprintln!("dropped global_string.");
        }

        let input: String = input.to_string();
        env.new_string(format!("Hello, {}!", input))
    });
    outcome.resolve::<jni::errors::ThrowRuntimeExAndDefault>()
}
```

[...]

PR #749, opened 2026-02-23
Ensure `Global/Weak::Drop` has no exception_clear side effects

rib:
This updates the implementations of `Global/Weak::Drop` so that they don't clear pending exceptions as a side effect.

The issue was that #736 overlooked the internal use of `attach_current_thread` in these Drop implementations and didn't consider that these Drop methods need to remain exception safe and should leave any existing pending exception in place.

We now use `JavaVM::attach_current_thread_guard` so we can intentionally avoid `AttachGuard::detach_with_catch`, which would catch/clear any pending exceptions.

This adds unit tests to check that there are no exception_clear side effects from dropping a `Global` or `Weak` reference.

Addresses: #743

(It's not marked as a fix for #743 because #743 represents more than one underlying issue.)

### Definition of Done

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - `Env::get_[static_]method/field_id` APIs now correctly clear + map internal exceptions to `Error::Method/FieldNotFound` errors ([#748](https://github.com/jni-rs/jni-rs/pull/748))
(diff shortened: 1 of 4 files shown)


rib:
cc: @wuwbobo2021 and @catgirlconspiracy

rib:
~Oh, actually, I need to follow up to make sure we still create a scoped attachment in the case where the current thread is not already attached - my initial fix will result in a permanent attachment of the thread~

*edit*: ok pushed a fix for this

Issue #750, opened 2026-02-23
Fix `Env::throw` APIs to _actually_ return `JavaException`


rib:
I apparently rushed https://github.com/jni-rs/jni-rs/pull/738 and although I updated to api docs, changelog and migration guide to mention this I didn't _actually_ change the implementation 🤦‍♂️

I noticed this when writing the unit test for https://github.com/jni-rs/jni-rs/pull/749 because I initially went to assert that `env.throw` did in fact return `JavaException` and that failed. 

Lets also assert that this issue shouldn't be closed without adding explicit unit tests to ensure that these APIs really do behave as they are documented!

PR #755, opened 2026-02-24
Fix to return `JavaException` from `Env::throw*` APIs

rib:
The PR for #738 was unfortunately rushed and since it didn't add a unit test then I missed the fact that I had actually only updated all the docs + changelog but omitted the functional change 🤦

This makes the functional change that was documented in #738.

This updates existing `env.throw*` tests and adds some additional tests to ensure the APIs return `Err(JavaException)` as expected - and double checks the exceptions match what we threw.

Fixes: #750

### Definition of Done


rib:
cc: @catgirlconspiracy

Issue #761, opened 2026-02-25
`convert_byte_array` should check for exceptions after calling `GetByteArrayRegion`


rib:
While auditing all internal usage of `jni_call_no_post_check_ex` (i.e. places where we make JNI calls without checking for exceptions afterwards) and cross referencing with the JNI spec to check whether the JNI function can throw exceptions I found this case where we call `GetByteArrayRegion` in `Env::convert_byte_array` without checking for exceptions.

The JNI spec says that `GetByteArrayRegion` can throw an `ArrayIndexOutOfBoundsException`.

Although the implementation of `convert_byte_array` explicitly queries the length of the array and it should be unlikely to trigger an `ArrayIndexOutOfBoundsException`, there is no locking that would stop a race condition such that the length could be incorrect when `GetByteArrayRegion` is called.

Note that we do check for exceptions where we call `Get<Type>ArrayRegion` in the implementation of `JPrimitiveArrays`.

Issue #744, opened 2026-02-23
Should `jni-rs` change its design so it catches exceptions eagerly?


rib:
The current design of `jni-rs`, where it checks for exceptions so it can return a `JavaException` (but leaves the exception pending) is something that pre-dates my involvement in maintaining the crate, but I have always considered that to be a reasonable hands-off design that:
1. tries to avoid doing too much automatically behind your back (that's not necessarily needed from a safety pov)
2. allows you to intentionally leave the exception pending and return control the JVM so that the exception can be caught in Java

I've also assumed that information would be lost if you would try to always catch exceptions automatically into a `Global` + clear the exception and then later attempt to re-throw the exception before returning to the JVM (I would assume that the stack trace would be re-captured).

but this design has also come up as a pain point more than once.

For example there is this Draft PR https://github.com/jni-rs/jni-rs/pull/498 and most recently this discussion in #741:

> 
> It's really important that pending exceptions are consistently reported as JavaException errors so that code knows it can't continue to call into JNI until it has handled the pending exception.
> 
> As described in <https://github.com/jni-rs/jni-rs/issues/732#issuecomment-3941400063> by @catgirlconspiracy, <https://github.com/jni-rs/jni-rs/issues/743#issuecomment-3941719762> shows the side effect of clearing the previously produced exception found in the drop glue of `Global`. This might be fixed by not to do exception clearing when doing a "false" detach that doesn't involve a real thread detachment JNI call. However, there might be more problems of similar side affects occurring elsewhere.
> 
> There are many `Env` methods that wraps JNI calls which may throw an exception, and some of them are also used internally in `jni-rs`. As an high-level API, `jni-rs` has already managed to do automatic exception clearing in some cases; one may expect `jni-rs` to return `Error::JavaException` only from a few `Env` methods including `call_method`, and exceptions produced by JNI itself (not being thrown from the Java side object/class's method or intializer) should be converted to some other error type in the `Error` enum, and such exceptions needs to be cleared inside `jni-rs`.
> 
> Excuse me, let me go back to the question of why `jni-rs` does not return a JNI reference to the exception object in `Error::JavaException`. If so, it may clear the pending exception immediately, without worrying about the caller side not having the right to check the exception. I guess the current `Error::JavaException` is designed to avoid a lifetime bound (for the local reference) of the `Error` type, or the performance cost of creating a global reference.
> 
[...]

PR #762, opened 2026-02-25
Catch and Map Internal JNI Exceptions to Errors

rib:
_Just as a brief clarification; up until now `jni-rs` would handle exceptions (broadly speaking) by checking for thrown exceptions and reporting them as Rust errors via `Error::JavaException`, and then it was your responsibility to take and clear the exception before continuing (or return to Java)._

_This is a very hands-off approach that takes a lot of effort to handle correctly and reliably because practically *any* `jni` method could throw an exception that you might need to handle before calling another `jni` method._

This PR is trying to reduce the burden on developers checking for exceptions, so that they only need to expect exceptions when code is calling into Java (such as when constructing objects or calling methods).

This makes it so that all of the lower-level JNI function bindings will catch their exceptions internally and report these as plain `jni::errors::Error` enum variants, with no extra need to check and clear an exception.

The API is still the same in the sense that `Error::JavaException` is how `jni-rs` indicates that you need to handle a pending exception, but for most lower-level APIs you can now assume that you'll never get back an `Error::JavaException` (and the documentation has been updated to clarify when this can be assumed).

# New `Error` variants

Firstly, this adds `Error`s to cover all internally-caught exceptions

The new variants ensure we can map any exceptions we catch internally into a `jni::errors::Error` result.

# Internal `catch |env| {}` macros

This adds a `jni_call_with_catch` macro to catch internal exceptions, making it practical to match exception types and map then to `Error` values, like this:

```rust
jni_call_with_catch_and_null_check!(
    catch |env| {
        crate::exceptions::JClassFormatError =>
            Err(Error::ClassFormatError),
        crate::exceptions::JClassCircularityError =>
            Err(Error::ClassCircularityError),
        crate::exceptions::JOutOfMemoryError =>
            Err(Error::JniCall(JniError::NoMemory)),
        crate::exceptions::JSecurityException =>
            Err(Error::SecurityViolation),
        else => Err(Error::NullPtr("Unexpected Exception")),
    },
    self,
    v1_1,
    DefineClass,
    name,
    loader.as_raw(),
    buf,
    buf_len as jsize
)
```

A similar `jni_call_with_catch_and_null_check` additionally maps `null` return values to `Error::NullPtr` (in case no exception was caught).

A more-general `jni_try!` macro can be used to wrap arbitrary blocks of JNI code in a closure and use the same `catch |env| {}` syntax for matching and mapping exceptions to `Error` results, like this:

```rust
jni_try!(
    (env) -> ResultType {
        // JNI code here
    },
    catch |env| {
        ExceptionType1 => Err(Error::Type1),
        exception: ExceptionType2 => {
            let cause = exception.get_cause(env)?;
            let cause = env.new_global_ref(cause)?;
            Err(Error::Type2(cause))
        },
        // required catch-all
        else => Err(Error::JniCall(JniError::Unknown)),
    },
)
```

In particular this is used in the implementation of `LoaderContext`.

# Audit of all JNI calls that may throw exceptions

This change audits all of the internal JNI calls and adds explicit capturing and clearing of exceptions so they can be mapped to more-precise errors than `Error::JavaException` and so there is less burden on developers to have to handle exceptions manually.

In this context "internal" JNI calls are defined as calls that don't intentionally invoke arbitrary Java code - such as `Call*` or `NewObject` functions.

With this change, developers should be able to assume that most lower-level JNI `Env` methods (not involving arbitrary Java binding method calls) will not return with a pending exception and `Error::JavaException` result.

Updated methods now explicitly document that they catch exceptions internally and clarify what kinds of `Error`s may be returned in different circumstances.

# Catching exceptions in LoaderContext

The `LoaderContext` changes deserve a special mention here because, unlike other APIs that may directly map to a single JNI function (such as `FindClass`, the `LoaderContext::load_class` implementation is itself based on various `jni::objects::` method bindings (which can still throw exceptions).

Although the implementation had already attempted to catch exceptions internally, there was an oversight and the `::find_class()` fallback didn't catch exceptions.

This PR has made a much-more-rigorous update to this code to use the internal `jni_try!` macro and it has also ensured there is consistency between `Env::find_class` and `Env::load_class` +
`LoaderContext::load_class` with how exceptions are mapped to errors.

Note: from a documentation POV this looks like a breaking change because the documentation now makes it clear that these APIs return `Error::NoClassDefFound` if no class could be loaded, where as before `find_class` suggested it would return `NullPtr` and `load_class` suggested it would return `ClassNotFound` but in practice the previous documentation was incorrect in both cases because they would most likely just return `Error::JavaException`.

The newly added `Error::NoClassDefFound` matches the `NoClassDefFoundError` that `FindClass` typically throws on failure, and for consistency `load_class` will map any lower-level `ClassNotFoundException` from the loader into a `NoClassDefFoundError` similar to what `FindClass` does. In either case, the original exception is also preserved in the Rust error as the `cause`, in case more details are needed.

Fixes: #761
Addresses: #744

## Audit Log

These are the JNI APIs that the spec says can throw exceptions (and which we use in `jni-rs`) which this PR has audited and updated to ensure they catch exceptions and map them to an `Error` instead of returning `Error::JavaException`:

- [x] `DefineClass`
  - **Returns:** Returns a Java class object or `NULL` if an error occurs.
  - **Throws:**
    - `ClassFormatError`: if the class data does not specify a valid class.
    - `ClassCircularityError`: if a class or interface would be its own superclass or superinterface.
    - `OutOfMemoryError`: if the system runs out of memory.
    - `SecurityException`: if the caller attempts to define a class in the `"java"` package tree.
  - **Status:** Handled (jni_call_with_catch_and_null_check)

- [x] `FindClass`
  - **Returns:** Returns a class object from a fully-qualified name, or `NULL` if the class cannot be found.
  - **Throws:**
    - `ClassFormatError`: if the class data does not specify a valid class.
    - `ClassCircularityError`: if a class or interface would be its own superclass or superinterface.
    - `NoClassDefFoundError`: if no definition for a requested class or interface can be found.
    - `OutOfMemoryError`: if the system runs out of memory.
  - **Status:** Handled (jni_call_with_catch_and_null_check)

- [x] `NewWeakGlobalRef`
  - **Returns:** Return a global weak reference to the given `obj`.
  - **Throws:**
    - `OutOfMemoryError`: if the system runs out of memory.
  - **Status:** Handled (jni_call_with_catch)

- [x] `AllocObject`
  - **Returns:** Returns a Java object, or `NULL` if the object cannot be constructed.
  - **Throws:**
    - `InstantiationException`: if the class is an interface or an abstract class.
    - `OutOfMemoryError`: if the system runs out of memory.
  - **Status:** Handled (jni_call_with_catch_and_null_check)

- [x] `GetFieldID`
  - **Returns:** Returns a field ID, or `NULL` if the operation fails.
  - **Throws:**
    - `NoSuchFieldError`: if the specified field cannot be found.
    - `ExceptionInInitializerError`: if the class initializer fails due to an exception.
    - `OutOfMemoryError`: if the system runs out of memory.
  - **Status:** Handled (jni_call_with_catch)

- [x] `GetMethodID`
  - **Returns:** Returns a method ID, or `NULL` if the specified method cannot be found.
  - **Throws:**
    - `NoSuchMethodError`: if the specified method cannot be found.
    - `ExceptionInInitializerError`: if the class initializer fails due to an exception.
    - `OutOfMemoryError`: if the system runs out of memory.
  - **Status:** Handled (jni_call_with_catch)

- [x] `GetStaticFieldID`
  - **Returns:** Returns a field ID, or `NULL` if the specified static field cannot be found.
  - **Throws:**
    - `NoSuchFieldError`: if the specified static field cannot be found.
    - `ExceptionInInitializerError`: if the class initializer fails due to an exception.
    - `OutOfMemoryError`: if the system runs out of memory.
  - **Status:** Handled (jni_call_with_catch)

- [x] `GetStaticMethodID`
  - **Returns:** Returns a method ID, or `NULL` if the operation fails.
  - **Throws:**
    - `NoSuchMethodError`: if the specified static method cannot be found.
    - `ExceptionInInitializerError`: if the class initializer fails due to an exception.
    - `OutOfMemoryError`: if the system runs out of memory.
  - **Status:** Handled (jni_call_with_catch)

- [x] `NewStringUTF`
  - **Returns:** Returns a Java string object, or `NULL` if the string cannot be constructed.
  - **Throws:**
    - `OutOfMemoryError`: if the system runs out of memory.
  - **Status:** Handled (jni_call_with_catch_and_null_check)

- [x] `NewObjectArray`
  - **Returns:** Returns a Java array object, or `NULL` if the array cannot be constructed.
  - **Throws:**
    - `OutOfMemoryError`: if the system runs out of memory.
  - **Status:** Handled (jni_call_with_catch_and_null_check) (correctly treats null as an error now)

- [x] `GetObjectArrayElement`
  - **Returns:** Returns a Java object.
  - **Throws:**
    - `ArrayIndexOutOfBoundsException`: if `index` does not specify a valid index in the array.
  - **Status:** Handled (jni_call_with_catch)

- [x] `SetObjectArrayElement`
  - **Returns:** None (void).
  - **Throws:**
    - `ArrayIndexOutOfBoundsException`: if `index` does not specify a valid index in the array.
    - `ArrayStoreException`: if the class of `value` is not a subclass of the element class of the array.
  - **Status:** Handled (jni_call_with_catch)

- [x] `New<Boolean|Byte|Char|Short|Int|Long|Float|Double>Array`
  - **Returns:** Returns a Java array, or `NULL` if the array cannot be constructed.
  - **Throws:**
    - `OutOfMemoryError`: if the system runs out of memory.
  - **Status:** Handled (jni_call_with_catch_and_null_check)

- [x] `Get<Boolean|Byte|Char|Short|Int|Long|Float|Double>ArrayElements`
  - **Returns:** Returns a pointer to the array elements, or `NULL` if the operation fails.
  - **Throws:**
    - `OutOfMemoryError`: if the system runs out of memory.
  - **Status:** Handled (jni_call_with_catch_and_null_check)

- [x] `Get<Boolean|Byte|Char|Short|Int|Long|Float|Double>ArrayRegion`
  - **Returns:** None (void).
  - **Throws:**
    - `ArrayIndexOutOfBoundsException`: if one of the indexes in the region is not valid.
  - **Status:** Handled (jni_call_with_catch)

- [x] `Set<Boolean|Byte|Char|Short|Int|Long|Float|Double>ArrayRegion`
  - **Returns:** None (void).
  - **Throws:**
    - `ArrayIndexOutOfBoundsException`: if one of the indexes in the region is not valid.
  - **Status:** Handled (jni_call_with_catch)

- [x] `GetPrimitiveArrayCritical`
  - **Returns:** Returns a pointer to the array elements, or `NULL` if the operation fails.
  - **Throws:**
    - `OutOfMemoryError`: if the system runs out of memory.
  - **Status:** Handled (jni_call_with_catch_and_null_check)

- [x] `RegisterNatives`
  - **Returns:** Returns `"0"` on success; returns a negative value on failure.
  - **Throws:**
    - `NoSuchMethodError`: if a specified method cannot be found or if the method is not native.
  - **Status:** Handled (jni_call_with_catch)

- [x] `MonitorExit`
  - **Returns:** Returns `"0"` on success; returns a negative value on failure.
  - **Throws:**
    - `IllegalMonitorStateException`: if the current thread does not own the monitor.
  - **Status:** Handled - `IllegalMonitorStateException` isn't possible because `MonitorGuard` is not `Send`

- [x] `NewDirectByteBuffer`
  - **Returns:** Returns a local reference to the newly-instantiated `java.nio.ByteBuffer` object; returns `NULL` if an exception occurs, or if JNI direct-buffer access is not supported by this VM.
  - **Throws:**
    - `IllegalArgumentException`: if `capacity` is negative or greater than `Integer.MAX_VALUE`.
    - `OutOfMemoryError`: if allocation of the `ByteBuffer` object fails.
  - **Status:** Handled (jni_call_with_catch_and_null_check)

Checking whether we need to update `jni-macros` to automatically catch exceptions I double checked which JNI functions `bind_java_type` calls directly and none of them are affected by this:
- `IsAssignableFrom` - doesn't throw exceptions
- `NewObjectA` - we don't want to automatically catch constructor exceptions
- `Call[Static]<Type>MethodA` - we don't want to automatically catch method exceptions
- `Get[Static]<Type>Field` - doesn't throw exceptions

A special-case that was also reviewed closely was the `LoaderContext::load_class` implementation.

`load_class` internally calls a number of other binding methods (such as `JThread::current_thread` and `JThread::get_context_class_loader`) and since these are regular Java methods they can throw exceptions, which in turn means that `load_class` can throw exceptions.

This PR has updated the `LoaderContext` to use an internal `jni_try!` macro as documented earlier.

### Definition of Done

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,7 @@ Adds bindings for the following `java.lang` errors / exceptions ([#767](https://
 - Removed dependency on `paste` crate ([#752](https://github.com/jni-rs/jni-rs/pull/752))
 - `attach_current_thread*` APIs immediately return `Err(JavaException)` if a Java exception is pending, so they don't have the side effect of clearing exceptions not thrown in the given closure ([#756](https://github.com/jni-rs/jni-rs/pull/756))
 - Removed `proc-macro-crate` dependency from the `jni-macros` crate ([#758](https://github.com/jni-rs/jni-rs/pull/758))
+- All internal use of JNI functions (not general calls into Java code) now catch exceptions and map to `jni::errors::Error` ([#762](https://github.com/jni-rs/jni-rs/pull/762))
 
 ### Fixed
 
--- a/crates/jni/src/elements/auto_elements_critical.rs
+++ b/crates/jni/src/elements/auto_elements_critical.rs
@@ -42,11 +42,14 @@ where
         mode: ReleaseMode,
     ) -> Result<Self> {
         let mut is_copy: jboolean = true;
-        // There are no documented exceptions for GetPrimitiveArrayCritical() but
-        // it may return `NULL`.
 
         let ptr = unsafe {
-            jni_call_only_check_null_ret!(
+            jni_call_with_catch_and_null_check!(
+                catch |env| {
+                    crate::exceptions::JOutOfMemoryError =>
+                        Err(Error::JniCall(JniError::NoMemory)),
+                    else => Err(Error::NullPtr("Unexpected Exception")),
+                },
                 env,
                 v1_2,
                 GetPrimitiveArrayCritical,
--- a/crates/jni/src/env.rs
+++ b/crates/jni/src/env.rs
@@ -483,12 +483,31 @@ See the jni-rs Env documentation for more details.
         // Env lifetime for the top JNI stack frame
         self.assert_top();
         // Safety:
-        // DefineClass is 1.1 API that must be valid
-        // It is valid to potentially pass a `null` `name` to `DefineClass`, since the
-        // name can bre read from the bytecode.
-        // A null loader corresponds to the bootstrap class loader.
+        // - DefineClass is 1.1 API that must be valid
+        // - It is valid to potentially pass a `null` `name` to `DefineClass`, since the name can be
+        // read from the bytecode.
+        // - A null loader corresponds to the bootstrap class loader.
+        //
+        // Throws:
+        // - `ClassFormatError`: if the class data does not specify a valid class.
+        // - `ClassCircularityError`: if a class or interface would be its own superclass or
+        //   superinterface.
+        // - `OutOfMemoryError`: if the system runs out of memory.
+        // - `SecurityException`: if the caller attempts to define a class in the `"java"` package
+        //   tree.
         unsafe {
-            jni_call_post_check_ex_and_null_ret!(
+            jni_call_with_catch_and_null_check!(
+                catch |env| {
+                    crate::exceptions::JClassFormatError =>
+                        Err(Error::ClassFormatError),
+                    crate::exceptions::JClassCircularityError =>
+                        Err(Error::ClassCircularityError),
+                    crate::exceptions::JOutOfMemoryError =>
+                        Err(Error::JniCall(JniError::NoMemory)),
+                    crate::exceptions::JSecurityException =>
+                        Err(Error::SecurityViolation),
+                    else => Err(Error::NullPtr("Unexpected Exception")),
+                },
                 self,
                 v1_1,
                 DefineClass,
@@ -514,6 +533,17 @@ See the jni-rs Env documentation for more details.
     ///
     /// Alternatively, call [`Self::define_class_jbyte`] if your data comes from a [`JByteArray`] or
     /// `&[jbyte]`.
+    ///
+    /// - Returns [`Error::ClassFormatError`] if the class data does not specify a valid class.
+    /// - Returns [`Error::ClassCircularityError`] if a class or interface would be its own
+    ///   superclass or superinterface.
+    /// - Returns [`Error::JniCall`] + [`JniError::NoMemory`] if the JVM reports running out of
+    ///   memory.
+    /// - Returns [`Error::SecurityViolation`] if the caller attempts to define a class in the
+    ///   `"java"` package tree.
+    ///
+    /// This API catches exceptions internally and is not expected to return
+    /// [`Error::JavaException`] (unless called while there is a pending exception).
     ///
     /// # Portability Note
     ///
@@ -558,6 +588,17 @@ See the jni-rs Env documentation for more details.
     ///
     /// This is the same as [`Self::define_class`] but takes a `&[jbyte]` instead of `&[u8]`.
     ///
+    /// - Returns [`Error::ClassFormatError`] if the class data does not specify a valid class.
+    /// - Returns [`Error::ClassCircularityError`] if a class or interface would be its own
+    ///   superclass or superinterface.
+    /// - Returns [`Error::JniCall`] + [`JniError::NoMemory`] if the JVM reports running out of
+    ///   memory.
+    /// - Returns [`Error::SecurityViolation`] if the caller attempts to define a class in the
+    ///   `"java"` package tree.
+    ///
+    /// This API catches exceptions internally and is not expected to return
+    /// [`Error::JavaException`] (unless called while there is a pending exception).
+    ///
     /// # Portability Note
     ///
     /// The JNI `DefineClass` function may not be supported by all JVM implementations, even though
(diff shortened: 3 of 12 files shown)


rib:
cc: @wuwbobo2021

PR #770, opened 2026-02-27
Release javac 0.1.1

rib:


rib:
For reference @wuwbobo2021 your fix has been published to crates.io now: https://crates.io/crates/javac

PR #793, opened 2026-03-14
Add `objects::JCharSequence` binding

rib:
This adds a `java.lang.CharSequence` binding and declared `JString` to be an instance of a `JCharSequence`

For example this is useful in the context of Android bindings where it's quite common for APIs to accept a `CharSequence` instead of a `String`.

Note: this does not mark `JCharSequence` as a "core" type in jni-macros and so this should not break external code that may have already defined custom `java.lang.CharSequence` bindings. Still it would probably be best for external bindings to prefer to use this built-in type because `JString` now implements `AsRef<JCharSequence>`, which won't be true for any external `CharSequence` binding.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- `JCharSequence` bindings for `java.lang.CharSequence` (including `AsRef<JCharSequence>` + `.as_char_sequence()` for `JString`) ([#793](https://github.com/jni-rs/jni-rs/pull/793))
+
 ## [0.22.3] — 2026-03-05
 
 #### Fixed
--- a/crates/jni-macros/src/types.rs
+++ b/crates/jni-macros/src/types.rs
@@ -597,6 +597,12 @@ impl TypeMappings {
                 false,
             ),
             (
+                "JCharSequence",
+                "java.lang.CharSequence",
+                "objects::JCharSequence",
+                false,
+            ),
+            (
                 "JClassLoader",
                 "java.lang.ClassLoader",
                 "objects::JClassLoader",
--- a/crates/jni/src/lib.rs
+++ b/crates/jni/src/lib.rs
@@ -441,6 +441,8 @@ pub fn __test_bindings_init(env: &crate::Env, loader: &crate::refs::LoaderContex
     objects::JClassLoaderAPI::get(env, loader)
         .expect("Failed to initialize JClassLoaderAPI bindings");
     objects::JClassAPI::get(env, loader).expect("Failed to initialize JClassAPI bindings");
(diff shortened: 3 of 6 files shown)


rib:
cc: @wuwbobo2021 this might be of interest to you too, since I noticed your `jni-min-helper` readme includes `JCharSequence` bindings.

Actually I was going to initially declare `JCharSequence` to be a "core" type that could not be re-defined outside of the jni crate before I saw that you were already defining your own `JCharSequence` binding.

Since I've changed this to not declare `JCharSequence` as a core type then this shouldn't break your existing bindings but it could make sense for you to switch to using this built in type because then you would benefit from `JString` implementing `AsRef<jni::objects::JCharSequence>` (which you can't achieve with an external binding).

wuwbobo2021:
I just realized that built-in JNI types are probably not fully listed in the `bind_java_type` macro documentation.

Actually, the custom CharSequence binding is used in the example of jni-min-helper, but not used in the source of this library (I haven't used the custom CharSequence binding elsewhere). This alone can not be a significant reason of making any decision, but I know you're afraid of breaking programs owned by other people.

I guess that it is reasonable to have all class types directly involved with the low-level JNI functions wrapped by jni-rs as "core" types (e.g. java.nio.ByteBuffer). A list of these types may be determined so that there's no need of introducing other "core" types to be a SemVer breaking change.

However, I don't know if CharSequence should be a core type, because it is not involved in any JNI function (as I know): JNI string operation functions do not say that they support any instance with interface CharSequence be passed as an argument as an alternative of a String.
________________________________
发件人: Robert Bragg ***@***.***>
发送时间: 2026年3月15日 21:26
收件人: jni-rs/jni-rs ***@***.***>
抄送: wuwbobo2021 ***@***.***>; Mention ***@***.***>
主题: Re: [jni-rs/jni-rs] Add `objects::JCharSequence` binding (PR #793)


Merged #793<https://github.com/jni-rs/jni-rs/pull/793> into master.

―

rib:
> I just realized that built-in JNI types are probably not fully listed in the `bind_java_type` macro documentation.

Yes, ideally the documentation could clarify what types are builtin vs core, but really the idea is simply that all the jni::objects types will be builtin types, and then a smaller subset that are core to the jni crate (like `JObject` and `JClass`) are "core" types that can not be re-defined in custom bindings.

Although I initially went to make `JCharSequence` a core type, I think in retrospect it was correct to change it to just be a builtin type.

Like you say, `CharSequence` isn't a fundamental type that's required for core JNI functionality.

I initially thought it should be a core type because `JString` is effectively a subclass of `CharSequence` and it seemed odd for `JString` to be a core type but then have `CharSequence` not be a core type - but actually I think that's probably fine.

In general it's only expected to be a breaking change to add new core types and builtin types are just a convenience. Built-in types are convenient because you don't need to declare them via a `type_map` but you're allowed to re-define your own bindings for builtin types (as long as they are not core types).

PR #805, opened 2026-03-17
bind_java_type: support `#[jni(requires = <expr>)]` attributes for runtime version/requirement checks

### LINKED ISSUE #764: `bind_java_type` is missing support for conditional / version-checked method/field bindings
For Android, one might expect to use some deprecated methods on older API levels, and use some newly introduced methods on newer API levels. These newly introduced methods will not be found at runtime on older Android versions.

Maybe `{Type}API` can store IDs in `Options`, so that `None` values can be mapped to `MethodNotFound` and `FieldNotFound` errors. There can be another generated function of checking the availability of all fields and methods, or simply an option in `bind_java_type` for this issue.

rib:
This makes it possible to declare optional bindings with runtime-checked requirements.

For example, this can be used to add version checks to methods and fields like:

```rust
#[jni(requires = check_version(1, 2, 3)]
fn my_new_func() -> JString,
```

The 'requires' expression will be checked before `<Type>API::get()` does any corresponding method ID lookup or registers a native method implementation.

If multiple 'requires' attributes are given then the expressions will be logically chained together with `&&` to check that they are all true.

In the generated code, any 'requires' expressions are wrapped with an inner `fn __eval() -> bool {}` function before evaluation, so they can't capture any state from the scope in which they are evaluated (such as the `env`, `class` or `loader`).

If an API is called that depends on a missing ID (that failed its requirements check) then it will return `Error::UnsupportedVersion`.

This adds a '# Runtime Requirements and Version Checks' section to the `bind_java_type!` documentation to explain this functionality with a brief example.

The documentation also warns that it's the developer's responsibility to avoid any risk of recursion by attempting to use JNI within a check function to access a binding that itself needs to call the same check function.

Fixes: #764

### Definition of Done

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- `bind_java_type` now supports `#[jni(requires = ...)]` attributes on optional methods and fields, to conditionally enable them based on runtime checks like version checks ([#800](https://github.com/jni-rs/jni-rs/pull/800))
 
 ## [0.22.4] — 2026-03-16
 
--- a/crates/jni-macros/src/bind_java_type.rs
+++ b/crates/jni-macros/src/bind_java_type.rs
@@ -59,6 +59,18 @@ custom_keyword!(__jni_core);
 custom_keyword!(raw);
 custom_keyword!(non_null);
 
+/// Represents a requires expression for conditional binding
+#[derive(Clone)]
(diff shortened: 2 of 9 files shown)


rib:
If would be good if you could please take a look at this @wuwbobo2021 and confirm whether you think this would address your needs in #764

rib:
Btw, @wuwbobo2021 - in the context of Android API level checks, I would imagine implementing a check function based on the [`android_system_properties`](https://crates.io/crates/android_system_properties) crate to query the API level (as opposed to using JNI in the version check).

PR #811, opened 2026-04-14
Fix array/object type check in Env::{set_field, set_static_field}

alexjg:
## Overview

`Env::{set_static_field, set_field}` throw a `WrongJValueType` error when passing an array value to a field with an array signature. This is due to a check that the signature type matches the incoming value type which appears to not take account of the fact that an array signature will have type `JavaType::Array` whilst a newly created array will have `JavaType::Object`. In this PR I've modified the logic so that object values are acceptable for array signatures. This behavior seems to be compatible with the behavior of `call_method` so it seems like it's the right direction.

## Details

In `Env::{set_static_field, set_field}` we perform a check that the incoming `JValue` matches the `FieldSignature::ty()` of the `FieldSignature` we were passed. When passing an array as the value this check is inconsistent with the JNI spec.

For example, given this class

    class Example {
        public byte[] byteArrayField;
    }

The following code will fail with a WrongJValueType error:

    let instance = env.new_object(..)?; // Create a new Example

    // Set the `byteArrayField` field to a new value
    let arr = env.new_byte_array(1)?;
    env.set_field(
        instance,
        jni_str!("byteArrayField"),
        jni_sig!("[B"),
        JValue::Object(&arr),
    )?;

The reason this fails is that the type check in `set_field` and `set_static_field` looks like this:

    if value.java_type() != sig.ty() {
        return Err(Error::WrongJValueType(value.type_name(), "see java field"));
    }

But for an array the `JValue::java_type()` returns `JavaType::Object` whilst for a `[B` signature `FieldSignature::ty()` returns `JavaType::Array`

This commit adds a `FieldSignature::type_is_compatible_with(JavaType)` method which accepts either a `JavaType::Object` or `JavaType::Array` for array signatures.

### DIFF
--- a/crates/jni/src/signature.rs
+++ b/crates/jni/src/signature.rs
@@ -196,6 +196,17 @@ impl<'sig> FieldSignature<'sig> {
     pub fn ty(&self) -> JavaType {
         self.ty
     }
+
+    /// Check if the type of this field is compatible with the given type. This
+    /// is not quite an equality check on `Self::ty()` because `JavaType::Array`
+    /// signatures are considered compatible with `JavaType::Object` values
+    pub fn type_is_compatible_with(&self, ty: JavaType) -> bool {
+        match (self.ty, ty) {
+            (JavaType::Array | JavaType::Object, JavaType::Array | JavaType::Object) => true,
+            (a, b) if a == b => true,
+            _ => false,
+        }
+    }
 }
 
 /// A runtime-parsed JNI method signature.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+#### Fixed
+
+- `Env::{set_field, set_static_field}` no longer throw `WrongJValueType` when passed array values for array signatures
+
 
 ## [0.22.4] — 2026-03-16
 
--- a/crates/jni/src/env.rs
+++ b/crates/jni/src/env.rs
@@ -3997,9 +3997,8 @@ See the jni-rs Env documentation for more details.
         let obj = obj.as_ref();
         let obj = null_check!(obj, "set_field obj argument")?;
         let sig = sig.as_ref();
-        let field_ty = sig.ty();
 
-        if value.java_type() != field_ty {
+        if !sig.type_is_compatible_with(value.java_type()) {
             return Err(Error::WrongJValueType(value.type_name(), "see java field"));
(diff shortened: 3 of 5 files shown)


crates/jni/src/signature.rs
```diff
@@ -196,6 +196,17 @@ impl<'sig> FieldSignature<'sig> {
     pub fn ty(&self) -> JavaType {
         self.ty
     }
+
+    /// Check if the type of this field is compatible with the given type. This
+    /// is not quite an equality check on `Self::ty()` because `JavaType::Array`
+    /// signatures are considered compatible with `JavaType::Object` values
+    pub fn type_is_compatible_with(&self, ty: JavaType) -> bool {
+        match (self.ty, ty) {
+            (JavaType::Array, JavaType::Object) => true,
```
rib: I guess the opposite should also be considered valid? I.e. if we have an Object-type field then an array-type would be a valid Object too.

Maybe we can match like `(JavaType::Array | JavaType::Object, _) if ty.primitive_type().is_none() => true`

rib: or actually I suppose it makes more sense to be symmetric like `(JavaType::Array | JavaType::Object, JavaType::Array | JavaType::Object) => true,`

alexjg: I did consider this but went with the most conservative thing that made the check work. Happy to update it to be symmetric

rib: In practice, we'll never see a `ty` with `JavaType::Array` here currently (because a `JValue` passed to `set_field` for an array type already effectively normalizes that into an `::Object` type) but since this utility function only deals with `JavaType` then it should be a bit more general here (or otherwise we should put the check inline in `set_[static_]field` where we have more context)

rib (approved):
Thanks for looking at this - the change makes sense and looks good to me.

Thanks for adding tests too.

I made a minor change to `type_is_compatible_with` so that it logically also accepts a `JavaType::Array` value for an `Object|Array` field. Technically we won't see this currently because the `JValue` passed to `set_field` doesn't have an `Array` type (it would already be squashed to an `Object`) but this utility doesn't have that context and just defines the logical compatibility for `JavaType` pairs.

I took the liberty of tweaking the end of your commit message, where it mentioned the addition of the `type_is_compatible_with` utility.

crates/jni/src/signature.rs
rib: (I went ahead and made this change, and tweaked the end of your commit message)

alexjg:
Thanks for making those changes, commit message looks good to me!

PR #812, opened 2026-04-19
Extend docs for jni_mangle macro

FrankenApps:
## Overview

Adds a section to the documentation of the [`jni_mangle`](https://docs.rs/jni/latest/jni/attr.jni_mangle.html) macro, that demonstrates how to use the attribute when the *class name* is not already part of the *package name*.

This bit me when I first attempted to use the macro, therefore I thought this might help.

While at it I decided to also demonstrate how package names are mangled (which was my original assumption as to why the macro initially did not work as expected for me).


rib (approved):
Ah, right - thanks for the clarifying example

rib:
For reference, the (unrelated) CI failure here should be fixed by https://github.com/jni-rs/jni-rs/pull/815 and then I'll land this

PR #842, opened 2026-09-12
Update dependencies

djc:


crates/jni/src/vm/init_args/char_encoding_windows.rs
```diff
@@ -15,6 +15,13 @@ type WSize = c_int;
 // The type of Windows codepage numbers.
 type WCodepage = c_uint;
 
+// `windows-bindgen` emits these constants as `i32`, but `GetACP`, `WideCharToMultiByte` and
+// `MultiByteToWideChar` all take codepages and flags as `u32`.
+const CP_UTF7: WCodepage = windows_sys::CP_UTF7 as WCodepage;
```
djc, line 20: This seems pretty ugly, have filed an upstream issue about it:

- https://github.com/microsoft/windows-rs/issues/4920

djc:
@rib friendly ping?

Would be really nice to get rid of duplicate syn dependencies.

